PR #515, opened 2024-01-29
fix(landscape): Landscape reconnect signal does nothing

EduardGomezEscandell:
It just doesn't work.

This PR does multiple things around reconnection:
1. Fix the signal so that the Controller can ask Landscape to reconnect. (UDENG-2195)
2. Improve the way the connection is monitored, and restart it if it fails.
3. Implement a notification system in Config. Use this to reconnect to Landscape every time the relevant config is changed (UDENG-2196).

### DIFF
--- a/windows-agent/internal/proservices/landscape/landscape_test.go
+++ b/windows-agent/internal/proservices/landscape/landscape_test.go
@@ -102,7 +102,9 @@ func TestConnect(t *testing.T) {
 	for name, tc := range testCases {
 		tc := tc
 		t.Run(name, func(t *testing.T) {
-			ctx := context.Background()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
 			if wsl.MockAvailable() {
 				t.Parallel()
 				ctx = wsl.WithMock(ctx, wslmock.New())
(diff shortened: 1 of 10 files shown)


EduardGomezEscandell:
✅ end-to-end tests pass locally

windows-agent/internal/proservices/landscape/landscape_test.go
```diff
@@ -540,6 +538,157 @@ func TestAutoReconnection(t *testing.T) {
...
 
+func TestReconnect(t *testing.T) {
+	t.Parallel()
+
+	requestReconnect := func(ctx context.Context, s *landscape.Service, c *mockConfig) {
+		s.Controller().Reconnect(ctx)
+	}
+
+	changeAddress := func(ctx context.Context, s *landscape.Service, c *mockConfig) {
+		// We change the address to an equivalent one, so that the reconnect is triggered abd the connection succeeds
```
CarlosNihelton, line 549:
```suggestion
		// We change the address to an equivalent one, so that the reconnect is triggered and the connection succeeds
```

windows-agent/internal/proservices/landscape/landscape_test.go
```diff
@@ -540,6 +538,157 @@ func TestAutoReconnection(t *testing.T) {
...
+				return service.Connected()
+			}, 5*time.Second, 500*time.Millisecond, "Client should have reconnected after restarting the server")
+
+			ch := make(chan struct{})
+			go func() {
+				tc.trigger(ctx, service, conf)
+				close(ch)
+			}()
+
+			fmt.Println("AAAAAAAAA")
```
CarlosNihelton, line 643: left-over after some printf-debugging section?

CarlosNihelton (changes requested):
two minor nits

windows-agent/internal/proservices/landscape/landscape_test.go
EduardGomezEscandell: Hahaha you can see my desperation there

CarlosNihelton (approved):
LGTM

PR #517, opened 2024-01-29
feat(config): Implement notification system for Config

EduardGomezEscandell:
I want cloud-init & landscape to be pinged by the config every time something changes. However, I don't want these two utilities to be dependencies of config. If anything, I want the dependency the other way around! Cloud-init and Landscape should import config.

So I must somehow call these utilities from the config without importing them. My solution: callbacks. Any utility can now wait on config changes by using:

```go
conf.Notify(func() {
    fmt.Println("Hello, world")
})
```
and this function will be called every time the config is changed (via GUI, registry, or any future methods).

This is used in Landscape #515 and Cloud-init #506.

### DIFF
--- a/windows-agent/internal/config/config.go
+++ b/windows-agent/internal/config/config.go
@@ -47,7 +47,7 @@ func New(ctx context.Context, cachePath string) (m *Config) {
 	return m
 }
 
-// Notify appends a callback. It'll be called very time any configuration changes.
+// Notify appends a callback. It'll be called every time any configuration changes.
 func (c *Config) Notify(f func()) {
 	c.notifiers = append(c.notifiers, f)
 }
--- a/windows-agent/internal/config/config.go
+++ b/windows-agent/internal/config/config.go
@@ -33,8 +33,8 @@ type Config struct {
 	// Sync
 	mu *sync.Mutex
 
-	// notifiers are called after any configuration changes.
-	notifiers []func()
+	// observers are called after any configuration changes.
+	observers []func()
 }
 
 // New creates and initializes a new Config object.
@@ -49,7 +49,16 @@ func New(ctx context.Context, cachePath string) (m *Config) {
 
 // Notify appends a callback. It'll be called every time any configuration changes.
 func (c *Config) Notify(f func()) {
-	c.notifiers = append(c.notifiers, f)
+	c.observers = append(c.observers, f)
+}
+
+func (c *Config) notifyObservers() {
+	for _, f := range c.observers {
+		// This needs to be in a goroutine because notifyObservers is sometimes
+		// called under the config mutex. The callback trying to grab the mutex
+		// (to read the config) would cause a deadlock otherwise.
+		go f()
+	}
 }
 
 // Subscription returns the ProToken and the method it was acquired with (if any).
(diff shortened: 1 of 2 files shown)


windows-agent/internal/config/config.go
```diff
@@ -32,6 +32,9 @@ type Config struct {
 
 	// Sync
 	mu *sync.Mutex
+
+	// notifiers are called after any configuration changes.
+	notifiers []func()
```
CarlosNihelton, line 37: I'd have named this as `listeners` or `observers` since they receive a notification (qualifying the Config as the notifier - who emits the notifications).

windows-agent/internal/config/config.go
```diff
@@ -130,6 +138,10 @@ func (c *Config) set(ctx context.Context, field *string, value string) error {
 	old := *field
 	*field = value
 
+	for _, f := range c.notifiers {
+		go f()
```
CarlosNihelton, line 142: I assume here the callbacks will contain a reference to the Config, so the observers won't need to have the latest Config data passed in the call. Did you consider having such data as a parameter, thus avoiding the need for storing a reference to the Config? That's a common practice in callback systems of that nature, as well as passing copies of the data to prevent races is also common.

Finally, it's quite common as well to pass the old and the new data so the observer compares the bits it's interested in before taking any action , but I think that would be too much to our need, right.

windows-agent/internal/config/config.go
```diff
@@ -250,6 +262,10 @@ func (c *Config) collectRegistrySettingsTasks(ctx context.Context, data Registry
 		log.Warningf(ctx, "Could not obtain some updated registry settings: %v", errs)
 	}
 
+	for _, f := range c.notifiers {
+		go f()
```
CarlosNihelton, line 266: Is the possibility of a race condition of an observer being notified of a change and checking into the Config for data at the same time the data is being mutated in another goroutine irrelevant by design?

windows-agent/internal/config/config_test.go
```diff
@@ -638,6 +639,83 @@ func loadChecksums(t *testing.T, confDir string) (string, string) {
...
+// eventually solves the main issue with 'require.Eventually': you don't know what failed.
+// See what happens if you try to build the 'want vs. got' error message:
+//
+//	  require.Eventuallyf(t, func()bool{ return 5==got() },
+//			t, trate, "Mismatch: wanted %d but got %d", 5, got())
+//	                                                       ^^^
+//														   OUTDATED!
+//
+// Got is passed by value so we have the value obtained at t=0, not at t=timeout.
+//
```
CarlosNihelton, line 691: Should we propose a patch to the upstream require package?

CarlosNihelton (commented):
No blockers, just some design considerations to check.

windows-agent/internal/config/config.go
EduardGomezEscandell: I did consider it, but figured it was a bit silly:
```go
config.Notify(func(conf *Config) {
    // use conf
})
```
You already had access to the config to make the Notify call, so it seems a bit redundant. Config is always a pointer so it's always up to date.

I do not pass what the diff is because that'd introduce too much logic (at least for now, we only have two pieces of data anyways).

windows-agent/internal/config/config.go
EduardGomezEscandell: No possibility. All reads and writes to the config are protected by a mutex.

windows-agent/internal/config/config_test.go
EduardGomezEscandell: My solution is less generic as it requires an intermediate object:
```
predicate(getter())
```

Upstream does not require it:
```
predicate()
```
And this object must be representable as a string. In our specific case, this was not a problem. But it's a bit too restrictive for upstream I believe.

CarlosNihelton (approved):
LGTM 👍🏽

PR #519, opened 2024-01-30
maint: rename projects to Pro for WSL

didrocks:
Rename all projects and description to Pro for WSL

UDENG-2146


didrocks:
We did hit on the windows/common failure:

```
Run arduino/setup-protoc@v2
Error: Error: API rate limit exceeded for 13.93.155.208. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)
```

PR #523, opened 2024-01-30
ci: enable codecov

jibel:


didrocks (approved):
This is the correct file to upload ;)

.github/workflows/qa.yaml
```diff
@@ -231,7 +231,7 @@ jobs:
...
         run: |
           go test -shuffle=on ./... -race -tags=gowslmock
-      #- name: Upload coverage to Codecov
-      #  uses: codecov/codecov-action@v3
-      #  with:
-      #    file: /tmp/coverage.out
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v3
+        with:
+          file: /tmp/coverage.out
```
CarlosNihelton, line 237: The input option is `files` :p

https://github.com/codecov/codecov-action?tab=readme-ov-file#arguments

CarlosNihelton (changes requested):
I think I found a small mistake.

.github/workflows/qa.yaml
didrocks: Interesting, we are still using the `file:` in adsys which definitively works (https://github.com/ubuntu/adsys/blob/main/.github/workflows/qa.yaml#L80), but probably an initial option that has been deprecated and then unlisted. Let’s use the new format then!

EduardGomezEscandell (approved):
👍 

Looks like the gRPC action is rate-limiting us again 🙄. This PR does not touch any code so we can ignore the failure.

PR #527, opened 2024-01-31
maint(wsl-pro-service): Add vendored dependency copyright notices

EduardGomezEscandell:
I hated every second I was doing this.

---

UDENG-2213


EduardGomezEscandell:
Force-pushed to satisfy codecov

didrocks (approved):
At least, it’s not a year with leap second to double endure that given second :p

PR #530, opened 2024-01-31
ci: Report test coverage of the Flutter side of the project

CarlosNihelton:
And pick some low-hanging fruits to make it look better :p
The enum tests and exceptions were a pretty good point, though, as the added tests makes the l10n more future proof.

Closes UDENG-271

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/startup/startup_page.dart
+++ b/gui/packages/ubuntupro/lib/pages/startup/startup_page.dart
@@ -106,6 +106,8 @@ extension AgentStateL10n on AgentState {
       case AgentState.ok:
         // This state should not need translations.
         return '';
+      default:
+        throw UnimplementedError(toString());
     }
   }
 }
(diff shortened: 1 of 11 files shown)


gui/packages/ubuntupro/lib/pages/startup/startup_page.dart
```diff
@@ -106,6 +106,8 @@ extension AgentStateL10n on AgentState {
       case AgentState.ok:
         // This state should not need translations.
         return '';
+      default:
+        throw UnimplementedError(toString());
```
CarlosNihelton, line 110: Inacessible by design, only being caught if the enum is changed without this function being updated.

gui/packages/ubuntupro/lib/pages/subscription_status/subscribe_now_page.dart
```diff
@@ -115,6 +115,8 @@ extension PurchaseStatusl10n on PurchaseStatus {
         return lang.purchaseStatusServer;
       case PurchaseStatus.unknown:
         return lang.purchaseStatusUnknown;
+      default:
+        throw UnimplementedError(toString());
```
CarlosNihelton, line 119: Inacessible by design, only being caught if the enum is changed without this function being updated.

EduardGomezEscandell (approved):
I think we can ignore the warnings in this PR, it is the first one to collect the coverage anyways.

Still good to see you found a couple details to fix.

PR #541, opened 2024-02-06
fix(gui-ubuntupro): Update GUI strings to proprosed desgin

ashuntu:
The proposed changes for the Ubuntu Pro Flutter GUI design have differing strings, so this PR changes those strings and also the keys associated with them so they make more sense. These are all cosmetic changes and do not alter the functionality of the application.

*Force push was to correct the email used for the sake of CLA signing*

---

UDENG-1986

### DIFF
--- a/gui/packages/ubuntupro/end_to_end/end_to_end_test.dart
+++ b/gui/packages/ubuntupro/end_to_end/end_to_end_test.dart
@@ -80,7 +80,7 @@ Future<void> testManualTokenInput(WidgetTester tester) async {
   await tester.pumpAndSettle();
 
   // submits the input.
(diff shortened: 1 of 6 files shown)


CarlosNihelton:
The end-to-end tests won't run because they depend on secrets not available to forks. I should have warned you about that. You can merge this PR now.

ashuntu:
> The end-to-end tests won't run because they depend on secrets not available to forks. I should have warned you about that. You can merge this PR now.

Ok, I'll use a branch from this repo next time. I meant to, but I was running into issues with authentication and forked instead to rule out permission issues, but it didn't end up being the problem.

PR #547, opened 2024-02-06
maint(ci): inject token to report coverage with v4

didrocks:
codecov v4 action requires now a token.

UDENG-2246


CarlosNihelton (approved):
It would work if instead we passed the `directory` option. The issue is that codecov CLI doesn't treat our input as absolute paths, but rather filenames to search in the repository root or some other search directory provided.

jibel (approved):
All good. I checked the result on codecov too and the entire project seems to be covered.

PR #548, opened 2024-02-07
fix(WSL-Pro-Service): Increase timeout on the daemon's TestServeAndQuit

EduardGomezEscandell:
I honestly cannot understand why it is failing: I tried running with GOMAXPROCS=1, and adding time.Sleeps in there to mimic a slow machine. But I still pass the tests 🤷.

Let's hope this does it though.


didrocks (approved):
Let’s see. I agree, this is hard to grasp.

PR #550, opened 2024-02-08
feat(docs): Getting Started

CarlosNihelton:
The last image of the tutorial is meant to be a diagram explaining what we achieve. That's still pending.

Apart from that I think we have enough contents for a few passes of review.

The intent here is to demonstrate the automation provided by UP4W, but without Landscape beta it's hard to show more than `pro status`. So I struggled a lot until I figured out what I believe it's the easiest way to start a Landscape server on a WSL instance and making it accessible to the host and other instances using no custom configuration for WSL. It should be easy for WSL users. I wish we could stay more out of the Landscape business, so if you find suggestions to hide it further I'm all for it.

https://canonical-ubuntu-pro-for-wsl.readthedocs-hosted.com/en/tutorial/tutorial
---
UDENG-2254


docs/tutorial/index.md
```diff
@@ -0,0 +1,527 @@
...
+Hit:9 https://esm.ubuntu.com/infra/ubuntu noble-infra-security InRelease        # Notice the ESM repositories
+Reading package lists... Done
+Building dependency tree... Done
+Reading state information... Done
+All packages are up to date.
+```
+
+Great! Your new WSL instance is granted access to extended security patches automatically!
+
+## Check that the new Ubuntu-Preview instance with registered in Landscape
```
EduardGomezEscandell, line 416:
```suggestion
## Check that the new Ubuntu-Preview instance was registered in Landscape
```

docs/tutorial/index.md
```diff
@@ -0,0 +1,527 @@
...
+
+Good job! Your instances are transparently managed by Landscape thanks to the automation provided by Ubuntu Pro for WSL.
+
+## Conclusion
+
+In this tutorial you learned how to ensure new Ubuntu WSL instances are automatically attached to Ubuntu Pro and
+registered into a self hosted Landscape server thanks to the automation provided by Ubuntu Pro for WSL.
+Here is a complete overview of what we just achieved:
+
+[Diagram]
```
EduardGomezEscandell, line 526: Add a  so that we can grep for this sorts of things latter.

EduardGomezEscandell (changes requested):
Gave it a quick scan assuming that this comes pre-approved by Teodora as well :P

I see that you put the tutorial directly in `index.md` rather than its separate subpage. I have no strong--or really any--feelings about this, but was this on purpose? If so I suppose we should do the same with the how-to guide.

docs/tutorial/index.md
EduardGomezEscandell: Funny, if don't use `tildes` GitHub hides the comment 😅

CarlosNihelton:
Just rebased to fix a merge conflict. No changes in the contents.

tmihoc:
> Skipping empty layers is a great idea. 

But it might not be an empty layer here after all. (Just spoke to Didier and I am now convinced we need two basic guides, 

- Set up UP4W for a single Windows host
- - Using the GUI
- - Using cloud-init  

- Set up UP4W for multiple Windows hosts at once
-- Using the registry
-- Using cloud-init

EduardGomezEscandell (approved):

PR #555, opened 2024-02-08
fix(windows-agent): Add Stop function to Config

EduardGomezEscandell:
This prevents using it during the process teardown. Particulary relevant to async work like the observers.

Using the config during teardown is, I believe, the reason why this test fails:

https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/7813886532/job/21314974693?pr=506#step:8:38

I think cloud-init is writing `agent.yaml` at the same time as the test teardown in trying to remove its temp directory. On Windows, you cannot remove an open file.

### DIFF
--- a/windows-agent/internal/config/config.go
+++ b/windows-agent/internal/config/config.go
@@ -23,6 +23,9 @@ import (
 // Config manages configuration parameters. It is a wrapper around a dictionary
 // that reads and updates the config file.
 type Config struct {
+	ctx    context.Context
+	cancel func()
+
 	// data
 	subscription subscription
 	landscape    landscapeConf
@@ -35,6 +38,7 @@ type Config struct {
 
 	// observers are called after any configuration changes.
 	observers []func()
+	wg        sync.WaitGroup
 }
 
 // New creates and initializes a new Config object.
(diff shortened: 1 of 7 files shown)


windows-agent/internal/config/config.go
```diff
@@ -66,6 +109,10 @@ func (c *Config) Subscription(ctx context.Context) (token string, source Source,
 	c.mu.Lock()
 	defer c.mu.Unlock()
 
+	if c.stopped() {
+		return "", SourceNone, errors.New("config stopped")
+	}
+
 	if err := c.load(); err != nil {
 		return "", SourceNone, fmt.Errorf("could not load: %v", err)
 	}
```
CarlosNihelton, line 118: I think we could avoid some repetition here by creating a private helper method like:

```go
func (c *Config) loadGuarded() (unlocker func(), err error) {
        c.mu.Lock()
	// defer c.mu.Unlock() <- let callers defer unlock

	if c.stopped() {
		return nil, errors.New("config stopped")
	}

	if err := c.load(); err != nil {
		return nil, fmt.Errorf("could not load: %v", err)
	}
	
	return c.mu.Unlock, nil
}
```

Then callers would be simpler, such as the `Subscription` method.

```go
// Subscription returns the ProToken and the method it was acquired with (if any).
func (c *Config) Subscription(ctx context.Context) (token string, source Source, err error) {
	unlocker, err := c.loadGuarded()
	if err != nil {
		return "", SourceNone, err
	}
	
	defer unlocker()
	token, source = c.subscription.resolve()
	return token, source, nil
}
	
```

CarlosNihelton (changes requested):
I think we can deduplicate a little bit.

windows-agent/internal/config/config.go
EduardGomezEscandell: It is true that this code is duplicated, but I am wary of putting the duplicated part into a single function as the functionality is so disparate. I think extracting it would hurt readablity.

I think this duplication is acceptable so long as it is only twice, and the two instances are very close to one another. The latter we already fulfill and the former will be true once I get rid of `ProvisioningTasks` in #506.

EduardGomezEscandell:
Rebased to fix conflicts, no new code.

windows-agent/internal/config/config.go
CarlosNihelton: Looking at that branch, I see that snippet (lock-if-non-stopped-load) repeated in: `Subscription()`, `LandscapeClientConfig()`, `set(field *string, value string)`, `LandscapeAgentUID() ` and `collectRegistrySettingsTasks(ctx context.Context, data RegistryData, db *database.DistroDB)`. Will you remove more of those functions? They don't look so close to each other in my opinion. It looks not only as duplication, but a extra responsibility for the `Config`. Also, I don't see any call to `c.load()` without the lock-if-non-stopped preamble. While I realize that is not only loading that's protected under the mutex (and thus the reason for returning means to unlock the mutex ), either making `c.load()` lock the mutex and return the unlocker or implementing a lock guard mechanism action upon load should be done in my opinion.

Or better yet, segregate the sink and stream parts of the Config object, which right now seems like a complex stream with more than once sink (thus the need for the synchronization in every single method). I realize the second option is far more complex in the short term, though (ideally we'd have a queue of events causing the Config data to change and the existing notification mechanism propagating such changes once the last event in the queue completed).

CarlosNihelton: The get approach makes far more sense, as long as we don't hold the lock after it, that should be just fine.

CarlosNihelton (approved):
Great! :tada:

PR #558, opened 2024-02-08
fix(end-to-end): Ensures apt update before installing wsl-pro-service

CarlosNihelton:


CarlosNihelton:
I don't get why coverage is reporting changes 😕

EduardGomezEscandell (approved):
> I don't get why coverage is reporting changes 😕

Happens to me in all PRs, even some where I don't touch any code. Not sure what is going on (like this one 😓).

Good catch with the missing apt update btw.

didrocks:
If the coverage changes, it means that some tests are probably racy, and exits at different point.

We did this tough work on adsys to spot them (comparing the coverage between multiple runs and we are fully stable there. I think it’s worth investigating it once the beta is out or in any gardening tasks. Stable tests help to be confident in our coverage report.

PR #560, opened 2024-02-09
feat(logging): Make logs public and print version

EduardGomezEscandell:
Two-fold pull request:
- Windows Agent:
    - The logs are moved to the public directory @ $HOME/.ubuntupro/log
    - The logs print the version
- WSL Pro Service:
    - The logs print the version

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
@@ -3,6 +3,7 @@ package agent
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -108,11 +109,21 @@ func (a *App) serve(args ...option) error {
 		f(&opt)
 	}
 
-	publicDir, privateDir, err := setupDirectories(ctx, &opt)
+	publicDir, err := a.publicDir(opt)
 	if err != nil {
 		close(a.ready)
 		return err
 	}
+
+	log.Debugf(ctx, "Agent public directory: %s", publicDir)
+
+	privateDir, err := a.privateDir(opt)
+	if err != nil {
+		close(a.ready)
+		return err
+	}
+
+	log.Debugf(ctx, "Agent private directory: %s", privateDir)
 
 	proservice, err := proservices.New(ctx,
 		publicDir,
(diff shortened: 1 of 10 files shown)


EduardGomezEscandell:
Regarding codecov: these are changes to the main file, which cannot be tested.

EduardGomezEscandell:
Force-pushed to fix merge conflicts.

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
```diff
@@ -229,3 +201,50 @@ func (a App) RootCmd() cobra.Command {
...
+	return publicDir, nil
+}
+
+// publicDir is a wrapper around PublicDir to allow overriding its path with an option.
+func (a *App) publicDir(opts options) (string, error) {
+	if opts.publicDir == "" {
+		return a.PublicDir()
+	}
+
+	if err := os.MkdirAll(opts.publicDir, 0600); err != nil {
```
didrocks, line 227: I am a bit uneasy that both `PublicDir()` and `publicDir()` create the directory. Can we maybe simplify in one of those which takes an option?

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
```diff
@@ -181,6 +182,50 @@ func TestAppGetRootCmd(t *testing.T) {
...
+	// Not parallel because we modify the environment
+
+	testCases := map[string]struct {
+		emptyEnv bool
+		badPath  bool
+
+		wantErr bool
+	}{
+		"Success providing a public directory":               {},
+		"Error when %UserProfile% is empty":                  {emptyEnv: true, wantErr: true},
```
didrocks, line 195: I know there are 3 cases but still separating the success/error cases by a blank line sounds like a better way if we add future parameters to it :)

didrocks (changes requested):
My only concern is this `publicDir` vs `PublicDir` on the agent where we may be able to consolidate them in one (with a variadic optional parameter). That way, we would avoid having `MkdirAll` in 2 places.

Otherwise, great! Thanks for updating the path to separate user/sysadmin data vs config :)

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
EduardGomezEscandell: I just didn't want the public interface to take this optional argument as it is a bit redundant. But I guess it's fine.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
EduardGomezEscandell: Yep, my bad.

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
EduardGomezEscandell: In the end I inverted the logic (PublicDir calls publicDir).

I just did not want the App interface to have method `PublicDir(string)`. Feels like exposing its internals a bit.

PR #563, opened 2024-02-12
maint: Better logging

EduardGomezEscandell:
Went through the entire code-base making sure all error and log messages made sense and provide sufficient context. Also avoided stuttering as much as I reasonably could.

---

UDENG-926

### DIFF
--- a/windows-agent/internal/proservices/landscape/connection.go
+++ b/windows-agent/internal/proservices/landscape/connection.go
@@ -118,7 +118,7 @@ func newConnection(ctx context.Context, d serviceData) (conn *connection, err er
 // waits until the UID is received before returning.
 func handshake(ctx context.Context, d serviceData, conn *connection) (err error) {
 	defer decorate.OnError(&err, "could not complete handshake")
-	log.Debugf(ctx, "Landscape: starting handshake")
+	log.Debug(ctx, "Landscape: starting handshake")
 
 	// Send first message
 	info, err := newHostAgentInfo(conn.ctx, d)
--- a/windows-agent/internal/proservices/landscape/connection.go
+++ b/windows-agent/internal/proservices/landscape/connection.go
@@ -48,11 +48,11 @@ func newConnectionSettings(c landscapeHostConf) connectionSettings {
 // newConnection attempts to connect to the Landscape server, and blocks until the first
 // handshake is complete.
 func newConnection(ctx context.Context, d serviceData) (conn *connection, err error) {
-	defer decorate.OnError(&err, "could not connect to Landscape")
+	defer decorate.OnError(&err, "could not connect to Landscape server")
 
 	conf, err := newLandscapeHostConf(d.config())
 	if err != nil {
-		return nil, fmt.Errorf("could not read config: %v", err)
+		return nil, err
 	}
 	if conf.hostagentURL == "" {
 		return nil, errors.New("no hostagent URL provided in the Landscape configuration")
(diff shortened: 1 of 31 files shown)


windows-agent/internal/proservices/landscape/connection.go
```diff
@@ -118,11 +116,14 @@ func newConnection(ctx context.Context, d serviceData) (conn *connection, err er
 // If this is the first connection ever, the server will respond by assigning
 // the host a UID. This Recv is handled by receiveCommands, but handshake
 // waits until the UID is received before returning.
-func handshake(ctx context.Context, d serviceData, conn *connection) error {
+func handshake(ctx context.Context, d serviceData, conn *connection) (err error) {
+	defer decorate.OnError(&err, "could not complete handshake")
+	log.Debugf(ctx, "Landscape: starting handshake")
```
didrocks, line 121: `Debugf` but no parameter.

wsl-pro-service/internal/wslinstanceservice/wslinstanceservice.go
```diff
@@ -88,19 +89,20 @@ func (s *Service) sendInfo(ctx context.Context) error {
...
 	}
 
 	return nil
 }
 
 // ApplyLandscapeConfig serves LandscapeConfig messages sent by the agent.
 func (s *Service) ApplyLandscapeConfig(ctx context.Context, msg *wslserviceapi.LandscapeConfig) (empty *wslserviceapi.Empty, err error) {
-	defer decorate.OnError(&err, "ApplyLandscapeConfig error")
+	defer decorate.LogOnError(err)
+	defer decorate.OnError(&err, "ApplyLandscapeConfig")
```
didrocks, line 101: I find this chaining quite weird (return an error + logging is generally not a good pattern resulting in stuttering).

Also, on the error itself, you are prefixing here with the funciton name, that you are removing on other error log. I think we should be consistent

didrocks (changes requested):
I have 3 points (in 2 comments) that can be repeated throughout the diff, so only letting there. I think we should ensure we stay consistent on those ones.

Once addressed, I suggest that we take a holistic approach to this: merge, check the logs in a real prod environment, fix, merge, check… As this will be the best user experience with real code chaining.

wsl-pro-service/internal/wslinstanceservice/wslinstanceservice.go
EduardGomezEscandell: I only Log+return in one case: gRPC methods.

This is because this error is never printed, just sent back via gRPC (I think, not sure where it goes). The proper way of doing this would be with an interceptor, but his seemed like a quick-and-easy way to at least get the logs.

The inconsistency was accidental.

windows-agent/internal/proservices/landscape/connection.go
EduardGomezEscandell: I guess that linters don't notice this because it is not a `fmt` directive. Good catch.

wsl-pro-service/internal/wslinstanceservice/wslinstanceservice.go
didrocks: Ok, I think then we should log in a middleware when we return an error from the grpc server, wdyt?

EduardGomezEscandell: I agree, you caught me trying go get away with the lazy solution 😅. I'll try setting this up. Won't be too different from adsys I figure.

EduardGomezEscandell:
First new commit is [Fix Debugf without arguments -> Just use Debug](https://github.com/canonical/ubuntu-pro-for-wsl/pull/563/commits/20d9dfc351e81e1fc8bcd073e759f37e032e2988)

EduardGomezEscandell:
I moved yesterday's pair programming session results to #580

didrocks (approved):
Thanks, and splitting was great as I saw some tests were failing yesterday.

PR #583, opened 2024-02-16
fix: Align subscription status page with Provisioning design

ashuntu:
This PR redesigns the subscription status page to be more consistent with the designs for Provisioning. It includes having clickable text links, centering the title, text, and buttons, and outlining the text with a green background box.

This required changing the flow of the main layout and adding a `centered` option, so it may be more changes than you might suspect. This same layout will likely be used for the Landscape status page as well.

**New design**

image

**Old design (but with color changes and clickable link)**

Screenshot 2024-02-15 134408

---

UDENG-1786

### DIFF
--- a/gui/packages/ubuntupro/lib/l10n/app_en.arb
+++ b/gui/packages/ubuntupro/lib/l10n/app_en.arb
@@ -35,13 +35,10 @@
     "storeManaged": "This subscription is managed through Microsoft Store.",
     "manageSubscription": "Manage your subscription",
     "orgManaged": "This subscription is owned and managed by your organization.",
-    "manuallyManaged": "Visit [{proDashboardLink}]({proDashboardLinkHref}) to manage your subscription.",
+    "manuallyManaged": "Visit {proDashboardLink} to manage your subscription.",
     "@manuallyManaged": {
         "placeholders": {
             "proDashboardLink": {
-                "type": "String"
-            },
-            "proDashboardLinkHref": {
                 "type": "String"
             }
         }
--- a/gui/packages/ubuntupro/lib/l10n/app_en.arb
+++ b/gui/packages/ubuntupro/lib/l10n/app_en.arb
@@ -35,10 +35,13 @@
     "storeManaged": "This subscription is managed through Microsoft Store.",
     "manageSubscription": "Manage your subscription",
     "orgManaged": "This subscription is owned and managed by your organization.",
-    "manuallyManaged": "Visit {proDashboardLink} to manage your subscription.",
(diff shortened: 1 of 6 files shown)


gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -35,11 +35,14 @@
     "storeManaged": "This subscription is managed through Microsoft Store.",
     "manageSubscription": "Manage your subscription",
     "orgManaged": "This subscription is owned and managed by your organization.",
-    "manuallyManaged": "Visit {proDashboardLink} to manage your subscription.",
+    "manuallyManaged": "Visit [{proDashboardLink}]({proDashboardLinkHref}) to manage your subscription.",
```
CarlosNihelton, line 38: We must keep the localizable string as it was and producing the Markdown link syntax in the code, instead of in the translation (since that syntax is independent of l10n), to prevent translators from changing relative positioning of those placeholders by accident, thus breaking the syntax.

gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_page.dart
```diff
@@ -25,19 +26,19 @@ class SubscriptionStatusPage extends StatelessWidget {
...
         UserSubscriptionStatusModel() => SubscriptionStatus(
-            caption: lang.manuallyManaged('ubuntu.com/pro/dashboard'),
+            caption: lang.manuallyManaged(
+              'ubuntu.com/pro/dashboard',
+              'https://ubuntu.com/pro/dashboard',
+            ),
             actionButton: FilledButton(
               style: FilledButton.styleFrom(
-                backgroundColor: const Color(0xFFE86581),
+                backgroundColor: YaruColors.red,
```
CarlosNihelton, line 41: There is a todo item above you could have deleted, since you did what it was about. BTW I forgot to ask during the hangout: did you try `YaruColors.error` instead of `YaruColors.red`? Having some semantic names for the colors is a nice thing.

CarlosNihelton (changes requested):
Very good in general. There is only one show-stopper which I didn't consider when we were on the hangout about this: translators could break the syntax link by accident, thus I think we should compose it in the code, instead of doing it in the l10n resources.

gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_page.dart
ashuntu: There is `YaruColors.dark.error` and `YaruColors.light.error` but they didn't look as good or as close to the design as `YaruColors.red`.

CarlosNihelton (approved):
Good work! :tada:

PR #584, opened 2024-02-19
fix(windows-agent): Landscape install: create user when no cloud-init file is provided

EduardGomezEscandell:
Before supporting cloud-init, Landscape created distros with a default user. This PR:
1. Removes user creation when a cloud-init file is provided.
3. Refactors the old user creation when no cloud-init file is provided.


CarlosNihelton:
Since #1106 this should no longer be needed.

PR #586, opened 2024-02-20
feat(windows-agent): Daemon listens at the hyper-V adapter

EduardGomezEscandell:
Up until now we listened to all interfaces, this of course exposed us to the outside world.

With this change, we stay in our network, and we hope we'll also avoid ugly Windows Defender firewall pop-ups.

You can get this address in Powershell with:
```
Get-NetIPAddress -InterfaceAlias 'vEthernet (WSL (Hyper-V firewall))' -AddressFamily "IPv4" | Select-Object -ExpandProperty IPAddress
```

I prefered to avoid unnecessary subprocessing, so we instead use the Win32 API:

https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses

---

UDENG-2320

### DIFF
--- a/gui/packages/ubuntupro/lib/core/agent_api_paths.dart
+++ b/gui/packages/ubuntupro/lib/core/agent_api_paths.dart
@@ -19,8 +19,8 @@ String? agentAddrFilePath(String filename) {
 
 enum AgentAddrFileError { nonexistent, isEmpty, formatError, accessDenied }
 
-/// Reads the agent port from the addr file located at the full path [filepath].
-Future<Either<AgentAddrFileError, int>> readAgentPortFromFile(
+/// Reads the agent host and port from the addr file located at the full path [filepath].
+Future<Either<AgentAddrFileError, (String, int)>> readAgentPortFromFile(
   String filepath,
 ) async {
   try {
@@ -37,18 +37,37 @@ Future<Either<AgentAddrFileError, int>> readAgentPortFromFile(
       return const Left(AgentAddrFileError.isEmpty);
     }
 
-    final port = readAgentPortFromLine(lines[0]);
-    if (port == null) {
+    final address = parseAddress(lines[0]);
+    if (address == null) {
       // error: format error
       return const Left(AgentAddrFileError.formatError);
     }
 
-    return Right(port);
+    return Right(address);
   } on FileSystemException catch (_) {
     // error: permission denied
     return const Left(AgentAddrFileError.accessDenied);
   }
 }
 
-/// Parses [line] assuming it's from Windows Agent addr file. Returns null on error.
-int? readAgentPortFromLine(String line) => int.tryParse(line.split(':').last);
+/// Parses [line] assuming it's from Windows Agent addr file.
+/// Returns a (host, port) tuple on success, or null on error
+(String, int)? parseAddress(String line) {
+  final parts = line.split(':');
+  if (parts.length < 2) {
+    return null;
+  }
+
+  final host = parts.sublist(0, parts.length - 1).join(':');
+  final port = int.tryParse(parts.last);
+
+  if (port == null) {
+    return null;
+  }
+  if (port <= 0) {
+    // Port must be strictly positive.
+    return null;
+  }
+
+  return (host, port);
+}
--- a/windows-agent/internal/daemon/daemon.go
+++ b/windows-agent/internal/daemon/daemon.go
@@ -47,9 +47,14 @@ func (d Daemon) Serve(ctx context.Context) (err error) {
 
 	log.Debug(ctx, "Daemon: starting to serve requests")
 
-	// TODO: get a local port only, please :)
+	wslIP, err := getWslIP()
+	if err != nil {
+		log.Warningf(ctx, "Daemon: serving on all interfaces because we could not get the WSL adapter IP: %v", err)
+		wslIP = net.IPv4zero
+	}
+
 	var cfg net.ListenConfig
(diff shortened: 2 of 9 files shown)


windows-agent/internal/daemon/daemon.go
```diff
@@ -47,9 +47,14 @@ func (d Daemon) Serve(ctx context.Context) (err error) {
 
 	log.Debug(ctx, "Daemon: starting to serve requests")
 
-	// TODO: get a local port only, please :)
+	wslIP, err := getWslIP()
+	if err != nil {
+		log.Warningf(ctx, "Daemon: serving on all interfaces because we could not get the WSL adapter IP: %v", err)
```
didrocks, line 52: I think we should maybe just error out, wdyt?

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -144,6 +144,46 @@ func TestStartQuit(t *testing.T) {
...
+		server := grpc.NewServer()
+		var service testGRPCService
+		grpctestservice.RegisterTestServiceServer(server, service)
+		return server
+	}
+
+	d := daemon.New(ctx, registerer, addrDir)
+	defer d.Quit(ctx, false)
+
+	go func() { _ = d.Serve(ctx) }()
```
didrocks, line 165: should we have a channel that we close once finishing serving to ensure that we quit properly? (that way, we don’t potentially leak a goroutine even if those are for tests)

didrocks (changes requested):
I have just 2 comments, but excellent as usual. I followed along the Dart code, but I think we’ll see with Carlos once he is back to ensure that we didn’t introduce antipatterns for testing and so on.

So, as we are limiting the scope to hyperv addresses and if we do error out without one (which I really think we should always do), can we limit in the app xml definition the scope of addresses we are telling we are listening on?

didrocks:
I think you miss my global comment part:
> So, as we are limiting the scope to hyperv addresses and if we do error out without one (which I really think we should always do), can we limit in the app xml definition the scope of addresses we are telling we are listening on?

EduardGomezEscandell:
Yes, I did miss it 😅. I'll investigate.

EduardGomezEscandell:
> can we limit in the app xml definition the scope of addresses we are telling we are listening on?

Apparently not. The Firewall rules only allows for `Rule` sub-objects ([see here](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-desktop2-firewallrules)). These `Rule` objects only have the following properties ([see here](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-desktop2-rule)):

```xml
<desktop2:Rule
  Direction = 'A string that can have one of the following values: "in" or "out".'
  IPProtocol = 'A string that can have one of the following values: "ICMPv4", "ICMPv6", "TCP", "UDP", "GRE", "AH", "ESP", "EGP", "GGP", "HMP", "IGMP", "RVD", "OSPFIGP", "PUP", "RDP", or "RSVP".'
  Profile = 'A string that can have one of the following values: "domain", "private", "domainAndPrivate", "public", or "all".'
  LocalPortMin = 'An optional integer with a value between 1 and 65535 (inclusive).'
  LocalPortMax = 'An optional integer with a value between 1 and 65535.'
  RemotePortMin = 'An optional integer with a value between 1 and 65535.'
  RemotePortMax = 'An optional integer with a value between 1 and 65535.' />
```


I tried to reduce the permissions down to `private` newtworks only, but I got the Firewall pop-up.

didrocks (approved):
and thanks for the changes + the readability improvements! Unfortunate we can’t set a range of ip. Indeed, it’s still on a public address (public to your network, but not forwarded by routers).

We’ll see in the future if we will need to set a static port.

windows-agent/internal/daemon/export_test.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"testing"
+
+	wsl "github.com/ubuntu/gowsl"
+)
+
+// SetWslIPErr sets the WslIPErr variable to true, causing getWslIP to return an error.
+// This only works when the build tag is "linux" or "gowslmock".
+func SetWslIPErr(t *testing.T) {
```
CarlosNihelton, line 11: Honest question: why `t` as a parameter instead of receiver?

gui/packages/ubuntupro/lib/core/agent_api_paths.dart
```diff
@@ -19,8 +19,8 @@ String? agentAddrFilePath(String filename) {
 
 enum AgentAddrFileError { nonexistent, isEmpty, formatError, accessDenied }
 
-/// Reads the agent port from the addr file located at the full path [filepath].
-Future<Either<AgentAddrFileError, int>> readAgentPortFromFile(
+/// Reads the agent host and port from the addr file located at the full path [filepath].
+Future<Either<AgentAddrFileError, (String, int)>> readAgentPortFromFile(
```
CarlosNihelton, line 23: I'm not a big fan of tuples but this seems a very good use case for them, as having a class for that would be an overkill since there is no invariants between the port number and the host String. 👍🏽

CarlosNihelton (approved):
You excelled in the Dart side of the implementation! Well done!

windows-agent/internal/daemon/export_test.go
EduardGomezEscandell: `cannot define new methods on non-local type testing.T`

You can only use a type as a receiver if it is defined in the same package.

PR #588, opened 2024-02-20
fix(Landscape): fix bad time staircase

EduardGomezEscandell:
The "if error -> increase wait" logic did not work.

It now does work, and we also consider short-lived connections as errors (to avoid spamming 1-second connections that the server rejects for whatever reason, or spamming with flaky connections, etc).

This also fixes the recently flaky Landscape tests.

---

UDENG-2261

### DIFF
--- a/mocks/landscape/landscapemockservice/landscape_mock_service.go
+++ b/mocks/landscape/landscapemockservice/landscape_mock_service.go
@@ -71,7 +71,9 @@ type host struct {
 	send      func(*landscapeapi.Command) error
 	info      HostInfo
 	connected *bool
-	stop      func()
+
+	ctx  context.Context
+	stop func()
 }
 
 // Service is a minimalistic server for the landscape API.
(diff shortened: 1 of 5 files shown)


mocks/landscape/landscapemockservice/landscape_mock_service.go
```diff
@@ -263,6 +266,21 @@ func (s *Service) IsConnected(uid string) bool {
 	return s.isConnected(uid)
 }
 
+// WaitDisconnection returns a channel that will be closed when the connection with the host assigned to the
+// specified UID is terminated.
+//
+// If the UID is not registered, the second return value will be false.
+func (s *Service) WaitDisconnection(uid string) <-chan struct{} {
```
CarlosNihelton, line 273: I don't see this being used outside of tests. Is it for future PR's?

EduardGomezEscandell: Well it is a mock Landscape server, so it's only gonna be used it tests :P

PR #590, opened 2024-02-21
maint(ci): Allows that one percent coverage reduction

CarlosNihelton:
Should allow CI to be happier with racy tests
Until we figure out the source of fluctuations I guess.

All cases in CI which we could forgive coverage drop reports were well under 0.80%, so 1% seems not too tight.

Let see how this goes.

Ref: 
- https://docs.codecov.com/docs/commit-status
- https://docs.codecov.com/docs/codecov-yaml

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,9 @@
+coverage:
+  status:
+    project:
+      default:
+        threshold: 1%
+    patch:
+      default:
+        threshold: 1%
+
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,4 +1,5 @@
 coverage:
+  range: 85..100
   status:
     project:
       default:
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,8 +3,10 @@ coverage:
   status:
     project:
       default:
+        # Some tests are causing coverage fluctuation (race condition or flakyness)
+        # TODO: Investigate and fix, then reduce this threshold.
         threshold: 1%
     patch:
+      # TODO: Consider relaxing the patch target vs range for the global project in the long term, being more lenient on the patches themselves.
       default:
         threshold: 1%
-


CarlosNihelton:
> Could we use [coverage.range](https://docs.codecov.com/docs/codecovyml-reference#coveragerange) instead set to something like 85..100 like @EduardGomezEscandell suggested? With a percentage you can decrease progressively over time without noticing.

IIUC range must be applied in conjunction, not as a replacement for percents. So setting it to 85..100 will alert us if we drop too much, even though for the drop would be acceptable  a particular PR.

EduardGomezEscandell:
Without having read the docs and knowing if it's possible… I think we care about two things:
- Pre-existing code: no previously covered line becomes uncovered (or maybe keep it within some very strict tolerance).
- New code: keep it above a pre-set %.

I don't know if this is possible, but we should aim towards this direction.

jibel (approved):
Great addition! approved.

jibel:
On a side note, flaky tests, if any, must be reviewed and fixed. They usually hide something sneaky with the environment or the app.

codecov.yml
```diff
@@ -0,0 +1,10 @@
+coverage:
+  range: 85..100
+  status:
+    project:
+      default:
+        threshold: 1%
```
didrocks, line 6: I think we want a comment telling this is due to flaky tests and we should make it smaller once they are fixed.

codecov.yml
```diff
@@ -0,0 +1,10 @@
...
+    patch:
+      default:
+        threshold: 1%
```
didrocks, line 9: Let’s keep indeed `target: auto` by now, but we may need to relax `patch` target vs `range` for the global project in the long term, being more lenient on the patches themselves.

I suggest we revisit those behaviours after some PR.

didrocks (approved):
I think with the inheritence of `project` vs `patch`, this is mostly doing what Edu proposed too. I suggest adding a comment before merging on the threshold which is way too large, but due to the flaky tests for now, so that we think of revisiting it.

didrocks (approved):

PR #591, opened 2024-02-21
feat: Implement Landscape configuration page

ashuntu:
This comes with a few changes in multiple places:
- Ubuntu Pro Flutter GUI
    - Layouts were refactored to support the new two-column layout for the Landscape page
- Additional protobuf definitions in `agentapi/`
- Additional agent functions in `windows-agent/`

*(WIP so still a draft)*

image

---

UDENG-1785


gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -51,5 +51,14 @@
     "purchaseStatusUserGaveUp":"Purchase attempt was canceled.",
     "purchaseStatusNetwork":"A network issue prevented the purchase to proceed. Please try again once connected.",
     "purchaseStatusServer":"Something went wrong with Microsoft Store. Please try again later.",
-    "purchaseStatusUnknown": "Unknown error when trying to purchase the subscription. Consider restarting this app."
+    "purchaseStatusUnknown": "Unknown error when trying to purchase the subscription. Consider restarting this app.",
+
+    "landscapeHeading": "You can make your Ubuntu instances remotely with {landscapeLink}.",
```
CarlosNihelton, line 56:
```suggestion
    "landscapeHeading": "You can manage your Ubuntu instances remotely with {landscapeLink}.",
```

gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
```diff
@@ -0,0 +1,119 @@
...
+      default:
+        throw UnimplementedError('Unknown configuration status');
+    }
+
+    final subscriptionInfo = await client.subscriptionInfo();
+    final subscriptionType = subscriptionInfo.whichSubscriptionType();
+    return ![
+      SubscriptionInfo_SubscriptionType.none,
+      SubscriptionInfo_SubscriptionType.notSet,
+    ].contains(subscriptionType);
```
ashuntu, line 93: This almost certainly is not what we need to do to check if the config was applied successfully. Should we check at all, or just let the user go to the status page?

CarlosNihelton: I think we don't need to go that path. Following the gRPC method signature, if we receive an `Empty` the call completed successfully. Otherwise we'd have an error, which is translated into Dart as an exception.

ashuntu:
Moved to https://github.com/canonical/ubuntu-pro-for-wsl/pull/615

PR #600, opened 2024-02-23
fix(msix): Attempt to prevent firewall popus

CarlosNihelton:
By deploying firewall rules for all network profiles. I had success with this testing in VMs. I'd like to have others testing the build to make sure the firewall popups won't show up before merging.


didrocks (approved):
I think this is good for beta, I can’t wait on our next install! Then, we should log a Jira card to limit on some IP addresses and not listen to all interfaces.

EduardGomezEscandell:
I think we broke the build:

https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/8049768246/job/21984022791?pr=599#step:9:257

CarlosNihelton:
> I think we broke the build:
> 
> https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/8049768246/job/21984022791?pr=599#step:9:257

https://github.com/canonical/ubuntu-pro-for-wsl/pull/609

PR #601, opened 2024-02-23
maint(windows-agent): No TLS from agent to Landscape

CarlosNihelton:
Landscape quick setup for now at least is not talking to the agent over TLS. The team was advised and they are working to enable such feature. In the meanwhile let's use insecure credentials on the connection to Landscape. I'm doing this in a ugly but easy to revert way.

### DIFF
--- a/end-to-end/landscape_utils_test.go
+++ b/end-to-end/landscape_utils_test.go
@@ -3,23 +3,20 @@ package endtoend_test
 import (
 	"bytes"
 	"context"
-	"crypto/tls"
 	"fmt"
 	"log"
 	"log/slog"
 	"net"
-	"path/filepath"
 	"testing"
 	"time"
 
 	landscapeapi "github.com/canonical/landscape-hostagent-api"
-	"github.com/canonical/ubuntu-pro-for-wsl/common/testutils"
 	"github.com/canonical/ubuntu-pro-for-wsl/mocks/landscape/landscapemockservice"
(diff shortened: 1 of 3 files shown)


EduardGomezEscandell (changes requested):
If we want the smoothest experience, I'd move actually comment out these three lines:

https://github.com/canonical/ubuntu-pro-for-wsl/blob/016b024f66f657b6ed3bcab7e5809991d2aa7edc/windows-agent/internal/proservices/landscape/utils.go#L134-L137

This'll cause it to use insecure credentials thanks to this line:
https://github.com/canonical/ubuntu-pro-for-wsl/blob/016b024f66f657b6ed3bcab7e5809991d2aa7edc/windows-agent/internal/proservices/landscape/utils.go#L90-L92

This way, changing the ssl_public_key will not trigger an unnecessary reconnect.

Maybe we should also allow it with an environment variable `$env:UP4W_LANDSCAPE_HTTPS`.

CarlosNihelton:
> This way, changing the ssl_public_key will not trigger an unnecessary reconnect.
Good point!

> Maybe we should also allow it with an environment variable `$env:UP4W_LANDSCAPE_HTTPS`.
Not a big fan of the idea, HTTPS should be the way to go. Insecure setups can always be achieved with self-signed certificates, such as the Landscape quick setup does.

EduardGomezEscandell (approved):
LGTM once tests pass

PR #606, opened 2024-02-26
fix(Landscape): Don't log when there is no configuration

EduardGomezEscandell:
Before, Landscape tried to connect every 1, 2, 4, 8... seconds every time it failed, regardless of the cause of failure. Every failed attempt would produce a log line.

Now, when the issue is that Landscape is not configured (or there is no Ubuntu Pro Token), Landscape stops retrying, hence also no longer logging. If the config is changed, and enough info is in the config, a retrial will be attempted.

---

UDENG-2356
UDENG-2363


windows-agent/internal/proservices/landscape/utils.go
```diff
@@ -151,12 +169,15 @@ func newLandscapeHostConf(config Config) (conf landscapeHostConf, err error) {
...
-	if err == nil {
-		k, err := sec.GetKey("url")
-		if err == nil {
-			conf.hostagentURL = k.String()
-		}
+	if err != nil {
+		return landscapeHostConf{}, noConfigError{missing: "Host URL"}
+	}
+
+	k, err := sec.GetKey("url")
```
didrocks, line 176: I would name this u or hostagentURL, but not k :)

didrocks (changes requested):
One small other change in addition to what we dicussed, otherwise, looks good!

EduardGomezEscandell:
Build issue is being fixed here #612

EduardGomezEscandell:
Rebased to fix debian build. Only last two commits are new.

windows-agent/internal/proservices/landscape/utils.go
EduardGomezEscandell: k is for key, but yeah I can make it clearer.

CarlosNihelton (approved):
LGTM 👍🏽

windows-agent/internal/proservices/landscape/utils.go
didrocks: I realized from the get go it was for key, but here, we want to know what the is, not what it could be :)

didrocks (approved):

PR #612, opened 2024-02-27
fix(windows-agent): Fix launcher flags when installing a distro

EduardGomezEscandell:
Also sneak in the wsl-pro-service build fix


didrocks (approved):
\o/ when the test pass, ofc :)

PR #614, opened 2024-02-27
feat: Add automatic versioning

EduardGomezEscandell:
At least for the logs.

The version is 
```
(LAST TAG)[-(NUMBER OF COMMITS SINCE LAST TAG)-g(SHORT COMMIT SHA)]?[-DIRTY]?
```

---

Example for a release version
```
v0.1.0
```
Same thing but dirty:
```
v0.1.0-dirty
```

---

Example for a random commit, seven commits after tag v0.1.0.
```
v0.1.0-7-g65af1e
```

Same thing, but dirty:
```
v0.1.0-7-g65af1e-dirty
```

---

Example log:
```
time=2024-02-27T18:42:42+01:00 level=info msg=Version: v0.0.0-1-g37692ddf-dirty
```

---

UDENG-2262


EduardGomezEscandell:
This is not going to work until we have at least one tag.

tools/build/build-deb.sh
```diff
@@ -12,5 +12,8 @@ sudo DEBIAN_FRONTEND=noninteractive apt update
 sudo DEBIAN_FRONTEND=noninteractive apt install -y devscripts
 sudo DEBIAN_FRONTEND=noninteractive apt -y build-dep .
 
+# Update version if in repo
+git status &> /dev/null && go run ../tools/build/compute_version.go ../common/version || true
```
didrocks, line 16: I don’t think this is needed if you hook this up in debian/rules properly

didrocks (changes requested):
Multiple remarks on this PR:
- I don’t see how the binary are reading/using this ressource when compiled. I think the compiled binary should have the version set. This is the flag argument for the go command to include as a const.
- this is not hooked up in the package build system. We should have it setting our const. So we need to have this in debian/rules, and similar hooked up in msix.
- For dart, it seems we have something similar and should hook it up that way.

And finally, I don’t think the number of commits have any relevance here. As you can be in a nested branch or mainline, and we don’t have a rebase on mainline approach, this count will really be skewed. I think we can drop it and keep the commit sha which is the source or truth to see what’s the parent's commit are.

tools/build/compute_version.go
```diff
@@ -0,0 +1,70 @@
...
+		return "", fmt.Errorf("git rev-parse HEAD: %v. %s", err, out)
+	}
+	commit := strings.TrimSpace(string(out[:8]))
+
+	// Start with 0 so that it is always less than any tag.
+	return fmt.Sprintf("0-untagged-%s", commit), nil
+}
+
+func isDirty() (bool, error) {
+	out, err := exec.Command("git", "status").Output()
```
CarlosNihelton, line 61: `git status --porcelain`  outputs nothing if the working tree is clean, eliminating the need to verify the contents of the output if you just want to know that.

.github/workflows/qa-azure.yaml
```diff
@@ -23,6 +23,12 @@ jobs:
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
+      - uses: actions/setup-go@v5
+        with:
+          go-version-file: go.work
+      - name: Set up version
```
CarlosNihelton, line 29: We are not yet ingesting this information into the binaries, are we?

EduardGomezEscandell: No, not yet.

tools/build/build-deb.sh
EduardGomezEscandell: The issue is that this has to be computed at source-build time (when we have access to the whole repo). The source package does not contain the .git folder and hence cannot find the current tag, commit, or anything similar.

So we have to write the name of the version somewhere, and that may as well be `common/version`.

EduardGomezEscandell:
Rebased and squashed because the solution is essentially completely new.

.github/workflows/qa-azure.yaml
```diff
@@ -71,8 +77,22 @@ jobs:
...
         run: |
+          # Set up the full version (displayed in logs)
+          $env:UP4W_FULL_VERSION=$(go run .\tools\build\compute_version.go)
+          
+          # Set up the numeric version (used in the AppxManifest)
+          $UP4W_VERSION=$(go run .\tools\build\compute_version.go --numeric)
+          [Reflection.Assembly]::LoadWithPartialName("System.Xml.Linq")
+          $path = "$PWD/Msix/UbuntuProForWSL/Package.appxmanifest"
+          $doc = [System.Xml.Linq.XDocument]::Load($path)
+          $xName = [System.Xml.Linq.XName]::Get("{http://schemas.microsoft.com/appx/manifest/foundation/windows10}Identity")
```
didrocks, line 88: This is better than xpath!

gui/packages/ubuntupro/lib/constants.dart
```diff
@@ -9,3 +9,9 @@ const kDefaultMargin = 32.0;
 
 /// The path of the agent executable relative to the msix root directory.
 const kAgentRelativePath = 'agent/ubuntu-pro-agent-launcher.exe';
+
+/// The full decorated version string
+const kVersion = String.fromEnvironment(
+  'UP4W_FULL_VERSION',
```
didrocks, line 15: Unsure to understand, to me, this is taking from environment variable, but at runtime, not build-time, right? Or is this thingy use when compiled?

msix/UbuntuProForWSL/Package.appxmanifest
```diff
@@ -1,21 +1,6 @@
...
-  xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5"
-  xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities"
-  IgnorableNamespaces="desktop desktop2 desktop6 virtualization uap uap5 rescap">
-
-  <Identity
-    Name="CanonicalGroupLimited.UbuntuPro"
-    Publisher="CN=23596F84-C3EA-4CD8-A7DF-550DCE37BCD0"
-    Version="0.1.0.0" />
-
+<Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:desktop="http://schemas.microsoft.com/appx/manifest/desktop/windows10" xmlns:desktop2="http://schemas.microsoft.com/appx/manifest/desktop/windows10/2" xmlns:desktop6="http://schemas.microsoft.com/appx/manifest/desktop/windows10/6" xmlns:virtualization="http://schemas.microsoft.com/appx/manifest/virtualization/windows10" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" xmlns:uap5="http://schemas.microsoft.com/appx/manifest/uap/windows10/5" xmlns:rescap="http://schemas.microsoft.com/appx/manifest/foundation/windows10/restrictedcapabilities" IgnorableNamespaces="desktop desktop2 desktop6 virtualization uap uap5 rescap">
```
didrocks, line 2: did you need to do that for the maching in CI or was it autoformatted?

msix/UbuntuProForWSL/Package.appxmanifest
```diff
@@ -1,21 +1,6 @@
...
+  <Identity Name="CanonicalGroupLimited.UbuntuPro" Publisher="CN=23596F84-C3EA-4CD8-A7DF-550DCE37BCD0" Version="0.0.0.0" />
```
didrocks, line 3: you did reset the version btw (it means that we never updated it while uploading to the store too). As long as every manual build (even with vscode) takes a local version, this is fine.

msix/UbuntuProForWSL/Package.appxmanifest
```diff
@@ -30,61 +15,43 @@
...
-
   <Dependencies>
     <TargetDeviceFamily Name="Windows.Universal" MinVersion="10.0.0.0" MaxVersionTested="10.0.0.0" />
     <TargetDeviceFamily Name="Windows.Desktop" MinVersion="10.0.14393.0" MaxVersionTested="10.0.14393.0" />
     <PackageDependency Name="Microsoft.VCLibs.140.00.UWPDesktop" MinVersion="14.0.24217.0" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" />
   </Dependencies>
-
   <Resources>
-    <Resource Language="x-generate"/>
+    <Resource Language="x-generate" />
```
didrocks, line 24: yeah, I think you did run some automatter, no? I have no issue here, but on the ones below, I found the xml way more readable before the commit.

tools/build/build-appx.ps1
```diff
@@ -71,6 +71,23 @@ function Update-Certificate {
...
+    # Update the AppxManifest version
+    [Reflection.Assembly]::LoadWithPartialName("System.Xml.Linq")
+    $path = "$PWD/Msix/UbuntuProForWSL/Package.appxmanifest"
+
+    Write-Output "Setting version to $UP4W_VERSION in file $path"
+    Write-Output "Building with full-version $env:UP4W_FULL_VERSION"
+
+    $doc = [System.Xml.Linq.XDocument]::Load($path)
+    $xName = [System.Xml.Linq.XName]::Get("{http://schemas.microsoft.com/appx/manifest/foundation/windows10}Identity")
+    $doc.Root.Element($xName).Attribute("Version").Value = "$UP4W_VERSION.0";
```
didrocks, line 87: yeah, it seems you didn’t need to reset to `0.0.0.0`, no?

.github/workflows/qa-azure.yaml
```diff
@@ -71,8 +77,22 @@ jobs:
           Import-PfxCertificate -Password $pwd -CertStoreLocation Cert:CurrentUser\My -FilePath certificate\certificate.pfx
       - name: Build Bundle
         run: |
+          # Set up the full version (displayed in logs)
+          $env:UP4W_FULL_VERSION=$(go run .\tools\build\compute_version.go)
+          
+          # Set up the numeric version (used in the AppxManifest)
+          $UP4W_VERSION=$(go run .\tools\build\compute_version.go --numeric)
```
didrocks, line 84: hum, so we have something different between the agent components themselves and the version set in the package.

Do you think we can write something in the some description section of the package, so that we have the exact version with the git sha?

tools/build/compute_version.go
```diff
@@ -0,0 +1,80 @@
...
+	return matches[1]
+}
+
+func getTag() (string, error) {
+	// Note: we cannot use --dirty because it does not detect untracked files.
+	out, err := exec.Command("git", "describe", "--tags", "HEAD").CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("git describe: %v. %s", err, out)
+	}
+
```
didrocks, line 66: TBH, I would be more leniant and this would be `getTagVersion` and removing any `v` as prefix. Then, you simplify the regexp.

wsl-pro-service/debian/rules
```diff
@@ -14,6 +14,11 @@ export GOTOOLCHAIN := local
 	dh $@ --builddirectory=_build --with=apport
 
 execute_after_dh_auto_clean:
+	# Overriding version file
+	# We must do this during the source package build because we still have access to git.
+	# We can't do this during the binary package, so we cannot use the usual -ldflags.
```
didrocks, line 19: Be careful that debian/rules clean is called during package binary build too. Maybe we should have a `if source build` do that, and put everything under it (vendoring and so on).

I suggest that we only trigger this in one way.

wsl-pro-service/debian/rules
```diff
@@ -14,6 +14,11 @@ export GOTOOLCHAIN := local
...
+	[ "$(shell cat ../common/version)" = "Dev" ] && go run ../tools/build/compute_version.go > ../common/version || true
```
didrocks, line 20: That would not work right? because we will not have `../` anymore from the source package? It needs to be inside the source package itself (so under that directory).

didrocks (changes requested):
Some more comments and thoughts for you!

msix/UbuntuProForWSL/Package.appxmanifest
EduardGomezEscandell: It's autoformatted. It's the output of `$doc.Save($path)` in the powershell build script.

msix/UbuntuProForWSL/Package.appxmanifest
EduardGomezEscandell: This always gets overridden when you build, so this value does not really matter.

Hence, we decided for the least meaningful value, `0.0.0`.

gui/packages/ubuntupro/lib/constants.dart
CarlosNihelton: This `.fromEnvironment`  is a confusing name. It means "retrieve the value for the specified name in the compilation configuration environment declaration". It's a compile time constant defined during build via the command line option `--dart-define`.

msix/UbuntuProForWSL/Package.appxmanifest
CarlosNihelton: The formatting changes are consequence of the XML manipulation via powershell.

msix/UbuntuProForWSL/Package.appxmanifest
EduardGomezEscandell: Yeah, since building the Appx reformats the manifest, I decided to commit the autoformatted file. This way the diff after building is just the version number.

wsl-pro-service/debian/rules
EduardGomezEscandell: > Be careful that debian/rules clean is called during package binary build too

Yeah, that if is there to avoid calling it during binary build. I agree though, a single isSource signal would be best. You think the vendor cek is good enough?

didrocks: I’m not a fan of either approach, I’m looking if we can have a better decision factor.

msix/UbuntuProForWSL/Package.appxmanifest
CarlosNihelton: Just for the record, the funny validation regex allow building and side-loading a package with version `0.0.0.0` (in the beginning I thought we couldn't do that, but we can).

wsl-pro-service/debian/rules
didrocks: I would argue for the root of `wsl-pro-service`, so that it’s clear it’s the file that is generated from the project.

.github/workflows/qa-azure.yaml
EduardGomezEscandell: The difference is the tail end:
```
v0.1.0-7-gdeadbeef-dirty
0.1.0
```

tools/build/compute_version.go
EduardGomezEscandell: Removed the `v` from the version regex 👍

.github/workflows/qa-azure.yaml
didrocks: You probably miss the second part of my remark which was: can we write it in some part of package description which is readable once we install this manually?

wsl-pro-service/internal/consts/consts.go
```diff
@@ -1,9 +1,16 @@
...
 
 const (
 	// DefaultLogLevel is the default logging level selected without any option.
 	DefaultLogLevel = log.WarnLevel
 )
+
+// Version is the version of the service
+//
+// It is set at build time using the -ldflags option.
+var Version = "Dev"
```
didrocks, line 16: We do have ` consts/Version`  and ` common/Version`  then?

It seems that replacing common/Version works for the agent build. Do you know why it doesn’t work in the debian package build? We are in the same condition (the module is outside of the current package). My only bet is that this is due to vendoring maybe? (this is the only different I can see between the two), but can you assert on this before we commit having 2 `Version` variables?

If so, should the other remains in `common/`  as it’s not really common aymore :p

didrocks (commented):
I think we should investigate the consts vs common dilemna and see if we can add additional useful information to the msix itself, the rest looks good! :)

wsl-pro-service/internal/consts/consts.go
EduardGomezEscandell: I suspect it is due to vendoring, but I'm not 100% sure.

I agree that it makes no sense for the other to be in `common` then.

didrocks (approved):
Phew! That was more complex than anticipated, well done!

PR #615, opened 2024-02-27
Landscape page new

ashuntu:
This comes with a few changes in multiple places:
- Ubuntu Pro Flutter GUI
    - Layouts were refactored to support the new two-column layout for the Landscape page
- Additional protobuf definitions in `agentapi/`
- Additional agent functions in `windows-agent/`

*(WIP so still a draft)*

image

Old PR: https://github.com/canonical/ubuntu-pro-for-wsl/pull/591

---

UDENG-1785


gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -51,5 +51,30 @@
...
+        "placeholders": {
+            "landscapeLink": {
+                "type": "String"
+            }
+        }
+    },
+    "landscapeQuickSetup": "Quick Setup (Recommended)",
+    "landscapeQuickSetupHint": "Provide your Landscape enrollment information",
+    "landscapeFQDNLabel": "Landscape FQDN",
+    "landscapeFQDNError": "Invalid URI",
```
CarlosNihelton, line 67: We can be a bit more descriptive here, like `Invalid URI. Format should be HOST.DOMAIN or IP_ADDR.`

gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
```diff
@@ -0,0 +1,134 @@
...
+
+  set path(String value) {
+    _path = value;
+    validatePath();
+  }
+
+  String get path => _path;
+
+  bool validateFQDN() {
+    _fqdnError = _fqdn.isEmpty || Uri.tryParse(_fqdn) == null;
```
CarlosNihelton, line 59: FQDN also shouldn't have a port number. This check could then be something like:
```suggestion
   final uri = Uri.tryParse(_fqdn)
    _fqdnError = _fqdn.isEmpty || uri == null || uri.hasPort;
```

gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
```diff
@@ -0,0 +1,134 @@
...
+  Future<void> applyLandscapeConfig() async {
+    final file = File(_path);
+    final content = await file.readAsString();
+    await client.applyLandscapeConfig(content);
+  }
+
+  Future<void> applyManualLandscapeConfig() async {
+    final config = '''
+[host]
+url = $_fqdn:6554
```
CarlosNihelton, line 120: On the `host.url` line only we need to strip the `https://` out of the `_fqdn` per the experience we've had during the recent tests. Something like:

```suggestion
  Future<void> applyManualLandscapeConfig() async {
  final Uri uri = Uri.parse(_fqdn).replace(port: 6554);
    final config = '''
[host]
url = ${uri.authority}
```

Ref: https://api.flutter.dev/flutter/dart-core/Uri/authority.html

gui/packages/ubuntupro/lib/pages/landscape/landscape_page.dart
```diff
@@ -0,0 +1,316 @@
...
+    ).copyWith(
+      a: const TextStyle(
+        decoration: TextDecoration.underline,
+      ),
+    );
+
+    return ColumnLandingPage(
+      svgAsset: 'assets/Landscape-tag.svg',
+      title: 'Landscape',
+      onNext: !(model.fqdnError || model.fileError != FileError.none)
```
CarlosNihelton, line 39: This condition could be a getter inside the model:

```dart
bool get hasError => model.fqdnError || model.fileError != FileError.none;
```

Then you could rewrite this portion more easily:
```suggestion
      onNext: !model.hasError
```

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
```diff
@@ -0,0 +1,51 @@
...
+  bool? _isPurchaseAllowed;
+  SubscribeNowModel(this.client);
+
+  Future<SubscriptionInfo> applyProToken(ProToken token) async {
+    await client.applyProToken(token.value);
+    return client.subscriptionInfo();
+  }
+
+  void launchProWebPage() {
+    launchUrl(Uri.parse('https://ubuntu.com/pro'));
```
CarlosNihelton, line 21: A small enhancement since we're passing around.

```suggestion
    launchUrl(Uri.https('ubuntu.com/pro'));
```

gui/packages/ubuntupro/test/pages/landscape/landscape_model_test.dart
```diff
@@ -0,0 +1,99 @@
...
+@GenerateMocks([AgentApiClient])
+void main() {
+  group('landscape config', () {
+    const pluginChannel = MethodChannelP4wMsStore.methodChannel;
+    final pluginMessenger =
+        TestWidgetsFlutterBinding.ensureInitialized().defaultBinaryMessenger;
+    // Resets the plugin message handler after each test.
+    tearDown(() {
+      pluginMessenger.setMockMethodCallHandler(pluginChannel, null);
+    });
```
CarlosNihelton, line 20: Leftovers after a copy-paste? You don't need that plugin in these tests, so this can just be removed.

CarlosNihelton (changes requested):
We're almost there.

There are some formatting issues CI is complaining about and missing tests for the page, and some comments here and there.

gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_page.dart
```diff
@@ -39,19 +39,17 @@ class SubscriptionStatusPage extends StatelessWidget {
                 backgroundColor: YaruColors.red,
               ),
               onPressed: () async {
-                context.read<ValueNotifier<SubscriptionInfo>>().value =
-                    await model.detachPro();
+                await model.detachPro();
```
CarlosNihelton, line 42: One more thing I just noticed: For both the Store managed and the User managed status pages we need a `Configure Landscape` button that goes back specfically to the Landscape configure page.

ashuntu: Should the back button be removed from the page if you get to it that way?

ashuntu: For now, I went with adding a new route with a slightly modified Landscape page. If there is a better way to do this, I'd be happy to change the implementation.

CarlosNihelton: I'll review it more thoroughly but in the meantime did you consider using `Wizard.of(context).jump()` to go from the status pages back to the Landscape config page?

ashuntu: Every time I try to use `jump` I get errors that I believe mean that you cannot jump to previous pages, only next ones. At least, that's what I've experienced in my testing of it.

CarlosNihelton (changes requested):
We have a little more work to do. Let's pair as early as you can so I can help you with the integration test that's failing and the golangci-lint warnings plus a few other little things.

gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
```diff
@@ -0,0 +1,143 @@
...
+    validateFQDN();
+  }
+
+  String get fqdn => _fqdn;
+
+  set selected(LandscapeConfigType value) {
+    _selected = value;
+    _fqdnError = false;
+    _fileError = FileError.none;
+    validConfig();
```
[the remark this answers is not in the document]
ashuntu: The only issue is that if they enter, say, an invalid FQDN that shows an error, then select the file config instead, and then select quick setup again, the error is now gone unless they input something again. I suppose we could make it only validate on select if they've given input?

gui/packages/ubuntupro/lib/pages/landscape/landscape_page.dart
```diff
@@ -0,0 +1,340 @@
...
+      title: 'Landscape',
+      onNext: !model.hasError
+          ? () async {
+              if (await model.applyConfig() && context.mounted) {
+                onApplyConfig();
+              }
+            }
+          : null,
+      onBack: onBack ?? () => Wizard.of(context).back(),
+      onSkip: onSkip ?? () => Wizard.of(context).next(),
```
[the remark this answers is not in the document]
ashuntu: Making these tearoffs causes tests to fail, since there is an error that the widget does not have a Wizard parent it leads to a bunch of errors constructing the widget. Maybe this is something we can address in our meeting.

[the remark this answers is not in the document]
CarlosNihelton: Yeah, that could be fixed if you put the LandscapePage under a Wizard in tests, but I agree that we start to depend on many external things just for testing, which is not that good. Let's leave as is, for the sake of simpler tests.

ashuntu:
*Force-pushed to rebase with `main`*

CarlosNihelton (approved):
Phew! Good work!

PR #621, opened 2024-02-29
maint(wsl-pro-service): Bump wsl-pro-service version

EduardGomezEscandell:


EduardGomezEscandell:
Failed test fixed in #623

PR #622, opened 2024-02-29
fix(wsl-pro-service): Fix bug where IPv6 addresses where badly concatenated with port

EduardGomezEscandell:
Thanks @iosifache for noticing!

Co-authored-by: Didier Roche-Tolomelli <didrocks@ubuntu.com>


CarlosNihelton (approved):
Very good!

PR #623, opened 2024-02-29
fix(windows-agent): Fix race in ubuntupro tests

EduardGomezEscandell:
We were leaking the distro task processing. This caused a race in the removal of the distro's working dir.

See race here: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/8097649091/job/22129174184#step:7:72


didrocks (approved):
Let’s not fail on teardown randomly that way!

PR #626, opened 2024-03-01
feat(Landscape): Do not cancel commands in progress, and ping the server back on completion

EduardGomezEscandell:
The connection dropping should not stop a distro registration, for instance.
And for the server to know that a command worked, we should ping them when it has completed.

---

UDENG-2422
UDENG-2425

### DIFF
--- a/windows-agent/internal/proservices/landscape/connection.go
+++ b/windows-agent/internal/proservices/landscape/connection.go
@@ -218,7 +218,10 @@ func (conn *connection) receiveCommands(e executor) error {
 			return fmt.Errorf("could not receive commands: %v", err)
 		}
 
-		if err := e.exec(conn.ctx, command); err != nil {
+		// Removing the cancel context so that the command is executed even if the connection is lost.
+		ctx := context.WithoutCancel(conn.ctx)
+
+		if err := e.exec(ctx, command); err != nil {
 			log.Errorf(conn.ctx, "Landscape: %v", err)
 		}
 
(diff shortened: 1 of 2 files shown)


EduardGomezEscandell:
Uncovered lines cannot be asserted on as they simply log a warning.

windows-agent/internal/proservices/landscape/connection.go
```diff
@@ -218,9 +218,24 @@ func (conn *connection) receiveCommands(e executor) error {
 			return fmt.Errorf("could not receive commands: %v", err)
 		}
 
-		if err := e.exec(conn.ctx, command); err != nil {
+		// Removing the cancel context so that the command is executed even if the connection is lost.
+		ctx := context.WithoutCancel(conn.ctx)
```
didrocks, line 222: I keep forgetting about that API, this is really nice :)

Do you think we can have one test cancelling a context, and we pass that cancelled context here? (I don’t have enough "context" to see how deep we are in the stack)

didrocks (approved):
Sounds great and logical. Approving, but look at my comment if we can cancel the context before passing it.

windows-agent/internal/proservices/landscape/connection.go
EduardGomezEscandell: That'd be hard to exercise, line 108 right above has a cancelled context check.

didrocks (approved):
Re-approving then.

PR #628, opened 2024-03-01
feat(Landscape): Use new DefaultInstanceID message

EduardGomezEscandell:
Use the new datum introduced in https://github.com/canonical/landscape-hostagent-api/pull/14


windows-agent/internal/proservices/landscape/utils.go
```diff
@@ -84,10 +84,19 @@ func newHostAgentInfo(ctx context.Context, c serviceData) (info *landscapeapi.Ho
...
 	}
 
+	// Optional arguments
 	if conf.registrationKey != "" {
 		info.RegistrationKey = &conf.registrationKey
 	}
 
+	if defaultDistro, ok, err := gowsl.DefaultDistro(ctx); err != nil {
+		log.Warningf(ctx, "Landscape: could not get default distro: %v", err)
+		return info, nil
```
didrocks, line 94: isn’t there a case where your mock can return an error for this?

didrocks (commented):
I have one question as you can see about the tests.

It’s a good example on how to add a field in the current HostInfo. However, I’m wondering about the landscape mock too, you added a new field and return it’s value from `src`, but then, how is it updated with the set default instance call?

EduardGomezEscandell:
Ah, I mistakenly merged PR #690 into this branch 🤦. Anyhow, just review commit-per-commit I guess.

didrocks (approved):
Looks great with the additional test and the renamed variable! :)

PR #632, opened 2024-03-01
fix(wsl-pro-service): Validate port read from the address file

### LINKED ISSUE #629: Issue: Negative port may be returned by net.SplitHostPort in the Ubuntu service implementation.
### Is there an existing issue for this?

- [X] I have searched the existing issues and found none that matched mine

### Describe the issue

The latest version integrated the following changes in `wsl-pro-service/controlstream/controlstream.go`:

```go
-	fields := strings.Split(string(addr), ":")
-	if len(fields) == 0 {
-		// Avoid a panic. As far as I know, there is no way of triggering this,
-		// but we may as well protect against it.
-		return "", fmt.Errorf("could not extract port out of address %q", addr)
+	_, port, err := net.SplitHostPort(string(addr))
+	if err != nil {
+		return "", fmt.Errorf("could not parse port from %q: %v", addr, err)
 	}
-	port := fields[len(fields)-1]
```

The `net.SplitHostPort` function returns any string after `:`. For instance, `fmt.Println(net.SplitHostPort("127.0.0.1:-19"))` results in the string `127.0.0.1 -19 <nil>`. Given this, the returned port may be a string representing a negative value.

As no check is conducted in the `address` function, subsequent operations in the Ubuntu service may fail due to this unexpected negative port number. Please consider introducing an assertion for the port to be a positive value.

Many thanks!

### Steps to reproduce it

N/A

### Ubuntu System information

N/A

### Installed versions

`v0.1.2`

### Additional information

As discussed with @didrocks, this issue is sufficiently clear and doesn't require system logs. Hence, it was reported directly in this repository.

### Double check your logs

- [X] I have redacted any sensitive information from the logs

EduardGomezEscandell:
The port in the address file can be negative which would cause errors down the line. We also want to avoid port 0 as that usually means "any port" in go land.

Closes #629 

UDENG-2426

### DIFF
--- a/wsl-pro-service/internal/controlstream/controlstream_test.go
+++ b/wsl-pro-service/internal/controlstream/controlstream_test.go
@@ -218,17 +218,17 @@ func TestSend(t *testing.T) {
 	t.Parallel()
 
 	testCases := map[string]struct {
-		changePort string
+		forcePort string
 
 		wantErr bool
 	}{
-		"Success": {changePort: "FALSE"}, // Cannot use "" as default because we want to test the empty string
+		"Success": {forcePort: "-"},
 
-		"Error when port is wrong":        {changePort: "123", wantErr: true},
-		"Error when port is empty":        {changePort: "", wantErr: true},
-		"Error when port is not a number": {changePort: "abc", wantErr: true},
-		"Error when port is negative":     {changePort: "-1", wantErr: true},
-		"Error when port is zero":         {changePort: "0", wantErr: true},
+		"Error when port is wrong":        {forcePort: "123", wantErr: true},
+		"Error when port is empty":        {forcePort: "", wantErr: true},
+		"Error when port is not a number": {forcePort: "abc", wantErr: true},
+		"Error when port is negative":     {forcePort: "-1", wantErr: true},
+		"Error when port is zero":         {forcePort: "0", wantErr: true},
 	}
 
 	for name, tc := range testCases {
@@ -244,8 +244,8 @@ func TestSend(t *testing.T) {
 			_, agentMetaData := testutils.MockWindowsAgent(t, ctx, portFile)
 
 			// Override the port file with a different port
-			if tc.changePort != "FALSE" {
-				newAddr := fmt.Sprintf("127.0.0.1:%s", tc.changePort)
+			if tc.forcePort != "-" {
+				newAddr := fmt.Sprintf("127.0.0.1:%s", tc.forcePort)
 				err := os.WriteFile(portFile, []byte(newAddr), 0600)
 				require.NoError(t, err, "Setup: could not overwrite new address file")
 			}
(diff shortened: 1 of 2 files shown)


wsl-pro-service/internal/controlstream/controlstream_test.go
```diff
@@ -217,37 +217,68 @@ func TestConnect(t *testing.T) {
...
 	t.Parallel()
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	testCases := map[string]struct {
+		changePort string
+
+		wantErr bool
+	}{
+		"Success": {changePort: "FALSE"}, // Cannot use "" as default because we want to test the empty string
```
didrocks, line 225: we usually do `empty == "-"`, no?

wsl-pro-service/internal/controlstream/controlstream_test.go
```diff
@@ -217,37 +217,68 @@ func TestConnect(t *testing.T) {
...
 
-	require.Eventually(t, func() bool {
-		return agentMetaData.RecvCount.Load() > 1
-	}, 20*time.Second, time.Second, "The agent should have received another message via the control stream")
+			require.Eventually(t, func() bool {
+				return agentMetaData.RecvCount.Load() > 1
+			}, 20*time.Second, time.Second, "The agent should have received another message via the control stream")
 
-	require.Equal(t, int32(2), agentMetaData.RecvCount.Load(), "The agent should have received exactly two messages via the control stream")
+			require.Equal(t, int32(2), agentMetaData.RecvCount.Load(), "The agent should have received exactly two messages via the control stream")
```
didrocks, line 279: Argh, adding subtests makes the diff hard to read :p

wsl-pro-service/internal/controlstream/controlstream_test.go
```diff
@@ -217,37 +217,68 @@ func TestConnect(t *testing.T) {
 func TestSend(t *testing.T) {
 	t.Parallel()
 
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
+	testCases := map[string]struct {
+		changePort string
```
didrocks, line 221: `forcePort` maybe?

didrocks (changes requested):
This looks good to me, happy that this is now fixed! Did you check if there are other impacted places or this is the only one remaining?

Other than that, some small suggestions, tell me what you think!

wsl-pro-service/internal/controlstream/controlstream_test.go
EduardGomezEscandell: I don't think we have any standard yet in this repo (we defaulted to ""). Let's use "-" then.

EduardGomezEscandell:
> Did you check if there are other impacted places or this is the only one remaining?

This is the only place where we extract a port from an address. 
All other places with addresses:
- Landscape: we use the whole address verbatim
- Control stream: this fix
- WSL-side service: we pass a uint32 via gRPC and check for 0: https://github.com/canonical/ubuntu-pro-for-wsl/blob/e1b0699e09007dd897fd7bd49b303725f10f91ec/wsl-pro-service/internal/controlstream/controlstream.go#L107-L110

didrocks (approved):
Excellent!

PR #633, opened 2024-03-01
Update example `LandscapeConfig`

YanisaHS:
Updated to match the new example in the [Landscape Docs - How to configure Ubuntu Pro for WSL](https://ubuntu.com/landscape/docs/register-wsl-hosts-to-landscape).

Changes:
- `[host]` `url` shouldn't be sent with `https://`
- `ssl_public_key` must be included in the `[client]` section with the recent Landscape change for gRPC messages to use https


EduardGomezEscandell (approved):
Thank you!

PR #634, opened 2024-03-01
fix(gui): Regression: not checking subscription status on startup

CarlosNihelton:
For users it looks like the app forgotten the Pro token previously supplied. The current state is always starting in the subscribe now page. When refactoring the app code, we mistakenly placed the call to get the current subscription status in the last screen, it should be on the first.

Added an integration test to assert that behavior.

Also, notice the usage of `Wizard.of(context).replace()` instead of `.next()` or `.jump()`. It has the benefit of removing the current page from the navigator, thus allowing the `SubscriptionStatusPage` to consider itself the only one if the apps starts with a pre-existing subscription as well as prevents undesired back navigations (such as going back from the `SubscribeNowPage` to the `SubscriptionStatusPage` or the `StartupPage`).

### DIFF
--- a/gui/packages/ubuntupro/integration_test/ubuntu_pro_for_wsl_test.dart
+++ b/gui/packages/ubuntupro/integration_test/ubuntu_pro_for_wsl_test.dart
@@ -274,6 +274,58 @@ void main() {
           expect(find.byType(SnackBar), findsOneWidget);
         },
       );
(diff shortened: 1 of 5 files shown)


CarlosNihelton:
Let's see if rebasing fixes codecov.

ashuntu (approved):
I think this looks good. I tested locally and it fixes the issue

PR #637, opened 2024-03-04
fix(end-to-end): Fix log path in E2E testing

EduardGomezEscandell:
We moved the logs and forgot that the E2E tests look for them!


CarlosNihelton (approved):
LGTM 👍🏽

PR #645, opened 2024-03-04
feat(gui): Single instance app

CarlosNihelton:
There are many alternative ways to prevent launching more than one instance of the app, but making it so that an attempt activates the running window is more evolving, and requires some non-trivial Win32 code.

So I took the easy way. This PR leverages a third-party Windows-specific plugin to make the GUI a single-instance app. It's MIT licensed, as most of the Flutter ecosystem. Details can be found in https://pub.dev/packages/windows_single_instance.

Notice that we don't really care about which CLI arguments were passed.

Apart from the changes in the `main` function everything else is autogenerated by the Flutter tool as a consequence of having added that plugin.

### DIFF
--- a/gui/packages/ubuntupro/lib/main.dart
+++ b/gui/packages/ubuntupro/lib/main.dart
@@ -1,8 +1,13 @@
 import 'package:flutter/material.dart';
+import 'package:windows_single_instance/windows_single_instance.dart';
 import 'package:yaru_widgets/yaru_widgets.dart';
 import 'app.dart';
 
 Future<void> main() async {
   await YaruWindowTitleBar.ensureInitialized();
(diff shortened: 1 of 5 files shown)


ashuntu (approved):
Looks good, and works locally.

Would it be worth it to test for this in `integration_test/ubuntu_pro_for_wsl_test.dart`? I see we manually call the app's `main` function, maybe we could call `main` in the middle of a test and assert a new window isn't created?

CarlosNihelton:
I considered and reconsidered the test case, but this is one kind of scenario not directly reproduceable with just running a second `app.main()` right away. We could mock the plugin in some ways, but that resorts to some implementation details of the plugin that may turn the test brittle, as those details can change at any time without notice. So I think it's not worth investing in that.

PR #646, opened 2024-03-04
docs: update bug reporting template

jibel:


didrocks (approved):
Let’s iterate if we did anything that is unreadable :)

PR #647, opened 2024-03-04
feat(windows-agent): Obfuscate pro token when logged

ashuntu:
Previously, the pro token was being logged fully when applied. This PR obfuscates that token so that only the first and last 2 characters are visible.

---

UDENG-2354

### DIFF
--- a/windows-agent/internal/proservices/ui/ui.go
+++ b/windows-agent/internal/proservices/ui/ui.go
@@ -62,12 +62,12 @@ func (s *Service) ApplyProToken(ctx context.Context, info *agentapi.ProAttachInf
 		return nil, fmt.Errorf("some distros could not pro-attach: %v", err)
 	}
 
-	subs, err := s.getSubscriptionInfo()
+	subs, err := s.getSubscriptionSource()
 	if err != nil {
 		return subs, fmt.Errorf("could not assemble response: %v", err)
 	}
 
-	log.Debugf(ctx, "UI service: responding ApplyProToken with info: %v", info)
+	log.Debugf(ctx, "UI service: responding ApplyProToken with following info: %v", subs)
 	return subs, nil
 }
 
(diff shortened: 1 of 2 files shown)


windows-agent/internal/proservices/proservices_test.go
```diff
@@ -47,6 +53,10 @@ func TestNew(t *testing.T) {
 			reg := registry.NewMock()
 			k, err := reg.HKCUCreateKey("Software/Canonical/UbuntuPro")
 			require.NoError(t, err, "Setup: could not create Ubuntu Pro registry key")
+			if tc.createToken {
+				err := reg.WriteValue(k, "UbuntuProToken", defaultProToken, false)
```
didrocks, line 57: ok, you are writing the token here but the only assertion you are doing is that there is no error.

You don’t assert that:
- there is a token loaded when we got one in the registry
- the token itself is obfuscated (checking for `*` in the retrieved token.

didrocks (changes requested):
Thanks for looking into this and adding a test for it! My few notes is on the technical side itself with where things could be improved.

Your changes matches the card description itself which may not have been accurate. IIRC, we have one place where the Windows agent is logging the token in plain text and that was what supposed to be replaced. I don’t think that loading it on `New` is required with the event mechanism that Edu put in place, but I will let @EduardGomezEscandell  comment on that. However, conceptually, we are loading the token from the registry, this is the place where there is a change that it should be logged (apart from the error handling of course) and obfuscated.



See my remark inline about the

CarlosNihelton (approved):
Looking at both agent's and wsl-pro-service logs I didn't spot any other place where we'd log the pro token verbatim.

So, LGTM!

didrocks (approved):
Excellent!

Do you mind renaming the PR title itself (and the description), as it’s not accurate anymore and will be used in the merge commit? Then, feel free to merge :)

@EduardGomezEscandell: FYI, we took the liberty to rename `getSubscriptionInfo` to `getSubscriptionSource`, because this methods returns an info object, but only with the source field set. I was undecided in between this and returning only the source and wrapping it again afterward, but let’s start with that. At least, this is a little bit more clear now that we don’t return the token from it!

PR #651, opened 2024-03-05
fix(gui): Promotes the "Confirm" button

CarlosNihelton:
Team's feedback was that the button was not looking like a Yaru styled button. It was, but a `TextButton` which is a low emphasis component. Upon discussion, we settled with the higher emphasis `ElevatedButton`, which is even higher in Yaru.

So this removes any further styling, convert the button to `ElevatedButton` and detaches it from the text input field. No functionality changes, yet some tests required updates in how they find that button.

image

UDENG-2376

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
+++ b/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
@@ -77,12 +77,8 @@ class _ProTokenInputFieldState extends State<ProTokenInputField> {
             hintText: lang.tokenInputHint,
             errorText: _token.errorOrNull?.localize(lang),
             counterText: '',
-            suffixIcon: TextButton(
+            suffixIcon: ElevatedButton(
               onPressed: canSubmit ? _handleApplyButton : null,
-              style: TextButton.styleFrom(
-                // allows the suffix button to stand out when enabled while keeping its custom look.
-                foregroundColor: Theme.of(context).colorScheme.onSurface,
-              ),
               child: Text(lang.confirm),
             ),
           ),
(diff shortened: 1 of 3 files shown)


ashuntu (changes requested):
The button's sharp corners looks a bit strange when it's not clickable, see this image:

image

In my personal opinion, the sharp button looks a bit out of place. When it has invalid input there's an outline around the whole text field and on top of the button, but when it's valid then the button is on top of the rounded text field. Maybe it would be better to put the button to the right of the text field so the text field doesn't take up the full width (like the file selector on the Landscape page), or put the button below the text field?

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
```diff
@@ -70,24 +70,29 @@ class _ProTokenInputFieldState extends State<ProTokenInputField> {
...
-          decoration: InputDecoration(
-            hintText: lang.tokenInputHint,
-            errorText: _token.errorOrNull?.localize(lang),
-            counterText: '',
-            suffixIcon: TextButton(
-              onPressed: canSubmit ? _handleApplyButton : null,
-              style: TextButton.styleFrom(
-                // allows the suffix button to stand out when enabled while keeping its custom look.
-                foregroundColor: Theme.of(context).colorScheme.onSurface,
+        builder: (context, _, __) => Row(
```
ashuntu, line 73: Looks like the button gets pushed down when there's an error present:
image

Probably need `CrossAxisAlignment.start` on the `Row`.

ashuntu (changes requested):

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
CarlosNihelton: Well spotted.

CarlosNihelton: I clearly undersestimate this task and now even Go vulnerability reports are in my way.

ashuntu (approved):
Aside from some of the CI failing right now, looks good!

CarlosNihelton:
Rebasing to allow CI to pass. No changes in the code.

CarlosNihelton:
What a button!

Feichtmeier:
Hi 👋 
We've consolidated yaru theme, widgets and icons into yaru.dart (4.0.0+)
Also updated the web app to show everything:
https://ubuntu.github.io/yaru.dart/
best regards

PR #669, opened 2024-03-06
maint(go): Go 1.22.0

didrocks:
Update all modules and code to Go 1.22.0
Adapt some code to avoid suttering on `foo := foo`
Prefer 1.22.1 to fix security vulnerability

UDENG-2352


windows-agent/internal/proservices/wslinstance/wslinstance_test.go
```diff
@@ -95,7 +95,7 @@ func TestConnected(t *testing.T) {
 		ctx = wsl.WithMock(ctx, wslmock.New())
 	}
 
-	distroName, _ := wsltestutils.RegisterDistro(t, ctx, false)
+	defaultDistroName, _ := wsltestutils.RegisterDistro(t, ctx, false)
```
CarlosNihelton, line 98: This is a good change. Makes it clear that we are not copying the variable a few lines below just because of the old for-loop scope behavior.

EduardGomezEscandell (approved):
Thanks for dealing with this :)

CarlosNihelton (approved):
LGTM 👍🏽 

I'm looking at codecov reports but still don't get why coverage is reported to decrease so much.

windows-agent/internal/proservices/wslinstance/wslinstance_test.go
didrocks: Yeah! I prefer we didn’t override it and tell at some point "oh, we forgot this" :p

PR #672, opened 2024-03-06
fix(windows-agent): Truncate log files

ashuntu:
This PR renames any existing log to `log.old` when the program starts, overwriting the file if it already exists. It also tests that if a log exists a log.old is successfully created.

---

UDENG-2355

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/main_test.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/main_test.go
@@ -45,17 +45,23 @@ func TestRun(t *testing.T) {
 	t.Parallel()
 
 	tests := map[string]struct {
-		createLog        bool
+		existingLogContent string
+
 		runError         bool
 		usageErrorReturn bool
 		logDirError      bool
 
 		wantReturnCode int
+		wantOldLogFile bool
 	}{
 		"Run and exit successfully":                                {},
 		"Run and exit successfully despite logs not being written": {logDirError: true},
-		"Run and exit successfully when logs already exist":        {createLog: true},
 
+		// Log file handling
+		"Existing log file has been renamed to old": {existingLogContent: "foo", wantOldLogFile: true},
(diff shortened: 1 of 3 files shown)


ashuntu:
Seems like the test is failing only on a Linux host, it passes on my Windows machine. It looks like if I run it on Linux it treats the log directory as the current directory and incorrectly creates logs there. I'm not sure what the best approach is to fix this.

windows-agent/cmd/ubuntu-pro-agent/main_test.go
```diff
@@ -42,16 +44,26 @@ func (a *myApp) PublicDir() (string, error) {
...
 
-		wantReturnCode int
+		wantReturnCode        int
+		wantOldLogFileContent *string
 	}{
 		"Run and exit successfully":                                {},
 		"Run and exit successfully despite logs not being written": {logDirError: true},
 
+		// Log file handling
+		"Existing log file has been renamed to old": {existingLogContent: "foo", wantOldLogFileContent: &fooContent},
```
didrocks, line 63: I would still keep the test case with the empty file, to document explicitely this behaviour on the edge case now that the code has been modified (the fact that we rename it, and the old log file content is empty then).

didrocks (commented):
I know that you kept it as draft (and that the CI is failing suspiciously on:

```
--- FAIL: TestRun (0.00s)
    --- FAIL: TestRun/Run_and_exit_successfully_despite_logs_not_being_written (0.10s)
        main_test.go:123: 
            	Error Trace:	/home/runner/work/ubuntu-pro-for-wsl/ubuntu-pro-for-wsl/windows-agent/cmd/ubuntu-pro-agent/main_test.go:123
            	Error:      	file "log.old" exists
            	Test:       	TestRun/Run_and_exit_successfully_despite_logs_not_being_written
            	Messages:   	Old log file should not exist
```

If we want us to have a look together at the start of your day, do not hesitate!

I have an additional comment on the test itself, but code-wise, this looks good :)

didrocks (approved):
Excellent work!

PR #675, opened 2024-03-07
maint(wsl-pro-service): Better logging of subprocess errors in WSL-Pro-Service

EduardGomezEscandell:
We were not printing Stderr when something failed, meaning that we lost most of the error message.


didrocks (approved):
No-brainer, thanks for looking at them!

PR #679, opened 2024-03-07
feat(gui): Status bar containing app version, bug report link and agent status

CarlosNihelton:
This looks bigger than it really is, but it's still big. Better digest it commit by commit, as I put effort in telling the whole story there.

There are a few small tests I sneaked into this PR because it was already looking big anyway.

The intent of this PR is to add an always present status bar to the GUI featuring:
- the app version
- a button to report bugs in GH
- a button to restart the agent, if down.

Like this:

ubuntupro_fDbAo46bi7

The changes are quite invasive because the status bar is always present, and requires allowing the agent to be restarted at any moment during the app lifecycle.

When reviewing this, have in mind that:

- The first 4 commits are test refactorings (either the fixes I mentioned above or preparation for what is coming)
- Commits 5 to 12 are refactorings over AgentApiClient and AgentStartupMonitor. They are necessary because:

  - The AgentApiClient must remain the same object accross the app (since its registered as a global service)
  -  The AgentApiClient must support connecting to a different port at any time.
  -  The AgentApiClient must expose the gRPC connectivity status change events so we can report a disconnect event (notice it doesn't say it's connected, it only reports changes).
  - Automatic reconnection on drop is out of scope because the agent doesn't crash. If that happens, it's a real issue. We don't want to stay on a hot loop restarting it forever.
  - The AgentStartupMonitor was originally designed to only exist in the scope of the startup page, never accessible outside of it.
  - The AgentStartupMonitor was meant to notify of a succesful agent startup only those subscribed at its creation time, no way to add subscribers ad hoc.
  - The AgentStartupMonitor didn't care to have a reference to the AgentApiClient since upon a succesful startup it would be out of the game.

- The subsequent commits are about the meat of this PR: 

  - adding the AgentConnection ChangeNotifier capable of reporting a connection loss and restarting the agent (not by itself, but reusing the AgentStartupMonitor) 
  - and a StatusBar consuming this ChangeNotifier to report the app version, and featuring a button to open an issue in GitHub and to restart the agent if down.

There is one caveat: the StartupErrorWidget, only shown if the agent cannot start to begin with, turns the button to restart the agent meaningless, so only for that widget, we hide that button, like this:

image


ashuntu (approved):
Thanks for making the changes I suggested; I think this looks good!

(And thanks for figuring out the screen sizing issue during tests, that was a headache :))

CarlosNihelton:
end-to-end test failure is being handled in PR #682

CarlosNihelton:
Rebasing on main after #682 was merged, no changes in code.

PR #682, opened 2024-03-08
fix(wsl-pro-service): Don't break when processes print to Stderr

EduardGomezEscandell:
cmd.exe prints random stuff to stderr. When attempting to parse the output of `cmd.exe`, Stderr gets in the way and we fail.

Regression introduced in 373a97d

---

UDENG-2476

### DIFF
--- a/wsl-pro-service/internal/system/backend.go
+++ b/wsl-pro-service/internal/system/backend.go
@@ -1,7 +1,9 @@
 package system
 
 import (
+	"context"
 	"os"
+	"os/exec"
 	"path/filepath"
 )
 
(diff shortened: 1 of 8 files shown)


wsl-pro-service/internal/system/backend.go
```diff
@@ -22,24 +24,29 @@ func (b realBackend) GetenvWslDistroName() string {
...
 // WslinfoExecutable returns the full command to run the wslinfo executable with the provided arguments.
-func (b realBackend) WslinfoExecutable(args ...string) (string, []string) {
-	return "wslinfo", args
+func (b realBackend) WslinfoExecutable(ctx context.Context, args ...string) *exec.Cmd {
+	return exec.CommandContext(ctx, "wslinfo", args...)
 }
 
-func (b realBackend) CmdExe(path string, args ...string) (string, []string) {
-	return path, args
+func (b realBackend) CmdExe(ctx context.Context, path string, args ...string) *exec.Cmd {
```
CarlosNihelton, line 45: Most of this file is inherently uncoverable by package tests, but I guess we could add a small test function in `system_test.go` that asserts the returned `exec.Cmd`object's Dir respects the invariant herein proposed. I'm pretty sure that small test will be enough to please codecov.

CarlosNihelton (approved):
Thanks for addressing this bug so well!

PR #683, opened 2024-03-11
feat(wsl-pro-service): Add default wsl tag to Landscape configs

ashuntu:
This PR extends the already-existing modifications we make to Landscape configs in `wsl-pro-service`. It appends `tags = wsl` to the Landscape config if a `tags` key doesn't already exist. Any `tags` already present in the config (including empty tags) are *not* overridden.

*For review, the [Landscape example configuration](https://github.com/canonical/landscape-client/blob/master/example.conf#L163-L167) may be useful.*

---

UDENG-2372


wsl-pro-service/internal/system/testdata/TestLandscapeEnable/golden/success_when_empty_tags_are_present
ashuntu, line 1: This weird diff is a side effect of me deleting an unused golden file

ashuntu:
Codecov is complaining about lines that were also previously not covered. Everything added is covered to the same degree as was previously.

EduardGomezEscandell (approved):
Great!

CarlosNihelton (approved):
It's hard to break that `createKey` function because most of the things we could do in test environment to break it would break before reaching it.

wsl-pro-service/internal/system/system_test.go
```diff
@@ -399,6 +399,8 @@ func TestLandscapeEnable(t *testing.T) {
 	}{
 		"Success":                                    {},
 		"Success overriding computer_title":          {},
+		"Success when empty tags are present":        {},
+		"Success when custom tags are present":       {},
```
didrocks, line 403: nitpick: I would change the name here to be more explicit than just "Success", like "Do not append wsl when config tag is provided but empty" and "Do not append wsl tag when config tag is provided"

didrocks (changes requested):
I do have one small request in test name, but otherwise, excellent, thanks for looking at the fixtures and golden files!

PR #685, opened 2024-03-12
refactor(common): Some gardening for the common package 🌱 

EduardGomezEscandell:
A wise man once said:
> hey, when you are in mood for some gardening, I think there are quite some opportunities:
> - common/golden -> can it be moved to common/testutils?
> - common/utils.go -> I think Obfuscate can win having some tests
> - common/utils.go -> FindWorkspaceRoot can probably be moved in common/testutils too
>
> Basically, all this guided by looking at the coverage :slightly_smiling_face:

--- 

UDENG-2252


EduardGomezEscandell:
> cannot use -fuzz flag with multiple packages

😭

PR #686, opened 2024-03-12
fix(docs): Clean-up section in the tutorial

CarlosNihelton:
This removes a TODO item from the tutorial document, by adding a clean-up section that removes the Ubuntu WSL apps and instances and optionally the WSL app as well.

I'm intentionally skipping cleaning-up the registry because all references to the registry will be replaced by the GUI, so no reason to add dead content.

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -60,6 +60,7 @@ All set. From now on you can use it to launch  WSL instances.
 
 ### Install the `Ubuntu 22.04 LTS` and `Ubuntu (Preview)` applications
 
+(ref::backup-warning)=
 ```{warning}
 **If you already have them pre-installed:** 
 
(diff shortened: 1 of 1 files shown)


EduardGomezEscandell (changes requested):
Looks like something is up with the image import

https://readthedocs.com/projects/canonical-ubuntu-pro-for-wsl/builds/2078639/


Besides this, should we have a section about removing the `%UserProfile%\.ubuntupro` directory?

CarlosNihelton:
> Looks like something is up with the image import
> 
> https://readthedocs.com/projects/canonical-ubuntu-pro-for-wsl/builds/2078639/
 
Where is my screenshot? 🙈 

> Besides this, should we have a section about removing the `%UserProfile%\.ubuntupro` directory?
Fair point. I wanted to avoid exposing that implementation detail, but it can take us a while to land the clean-up verb in the agent.

EduardGomezEscandell (approved):

PR #687, opened 2024-03-12
deps(wsl-pro-service): update the package name for ubuntu-advantage-tools

CalvoM:
From v31, ubuntu-advantage-tools has been renamed to ubuntu-pro-client. The ubuntu-advantage-tools package name is now a transitional package. Noble currently has ubuntu-advantage-tools v31.


didrocks (approved):
Thanks for the changes!

Ignoring the Azure login V2 error on the end to end tests and I checked that indeed, the new package is available in every releases we support.

PR #688, opened 2024-03-12
maint(gui): Upgrade yaru, remove deprecated child packages and fixes imports.

CarlosNihelton:
The community deprecated yaru_icons and yaru_widgets, merging them into yaru. So now we must only depend on yaru and remove the deprecated dependencies from our `pubspec.yaml`. Since the theme, icons and widgets are in the same package now, some imports need fixes.


ashuntu (approved):
Looks good

PR #690, opened 2024-03-13
fix(windows-agent): Avoid panic in distro properties

EduardGomezEscandell:
This PR fixes a race where unregistering a distro at a very unfortunate time would cause a panic.

From [LBYL](https://docs.python.org/3.5/glossary.html#term-lbyl) to [EAFP](https://docs.python.org/3.5/glossary.html#term-eafp) 😃

---

UDENG-2385

### DIFF
--- a/common/go.mod
+++ b/common/go.mod
@@ -10,7 +10,7 @@ require (
 	github.com/sirupsen/logrus v1.9.3
 	github.com/snapcore/go-gettext v0.0.0-20201130093759-38740d1bd3d2
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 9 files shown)


EduardGomezEscandell:
For some reason CI is not running 🤷

EduardGomezEscandell:
Force-pushed to trigger CI, no changes.

EduardGomezEscandell:
Regarding codecov: I don't think we need to test a panic 🙃

didrocks (approved):
With current tests still passing, and I agree there is no need to test the panic, so +1! Good work :)

PR #694, opened 2024-03-15
feat(tests): change golden files update flag

didrocks:
This will make it clear this is scope for tests, so that we don’t mix with other potential env available. Also, have it as verb/action rather than passive form.
We are going to use this in other repos.


EduardGomezEscandell (approved):
```powershell
${env:GITHUB_APPROVE_REQUEST} = "1"
```

didrocks:
> ```powershell
> ${env:GITHUB_APPROVE_REQUEST} = "1"
> ```

This is criminal! :)

PR #698, opened 2024-03-20
fix(end-to-end): Increase timeouts in pro attachment check

EduardGomezEscandell:
sometimes pro attach is very slow


CarlosNihelton (approved):
Let's see how tests perform after this.

PR #700, opened 2024-03-21
fix(landscape): Restrict landscape config file from 0604 to 0640

EduardGomezEscandell:
The landscape group needs read-access to the /etc/landscape/client.conf file. However, we do not want any user to be able to read it as it may contain confidential details such as the registration key. The solution is to restrict its visibility to its owner and group, and to add it to landscape's group

Thanks @iosifache for spotting this vulnerability.

---

UDENG-2540

### DIFF
--- a/wsl-pro-service/internal/system/system.go
+++ b/wsl-pro-service/internal/system/system.go
@@ -279,8 +279,8 @@ func (s *System) findCmdExe() (cmdExe string, err error) {
 	return "", fmt.Errorf("none of the mounted drives contains subpath %s", subPath)
 }
 
-// lookupGroup searches the group with the specified name and returns its GID.
-func (s *System) lookupGroup(name string) (int, error) {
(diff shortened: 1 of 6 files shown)


wsl-pro-service/internal/system/system.go
```diff
@@ -274,3 +278,33 @@ func (s *System) findCmdExe() (cmdExe string, err error) {
 
 	return "", fmt.Errorf("none of the mounted drives contains subpath %s", subPath)
 }
+
+// lookupGroup searches the group with the specified name and returns its GID.
+func (s *System) lookupGroup(name string) (int, error) {
```
didrocks, line 283: I would rename to something more explicit, like groupToGID as it doesn’t have the same signature than the rest of the lookup functions in go

didrocks (approved):
A small naming nitpick, otherwise, very good!

didrocks (approved):
Thanks!

PR #701, opened 2024-03-25
fix(gui): New icon

CarlosNihelton:
We got a new icon from Design. The original SVG is in GDrive, and also added in this repository for easy reference.

Last time we updated visual assets we talked about using symlinks for the `.ico` files around to prevent misses.
Symlinks on Windows are still tricky to this date. Git support is fairly new and must be enabled manually. The Flutter tool seems not to be able to treat symlinks resulting in failed builds. Creating a SymbolicLink type on Windows requires elevation as administrator or Developer Mode enabled.

So, instead of using symlinks, this PR makes sure we treat the MSIX Images directory as the source of truth, gitignoring the GUI `.ico` file inside the resources directory and relying on the build system to copy the correct icon on every build (which was already the case, but we had a TODO around which icon should be treated as the source of truth, now fixed).

This PR also removes one unused image asset (which is also not regenerated by VS) from the packaging project.

Closes UDENG-2466.


EduardGomezEscandell (approved):
Looks good 😃 

 Make sure to run the E2E tests (or the build at least).

PR #703, opened 2024-03-26
fix(windows-agent): Avoid leaking proservices Manager on app teardown

EduardGomezEscandell:
When we call agent.Quit, we don't wait for the proServices to perform their cleanup. This means that the process can exit while the Landscape, registry watcher or database are in the middle of an operation.

This PR fixes this, somewhat: It guarantees to exit the process after proservices.Stop returns. We trust that the implementation of the teardown of these proservices is implemented properly (I gave it a quick glance and it seems to be the case).

This problem also affected tests, as the proServices could be writing to disk while t.Cleanup tries to delete the temp directory, causing a race and then a test error.

---

Example test failure due to this: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/8436243216/job/23103334236?pr=702#step:7:63

Here, the proservices manager is still writing logs while we try to remove the log file, even though we already called Quit on the app.


didrocks (approved):
Nicely spotted!

PR #708, opened 2024-03-28
feat(windows-agent): Add support for importing a rootfs

ashuntu:
Added support for importing a root filesystem from a URL, as recently [supported by GoWSL](https://github.com/ubuntu/GoWSL/pull/108).

Importing will download and import the filesystem from a URL, assuming it's a correct tarball. If a checksum is provided, it will also be checked for a match before importing.

---

UDENG-2554


windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +197,126 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return err
+	}
+	gotChecksum := base64.StdEncoding.EncodeToString(h.Sum(nil))
```
didrocks, line 312: I think this should be rather `fmt.Printf("%x", h.Sum(nil))`

CarlosNihelton: `fmt.Sprintf` you mean, right?

ashuntu: I think `fmt.Sprintf("%x", h.Sum(nil))` is going to give a base 16 string, but we need a base 64 string since that's what `wantChecksum` is.

CarlosNihelton: We thought we needed, but the slice of bytes returned by the checksum function is already a hex-string, so we are just fine to output it with `%x`. We don't need the base64 encoding handling as we originally though.

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -223,6 +265,37 @@ func TestInstall(t *testing.T) {
 	}
 }
 
+func mockRootfsFileServer(t *testing.T, ctx context.Context) string {
```
CarlosNihelton, line 268:
```suggestion
//nolint:revive // Context goes after testing.T
func mockRootfsFileServer(t *testing.T, ctx context.Context) string {
```
You'll see comments like those a lot above test helper functions.

windows-agent/internal/proservices/landscape/service.go
```diff
@@ -81,9 +81,9 @@ func New(ctx context.Context, conf Config, db *database.DistroDB, args ...Option
 	}
 
 	if opts.homedir == "" {
-		homeDir := os.Getenv("UserProfile")
-		if homeDir == "" {
-			return nil, errors.New("could not locate the user profile dir: %UserProfile% is not set")
+		homeDir, err := os.UserHomeDir()
+		if homeDir == "" || err != nil {
+			return nil, errors.New("could not locate the user home dir")
```
CarlosNihelton, line 86: If `homeDir` would come empty an errror is returned instead, so checking the `err` variable is enough.

ashuntu:
*Continued in https://github.com/canonical/ubuntu-pro-for-wsl/pull/718*

PR #718, opened 2024-04-04
feat(windows-agent): Add support for importing a rootfs

ashuntu:
Added support for importing a root filesystem from a URL, as recently [supported by GoWSL](https://github.com/ubuntu/GoWSL/pull/108).

Importing will download and import the filesystem from a URL, assuming it's a correct tarball. If a checksum is provided, it will also be checked for a match before importing.

*This is a continuation of https://github.com/canonical/ubuntu-pro-for-wsl/pull/708 since the branch needed to be rebased onto `main`, and I could not force push.*

----

UDENG-2554


windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+		return err
+	}
+
+	return nil
+}
+
+func installFromURL(ctx context.Context, homeDir string, distro gowsl.Distro, rootfs *landscapeapi.Command_Install_Rootfs) (err error) {
+	// Since we are going to remove the tarball soon, I'd say it's fine to write it to a subfolder of the current working directory,
+	// as it will be somewhere inside the Windows Apps private folders:
+	// %USERPROFILE%\AppData\Local\Packages\CanonicalGroupLimited.Ubuntu_79rhkp1fndgsc\LocalState\
```
CarlosNihelton, line 215:
```suggestion
	// %USERPROFILE%\AppData\Local\Packages\CanonicalGroupLimited.UbuntuPro_79rhkp1fndgsc\LocalState\
```

When I wrote that comment I didn't mean it to stay, but I guess it's useful, so let's just correct this path.

CarlosNihelton (approved):
One last nitpick, otherwise I think it's good to go.

Let's see if the last test suggestion helps completing the code coverage target, if not, we'll get it

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
 func (e executor) shutdownHost(ctx context.Context, cmd *landscapeapi.Command_ShutdownHost) error {
 	return gowsl.Shutdown(ctx)
 }
+
+func installFromMicrosoftStore(ctx context.Context, distro gowsl.Distro) error {
```
didrocks, line 200: I would defer decore.OnError so that we prewrap with `can’t install from Microsoft Store:`

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+		return err
+	}
+
+	return nil
+}
+
+func installFromURL(ctx context.Context, homeDir string, distro gowsl.Distro, rootfs *landscapeapi.Command_Install_Rootfs) (err error) {
+	// Since we are going to remove the tarball soon, I'd say it's fine to write it to a subfolder of the current working directory,
+	// as it will be somewhere inside the Windows Apps private folders:
+	// %USERPROFILE%\AppData\Local\Packages\CanonicalGroupLimited.UbuntuPro_79rhkp1fndgsc\LocalState\
```
didrocks, line 215: ditto, but I would defer decore.OnError so that we prewrap with `can’t install from URL %q:`

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	}
+
+	// Create the directory that will contain the vhdx
+	vhdxDir := filepath.Join(homeDir, "WSL", distro.Name())
+	if err := os.MkdirAll(vhdxDir, 0700); err != nil {
+		return err
+	}
+
+	if _, err := gowsl.Import(ctx, distro.Name(), tarball, vhdxDir); err != nil {
+		_ = os.RemoveAll(vhdxDir)
```
didrocks, line 251: I would still capture this error and log it as a warning to have a trace we have a leftover on disk.

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+		return err
+	}
+	return nil
+}
+
+func download(w io.Writer, url string) (err error) {
+	defer func() {
```
didrocks, line 258: this is what our decorate package provides in a single line :)

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+		if err != nil {
+			err = fmt.Errorf("could not download %q: %v", url, err)
+		}
+	}()
+
+	//nolint:gosec // G107 expects url to be const, but we are reading it from Landscape.
```
didrocks, line 264: don’t tell what G107 is avoid, just annotate that we are ignoring G107 because we are receiving the url fom Landscape.

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+
+	if _, err := io.Copy(w, resp.Body); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checksumMatches(path, wantChecksum string) (err error) {
+	defer func() {
```
didrocks, line 283: ditto

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	}
+
+	if _, err := io.Copy(w, resp.Body); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func checksumMatches(path, wantChecksum string) (err error) {
```
didrocks, line 282: I would return 2 values here: (bool, err)

bool is the returned value if the checksum matches, and error only in the internal (like reading the file)

Also, I would instead of taking a path, use an io.Reader here, but see check my suggestion before :)

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	if err != nil {
+		return err
+	}
+	defer f.Close()
+
+	if err := download(f, rootfs.GetUrl()); err != nil {
+		return err
+	}
+
+	if err := checksumMatches(tarball, rootfs.GetSha256Sum()); err != nil {
```
didrocks, line 240: So, here, you are reading the file twice: once to download/copy it to disk and another one for the checksuming.

I would suggest using a `io.TeeReader` to perform both at the same time (https://pkg.go.dev/io#TeeReader). That way, we only read it once, write to disk and also pass it to the checksumer :)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -142,21 +146,46 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
 	testCases := map[string]struct {
-		distroAlredyInstalled bool
-		emptyDistroName       bool
-		wslInstallErr         bool
-		appxDoesNotExist      bool
+		distroAlreadyInstalled bool
+		emptyDistroName        bool
+		wslInstallErr          bool
+		appxDoesNotExist       bool
+		// to improve testing coverage
```
didrocks, line 163: Is this comment needed? We do already have special case before to improve the coverage :)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -142,21 +146,46 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
+		breakTarballCreation bool
+
+		sendRootfsURL      string
+		sendRootfsChecksum string
 
 		wantInstalled   bool
 		wantNonRootUser bool
 	}{
-		"Success": {wantInstalled: true, wantNonRootUser: true},
+		"Success":                              {wantInstalled: true, wantNonRootUser: true},
```
didrocks, line 174: `From the store` (as the test is prefix with `TestInstall`)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -142,21 +146,46 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
+		"Success with a rootfs":                {sendRootfsURL: "goodfile", wantInstalled: true},
```
didrocks, line 175: `From a rootfs url` (and you get the idea for the other test naming)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -142,21 +146,46 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
+		// to improve testing coverage
+		nonResponsiveServer  bool
+		breakVhdxDir         bool
+		breakTarballCreation bool
+
+		sendRootfsURL      string
+		sendRootfsChecksum string
 
 		wantInstalled   bool
 		wantNonRootUser bool
```
didrocks, line 172: I see this one is only use when downloading from the Store, I’m a little bit puzzled in its usage, mind looking at it?

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -223,6 +291,39 @@ func TestInstall(t *testing.T) {
...
+		//nolint:gosec // For the sake of simplicity and readability.
+		err := http.Serve(lis, mux)
+		if err != nil {
+			t.Logf("mockRootfsFileServer: serve error: %v", err)
+		}
+	}()
+
+	addr := "http://" + lis.Addr().String()
+	t.Logf("Serving on %s", addr)
+	return addr
```
didrocks, line 324: we don’t return a way to teardown the server, do we? (closing `lis` connection). I think we should, even if this is in tests.

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -223,6 +291,39 @@ func TestInstall(t *testing.T) {
 	}
 }
 
+//nolint:revive // Context goes after testing.T
+func mockRootfsFileServer(t *testing.T, ctx context.Context) string {
+	t.Helper()
+
+	mux := http.NewServeMux()
+
+	mux.HandleFunc("/goodfile", func(w http.ResponseWriter, r *http.Request) {}) // Return empty file
```
didrocks, line 300: I would use the 1.22 syntax: `"GET /goodfile"` (if I’m correct) to restrict to what we support.

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -223,6 +291,39 @@ func TestInstall(t *testing.T) {
...
+	})
+	mux.HandleFunc("/badresponse", func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusNotFound)
+	})
+
+	lis, err := (&net.ListenConfig{}).Listen(ctx, "tcp", "localhost:")
+	require.NoError(t, err, "Setup: mockRootfsFileServer could not listen")
+
+	go func() {
+		//nolint:gosec // For the sake of simplicity and readability.
```
didrocks, line 315: unsure what you are linting on, I thin the comment should be more descriptive :)

windows-agent/internal/proservices/landscape/landscape_test.go
```diff
@@ -51,6 +51,40 @@ account_name = testuser
...
+		require.NoError(t, err, "Setup: database New should not return an error")
+
+		// Note that these tests cannot be run in parallel due to manipulating
+		// these test environment variables
+		if tc.breakHome {
+			t.Setenv("UserProfile", "")
+			t.Setenv("HOME", "")
+		}
+
+		_, err = landscape.New(ctx, conf, db)
```
didrocks, line 78: you can still get the object and assertion after the `NoError` case that it’s not empty.

didrocks (changes requested):
Nice! I see you exercised a lot of patterns in this case. I have some suggestions.

The biggest will be to use the TeeReader to avoid reading the downloaded file twice (especially as we are going to import it anyway). The rest is more on idiomatic pattern and reusing our error decoration to get more meaningful errors when encountered.

Tell me if you need any help on anything!

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +196,119 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	// as it will be somewhere inside the Windows Apps private folders:
+	// %USERPROFILE%\AppData\Local\Packages\CanonicalGroupLimited.UbuntuPro_79rhkp1fndgsc\LocalState\
+	cwd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+
+	tmpDir := filepath.Join(cwd, distro.Name())
+	if err := os.MkdirAll(tmpDir, 0700); err != nil {
+		return err
```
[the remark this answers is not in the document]
ashuntu: I'm not sure what you mean here, but take a look at my recent change and see if that's ok.

windows-agent/internal/proservices/landscape/executor_test.go
ashuntu: With the other tests it still seems like they have a nonroot user anyway, so I'm not sure.

windows-agent/internal/proservices/landscape/executor.go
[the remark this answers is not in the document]
CarlosNihelton: There is a test case in which we break the creation of the tarball by creating a directory with the same name as the file that would be created, causing this function to fail, bailing out in line 229.
Didier suggested us to add another test case in which, instead of breaking `os.Create(tarball)` we'd break `os.MkDirAll(tmpDir,...` by creating a file with the same name, thus causing the line 220 to be covered.

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: I guess we could remove the assertion about whether the default user is root or not, since it's only relevant in one test case, which is related to installing from the MS Store.

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -189,3 +197,101 @@ func (e executor) setDefault(ctx context.Context, cmd *landscapeapi.Command_SetD
...
+	if err := os.MkdirAll(vhdxDir, 0700); err != nil {
+		return err
+	}
+
+	if _, err := gowsl.Import(ctx, distro.Name(), tarball, vhdxDir); err != nil {
+		err = os.RemoveAll(vhdxDir)
+		if err != nil {
+			log.Warningf(ctx, "could not cleanup install directory: %v", err)
+		}
+		return err
```
CarlosNihelton, line 249: Now if both `gowsl.Import` and `os.RemoveAll` fail, we'll loose the error information on the first. I'd rather name the second err variable to something else and return `errors.Join`

ashuntu: Ah yeah, an oversight on my part. Good spot :)

windows-agent/internal/proservices/landscape/executor.go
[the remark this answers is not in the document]
ashuntu: Oh ok, I was misreading the original message. Most recent commit should add a test for that

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -167,16 +204,36 @@ func TestInstall(t *testing.T) {
...
+				f.Close()
+			}
+
+			if tc.breakTarFile {
+				err := os.MkdirAll(filepath.Join(os.TempDir(), settings.name, settings.name+".tar.gz"), 0700)
+				require.NoError(t, err, "Setup: breaking the destination tarball shouldn't fail")
+			}
+
+			if tc.breakTempDir {
+				_, err := os.Create(filepath.Join(os.TempDir(), settings.name))
```
ashuntu, line 232: `os.TempDir()` is probably not good in a test, but I'm unsure if we can get around unless we add support for injecting a custom temporary directory? I don't think this realistically is that big a problem since its 1. a temporary dir in the OS temp directory, and 2. its name should be unique to the test anyway since we use `settings.name`.

CarlosNihelton: As long as `settings.name` doesn't conflict we should be fine.

windows-agent/internal/proservices/landscape/executor.go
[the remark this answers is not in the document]
didrocks: yeah, that was it!

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -254,61 +252,47 @@ func installFromURL(ctx context.Context, homeDir string, distro gowsl.Distro, ro
...
 	}
-	defer resp.Body.Close()
 
 	if resp.StatusCode >= 400 {
 		return fmt.Errorf("http request failed with code %d", resp.StatusCode)
 	}
 
-	if _, err := io.Copy(w, resp.Body); err != nil {
+	// Verify checksum and write file to disk
+	reader := io.TeeReader(resp.Body, f)
```
didrocks, line 268: nitpick: this is clearly one case, where a `r` for the reader variable is clear enough and don’t collide with potential generic/package name.

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -254,61 +252,47 @@ func installFromURL(ctx context.Context, homeDir string, distro gowsl.Distro, ro
...
+		return match, err
 	}
 	gotChecksum := fmt.Sprintf("%x", h.Sum(nil))
 
-	out, err := os.ReadFile(path)
-
 	// Compare checksums
-	if gotChecksum != wantChecksum {
-		return fmt.Errorf("checksums do not match (want: %s got: %s) (from %s)", wantChecksum, gotChecksum, string(out))
+	match = wantChecksum == "" || wantChecksum == gotChecksum
```
didrocks, line 291: Hum, please avoid ternary expression with assignement and check at the same time. This is prone to logical errors.

Also, with what you get below, you are not erroring out (only logging) if the checksums don’t match, but still logging it. This will potentially end up with a double logging of checksum mismatching. Also, I don’t think that printing the current downloaded checksum is meaningfull to the user.

I suggest that:
1. we don’t the checksum function at all if `wantChecksum == ""`. You will still need 
2. This function don’t log at all. Let the logging be done by the caller with returning an error in download() if match is false.
3. You need to `io.Copy` though on the caller to consume (as a no-op) the reader so that it’s written to a file, even if we don’t check the checksum (but at least, we don’t pretend and do a checksum for nothing).

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -312,7 +311,7 @@ func mockRootfsFileServer(t *testing.T, ctx context.Context) string {
 	require.NoError(t, err, "Setup: mockRootfsFileServer could not listen")
 
 	go func() {
-		//nolint:gosec // For the sake of simplicity and readability.
+		//nolint:gosec // ignore G114 for the sake of simplicity and readability.
```
didrocks, line 328: ok, so now I see this is for G114. I still think that we can avoid that annotation by adding the timeouts on the server side. Even if this is for a mock, what happens if a test breaks and await forever on the client? Then, we are stuck in our server which may never exists as hanging.

I think something along those lines should be enough:

```
s := &http.Server{
    ReadTimeout: 5 * time.Second,
    WriteTimeout: 5 * time.Second,
}
srv.ListenAndServe()
```

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -149,7 +150,13 @@ func TestInstall(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
 
-	fileServerAddr := mockRootfsFileServer(t, ctx)
+	fileServerAddr, closer := mockRootfsFileServer(t, ctx)
+	t.Cleanup(func() {
+		err := closer.Close()
+		if err != nil {
```
didrocks, line 156: nitpick: you can typically have a one liner here:

`if err := closer.Close(); err != nil` (somewhat a standard pattern when you return only one value, which is an error).

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -149,7 +150,13 @@ func TestInstall(t *testing.T) {
...
+			t.Logf("Could not close mock fileserver: %v", err)
```
didrocks, line 157: Please prefix with "Teardown:" so that we can spot this is not the test call itself.

windows-agent/internal/proservices/landscape/executor_test.go
didrocks: Ah, see my previous comment :) I don’t see why we don’t use `t.Tempdir` TBH, as it will ensure we don’t touch the host system when running the tests and potentially leaving some leftover in case of breakage in the middle of a test?

didrocks (changes requested):
Thanks for the changes! I still think some modifications in the new modifications will make the code more maintainable in the future, but we are getting there :) Do not hesitate if anything is unclear.

As a general note: please do not resolve the conversation yourself. Reminder of the workflow we discussed during last sprint, where it’s the requester doing the conversation resolution himself once he saw that the request was understood and fixed, just +1/comment on each one, please :)

windows-agent/internal/proservices/landscape/executor.go
ashuntu: Its still useful to have the checksums logged during testing, so is a debug log here instead alright?

windows-agent/internal/proservices/landscape/executor_test.go
ashuntu: `installFromURL` would then need to have a way to use the test temp dir, correct?

windows-agent/internal/proservices/landscape/executor.go
didrocks: I don’t see an explicit need TBH that the test doesn’t cover (getting the checksum in case of failure won’t really help debugging it if I’m correct), I think let’s remove it (as you did) and then, we can reintroduce as needed.

windows-agent/internal/proservices/landscape/executor_test.go
ashuntu: Yeah we pass `homeDir` to it and inject it in the test case with a temporary directory. `installFromURL` uses `filepath.Join(os.TempDir(), distro.Name())` as a temporary download path so we'd probably also need to add the ability to inject a custom temp dir if we are to use `t.TempDir()`.

didrocks: Right, and it will make more sense from a system perspective that we can control over where the temp download is processed.

didrocks (changes requested):
Ok, so apart from the TempDir part, all the other points are resolved! Thanks!

ashuntu:
I think everything is good to go aside from the test that is failing. The test itself I think is fine, but its failing to cleanup. I'm not sure why this is happening since it works locally on both Windows and Ubuntu

didrocks:
Ok I think the remaining issues are then the tests cleanup failing. Let us know (Carlos or I) if you need any help to debug this! Then, we’ll be good to merge.

windows-agent/internal/proservices/landscape/service.go
```diff
@@ -364,6 +385,14 @@ func (s *Service) hostname() string {
...
 }
 
+func (s *Service) homeDir() string {
+	return s.homedir
+}
+
+func (s *Service) downloadDir() string {
+	return s.downloaddir
+}
+
```
CarlosNihelton, line 395: If the methods are themselves private, why not just accessing the fields themselves?

ashuntu: Most of the fields already had a pattern doing that, so I just followed what already existed (e.g. `hostname()`, `database()`, etc.).

CarlosNihelton:
The only thing lasting is resolving the merge conflict and we can land this one, right? @ashuntu @didrocks ?

ashuntu:
> The only thing lasting is resolving the merge conflict and we can land this one, right? @ashuntu @didrocks ?

Yep, I can resolve the conflict after this is approved.

didrocks (approved):
Yeah, it’s approved now! Nice work.

CarlosNihelton (approved):

PR #719, opened 2024-04-05
docs: fix firewall diagram

jibel:
Fixed the direction of the arrows and their label. Update the protocols and ports in the table.
Do not treat contracts.canonical.com as a link and remove it from the list of links to ignore in the custom configuration.


didrocks (approved):
The diagram looks now correct to me. Also the arrangement of elements make it way more readable!

PR #720, opened 2024-04-08
maint(gui): Upgrade flutter deps

CarlosNihelton:
Even though there is no report from dependabot, it is not good to stay too many versions behind Flutter stable releases, as we risk having too many breaking changes to address later.

This upgrade doesn't introduce any breaking change that affected us, yet, there are tons of changes in the lockfile.

What really matters here is that CI is aware of the upgrade (by chaging `tools/flutter-version`) and the pubspec.yaml file upgraded the direct dependences as well.

The rest is implementation details.


ashuntu (approved):
Looks good and works locally (after `flutter upgrade` :))

PR #721, opened 2024-04-10
feat(windows-agent): Add support for configuration files

ashuntu:
This adds support for a configuration file to be read and used. Currently, the only options available are verbosity level.

Config files can be located in the current working directory, alongside the binary, in the user's home directory, or in `%UserProfile%/.ubuntupro`. The file must be named `ubuntu-pro-agent.<type>` where `<type>` is one of the file types [supported by Viper](https://github.com/spf13/viper?tab=readme-ov-file#reading-config-files), e.g. `ubuntu-pro-agent.yaml`.

A custom configuration path can be specified by passing the new `--config` flag, which should point to a file that meets the same requirements of Viper described above.

Largely the same as #732.

---

UDENG-2369

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
@@ -102,7 +102,7 @@ func New(o ...option) *App {
 	a.viper = viper.New()
 
 	installVerbosityFlag(&a.rootCmd, a.viper)
-	installConfigFlag(&a.rootCmd, a.viper)
+	installConfigFlag(&a.rootCmd)
 
 	// subcommands
 	a.installVersion()
@@ -186,6 +186,13 @@ func (a App) RootCmd() cobra.Command {
 // SetArgs changes the root command args. Shouldn't be in general necessary apart for tests.
 func (a *App) SetArgs(args ...string) {
 	a.rootCmd.SetArgs(args)
+}
+
+// Config returns the DaemonConfig for test purposes.
+//
+//nolint:revive
+func (a App) Config() daemonConfig {
+	return a.config
 }
 
 // PublicDir creates a directory to store public data in.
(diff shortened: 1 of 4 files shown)


windows-agent/cmd/ubuntu-pro-agent/agent/config.go
```diff
@@ -0,0 +1,97 @@
...
+	if err := vip.ReadInConfig(); err != nil {
+		var e viper.ConfigFileNotFoundError
+		if errors.As(err, &e) {
+			log.Infof(context.Background(), "No configuration file: %v", e)
+		} else {
+			return fmt.Errorf("invalid configuration file: %v", err)
+		}
+	} else {
+		log.Infof(context.Background(), "Using configuration file: %v", vip.ConfigFileUsed())
+	}
```
CarlosNihelton, line 55: You could add one small test function in agent_test.go, similar to `TestNoUsageError` for example, that writes a config file setting the verbosity level to the maximum in one of the supported locations before starting the agent and looking for that file path in the captured stdout.

CarlosNihelton (approved):
Good job! But I think there are a few details to be sorted out. Nothing major though.

CarlosNihelton (changes requested):
Sorry, I approved by mistake.

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
```diff
@@ -212,6 +188,13 @@ func (a *App) SetArgs(args ...string) {
 	a.rootCmd.SetArgs(args)
 }
 
+// Config returns the DaemonConfig for test purposes.
+//
+//nolint:revive
+func (a App) Config() daemonConfig {
+	return a.config
+}
+
```
CarlosNihelton, line 197: If that's just for testing you should move that method to a file in the same package named `export_test.go`. This way the method is only visible in tests.

CarlosNihelton (changes requested):
Almost there, just one minor detail.

ashuntu:
Still a test failing in another area of the code, any idea why?

CarlosNihelton:
> Still a test failing in another area of the code, any idea why?

I think that's the flaky behavior Didier mentioned and we need to investigate.

CarlosNihelton (approved):
LGMT :tada:

PR #729, opened 2024-04-12
fix(security): TLS with self-signed certs in local gRPC communications

CarlosNihelton:
That means no insecure channels between agent and gui and agent and wsl-pro-service.

The strategy adopted here is mutual TLS, i.e. the agent authenticates itself with clients (as normal TLS does), _but clients also authenticate themselves to the agent_, which required extending the Dart `ChannelCredentials` to allow providing custom private keys and certificate chains. I got surprised to learn that there is no equivalent of `credentials.NewTLS(tls.Config)` for `grpc.dart`

Most of the logic to create certificates was placed in common so we can reuse in wsl-pro-service tests.

The specific call to create certificates at the right location was placed in the agent, though, as it must be the source of all truth.

This PR pushed waaay out of my comfort zone, so feel free to dissect and scrutinate any piece of code around generating and loading the certificates.


CarlosNihelton:
Pro client is showing internal server errors in the failed end to end test logs, so it's not an issue related to the chances herein introduced. I'll monitor the situation. In the mean time this is good for a first pass.

CarlosNihelton:
> Pro client is showing internal server errors in the failed end to end test logs, so it's not an issue related to the chances herein introduced. I'll monitor the situation. In the mean time this is good for a first pass.

Pro server issue is gone: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/8690804816/job/23876919448?pr=729

windows-agent/internal/proservices/proservices.go
```diff
@@ -111,7 +118,17 @@ func New(ctx context.Context, publicDir, privateDir string, args ...Option) (s M
...
+	if err := os.MkdirAll(destDir, 0700); err != nil {
+		return s, fmt.Errorf("failed to create certificates directory: %s", err)
+	}
+	certs, err := NewTLSCertificates(destDir)
+	if err != nil {
+		return s, fmt.Errorf("failed to create certificates: %s", err)
+	}
+
+	s.creds = credentials.NewTLS(certs.ServerTLSConfig())
+	return s, err
```
didrocks, line 131: `return s, nil`

didrocks (changes requested):
And here we go!

In general, I saw no fondamental issues on the logic itself, well done as it’s a complex topic!

TBH, I had to skim through the flutter part and I didn’t see anything hurting my eyes from a non expert eyes :) Maybe Ahston can give another look there too?

Do not hesitate if anything I wrote doesn’t make sense to you :)

common/certs/certs.go
```diff
@@ -24,6 +24,35 @@ import (
 // - https://github.com/grpc/grpc-go/blob/master/examples/features/encryption/mTLS
 // - and https://gist.github.com/annanay25/43e3846e21b30818d8dcd5f9987e852d.
 
+// CreateRootCA creates a new root certificate authority (CA) certificate and private key pair with the serial number and common name provided.
+// Only the cert is written into destDir in the PEM format. Being a CA, the certificate and private key returned can be used to sign other certificates.
+func CreateRootCA(commonName string, serialNo *big.Int, destDir string) (rootCert *x509.Certificate, rootKey *ecdsa.PrivateKey, err error) {
```
didrocks, line 29: So, you correctly move it now to the top, BUT, I think the tests should also use the same order then! :)

common/certs/certs.go
```diff
@@ -127,25 +127,25 @@ func template(commonName string, serial *big.Int) *x509.Certificate {
...
-	defer certOut.Close()
-	return pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: DER})
+	defer w.Close()
+	return pem.Encode(w, &pem.Block{Type: "CERTIFICATE", Bytes: DER})
 }
 
 // writeKey writes a private key to disk in PEM format to the given filename.
 func writeKey(filename string, priv *ecdsa.PrivateKey) error {
-	keyOut, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
+	k, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
didrocks, line 141: I’s a file you write to, I would just call it "w" too.

common/certs/certs.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to open %q for writing: %v", filename, err)
+	}
+	defer certOut.Close()
+	return pem.Encode(certOut, &pem.Block{Type: "CERTIFICATE", Bytes: DER})
+}
+
+// writeKey writes a private key to disk in PEM format to the given filename.
+func writeKey(filename string, priv *ecdsa.PrivateKey) error {
+	keyOut, err := os.OpenFile(filename, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600)
```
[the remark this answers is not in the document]
didrocks: Saw  your changes, but as a convention, as it’s a smal file you are just reading, I would call it `w` as you write to it (or `f`)

windows-agent/internal/proservices/certs.go
```diff
@@ -0,0 +1,61 @@
...
+		return agentCerts{}, fmt.Errorf("failed to generate serial number for the CA cert: %v", err)
+	}
+
+	rootCert, rootKey, err := certs.CreateRootCA(common.GRPCServerNameOverride, serial, destDir)
+	if err != nil {
+		return agentCerts{}, err
+	}
+
+	// Create and write the server and client certificates signed by the root certificate created above.
+	serverCert, err := certs.CreateTLSCertificateSignedBy("server", common.GRPCServerNameOverride, serial.Rsh(serial, 2), rootCert, rootKey, destDir)
```
didrocks, line 31: So, here too, I would clarify in the filename the whole terminology:
- rootCA
- agent certificate
- client certificate

(ensure to align all the tests around those name).

That way, looking at the FHS, it’s clear what we are loading where.

windows-agent/internal/proservices/certs.go
```diff
@@ -0,0 +1,61 @@
...
+		ClientCAs:    ca,
+		ClientAuth:   tls.RequireAndVerifyClientCert,
+		MinVersion:   tls.VersionTLS13,
+	}
+}
+
+// agentCerts conveniently holds the root CA and server certificates to make it easy to create a TLS config.
+type agentCerts struct {
+	rootCA     *x509.Certificate
+	serverCert tls.Certificate
```
didrocks, line 60: This is still named `serverCert`

wsl-pro-service/internal/daemon/daemon.go
```diff
@@ -301,6 +309,33 @@ func (d *Daemon) connect(ctx context.Context) (server *streams.Server, err error
 	return streams.NewServer(ctx, d.system, conn), nil
 }
 
+// newTLSConfigFromDir loads certificates from the provided certs path and returns a matching tls.Config.
+func newTLSConfigFromDir(certsPath string) (conf *tls.Config, err error) {
+	decorate.OnError(&err, "could not load TLS config")
+
+	cert, err := tls.LoadX509KeyPair(filepath.Join(certsPath, "client_cert.pem"), filepath.Join(certsPath, "client_key.pem"))
```
didrocks, line 316: I notice that the filename is hardcoded more than in one place, this should probably go in common so that those reused name across multiple projects align on what they fetch? (also, as said in a previous comment, this should be more explicit, like wslinstance_cert.pem or any better name)

windows-agent/internal/proservices/certs_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+			_, err := ps.NewTLSCertificates(dir)
+			if tc.wantErr {
+				require.Error(t, err, "NewTLSCertificates should have failed")
+				return
+			}
+			require.NoError(t, err, "NewTLSCertificates failed")
+		})
+	}
+}
```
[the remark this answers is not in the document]
didrocks: Just to confirm (it’s ok as it’s not exported, covered by other package tests when calling `New()` and the function is short): you are not testing this one as unit test because you think it’s trivial?

didrocks (changes requested):
Nearly there! Thanks for the changes :)

There are still some comments opened for clarification.

Note that I try to not comment on all the side effects of a change anymore of a discussion (I did this time though on what changes I would have considered associated with my previous demands). This is changes like renaming the struct field impacts the associated file name on disk, changing the order of functions in a package impacts the order of the associated tests, and so on. Tell me if this is fine for you!

wsl-pro-service/internal/daemon/daemon.go
CarlosNihelton: > also, as said in a previous comment, this should be more explicit, like wslinstance_cert.pem or any better name

Let me interject on that part: the same client certificates are used by all WSL instances and the GUI, that's why the generic name. I don't think it's worth creating an exclusive pair of certificates for the GUI, it's doable, but unnecessary in my opinion.

Moving to common as constants and adding comments will clarify that more.

windows-agent/internal/proservices/certs_test.go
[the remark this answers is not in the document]
CarlosNihelton: Not only because it's trivial, but because we'd be testing more the `x509.CertPool` then our logic, which is a different condition from when testing itt through `proservices.New()`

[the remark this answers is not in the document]
CarlosNihelton: Notice that we can panic but we can't error this function out.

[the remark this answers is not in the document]
didrocks: agreed!

CarlosNihelton:
> You didn’t rerequest for review, but I think this is good to go now! Thanks for all the changes, nice work on this non trivial task!

I was waiting on CI to pass. Thanks for the hand on this!

PR #731, opened 2024-04-16
docs: update documentation framework

jibel:
Update to the latest version of the documentation framework.


docs/.readthedocs.yaml
```diff
@@ -12,12 +12,12 @@ build:
     python: "3.11"
   jobs:
     post_checkout:
-      - python3 docs/build_requirements.py
+      - python3 .sphinx/build_requirements.py
```
CarlosNihelton, line 15: I think this should be:
```suggestion
      - python3 docs/.sphinx/build_requirements.py
```

As that's where I'm seeing that file and the build is currently broken on that file;
image

PR #732, opened 2024-04-16
feat(wsl-pro-service): Add support for using configuration file

ashuntu:
This adds support for a configuration file to be read and used. Currently, the only options available are verbosity level.

Config files can be located in the current working directory, alongside the binary, in the user's home directory, or in `/etc/`. The file must be named `wsl-pro-service.<type>` where `<type>` is one of the file types [supported by Viper](https://github.com/spf13/viper?tab=readme-ov-file#reading-config-files), e.g. `wsl-pro-service.yaml`.

A custom configuration path can be specified by passing the new `--config` flag, which should point to a file that meets the same requirements of Viper described above.

This also removes the default `-vv` flags passed to the service.

Largely the same as #721.

---

UDENG-2371

### DIFF
--- a/wsl-pro-service/cmd/wsl-pro-service/service/config.go
+++ b/wsl-pro-service/cmd/wsl-pro-service/service/config.go
@@ -70,7 +70,7 @@ func installVerbosityFlag(cmd *cobra.Command, viper *viper.Viper) *int {
 }
 
 // installConfigFlag adds the --config flag to allow for custom config paths.
-func installConfigFlag(cmd *cobra.Command, viper *viper.Viper) *string {
+func installConfigFlag(cmd *cobra.Command) *string {
 	return cmd.PersistentFlags().StringP("config", "c", "", i18n.G("configuration file path"))
 }
 
--- a/wsl-pro-service/cmd/wsl-pro-service/service/config.go
+++ b/wsl-pro-service/cmd/wsl-pro-service/service/config.go
@@ -1,0 +1,96 @@
+package service
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	log "github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logstreamer"
+	"github.com/canonical/ubuntu-pro-for-wsl/common/i18n"
+	"github.com/canonical/ubuntu-pro-for-wsl/wsl-pro-service/internal/consts"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+	"github.com/ubuntu/decorate"
+)
+
+func initViperConfig(name string, cmd *cobra.Command, vip *viper.Viper) (err error) {
+	defer decorate.OnError(&err, "can't load configuration")
+
+	// Use command-line flag for verbosity until configuration is parsed
+	v, err := cmd.Flags().GetCount("verbosity")
+	if err != nil {
+		return fmt.Errorf("internal error: no persistent verbosity flags installed on cmd: %w", err)
+	}
+	setVerboseMode(v)
+
+	// Find a valid configuration file
+	if v, err := cmd.Flags().GetString("config"); err == nil && v != "" {
+		vip.SetConfigFile(v)
+	} else {
+		vip.SetConfigName(name)
+		vip.AddConfigPath("./")
+		vip.AddConfigPath("$HOME/")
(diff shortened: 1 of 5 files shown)


wsl-pro-service/cmd/wsl-pro-service/service/config.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return r
+}
+
+// installConfigFlag adds the --config flag to allow for custom config paths.
+func installConfigFlag(cmd *cobra.Command, viper *viper.Viper) *string {
+	r := cmd.PersistentFlags().StringP("config", "c", "", i18n.G("configuration file path"))
+	if err := viper.BindPFlag("config", cmd.PersistentFlags().Lookup("config")); err != nil {
+		log.Warning(context.Background(), err)
+	}
```
CarlosNihelton, line 77: You're adding a configuration option to choose a configuration file in the configuration file parser. I think this recursion is unnecessary.

wsl-pro-service/cmd/wsl-pro-service/service/config.go
```diff
@@ -0,0 +1,96 @@
...
+	// Load the config
+	if err := vip.ReadInConfig(); err != nil {
+		var e viper.ConfigFileNotFoundError
+		if errors.As(err, &e) {
+			log.Infof(context.Background(), "No configuration file: %v", e)
+		} else {
+			return fmt.Errorf("invalid configuration file: %v", err)
+		}
+	} else {
+		log.Infof(context.Background(), "Using configuration file: %v", vip.ConfigFileUsed())
```
CarlosNihelton, line 53: This one shouldn't be hard to exercise: add a test similar to `TestNoUsageError` that writes a configuration file in one of the supported locations setting the verbosity level to the maximum and assert the configuration file path is present in the captured stdout. I don't think we need a whole table testing all possible supported locations, as that would be testing Viper, instead of testing our connecting logic.

CarlosNihelton (changes requested):
Pretty much the same as #721 . Just minor details. Good job so far.

wsl-pro-service/cmd/wsl-pro-service/service/config.go
ashuntu: The config won't be used at this point, so this log would rely on the presence of `-v` instead until it is. If you think it's still worthwhile, I can still test for the log output while passing in `-v`

CarlosNihelton: You can add `-v` and have the config file set verbosity to 3, to prove it works and assert that the captured stdout refers to it.
Additionally, you can set an invalid config and assert for errors. For inspiration:

```go
func TestConfigVerbosity(t *testing.T) {
	getStdout := captureStdout(t)

	sys, _ := testutils.MockSystem(t)
	a := service.New(service.WithSystem(sys))
	a.SetArgs("version", "-v")

	filename := "wsl-pro-service.yaml"
	configNextToBinaryPath := filepath.Join(filepath.Dir(os.Args[0]), filename)
	require.NoError(t, os.WriteFile(configNextToBinaryPath, []byte("invalid-config"), 0644), "Couldn't write config file")
	// require.NoError(t, os.WriteFile(configNextToBinaryPath, []byte("verbosity: 3"), 0644), "Couldn't write config file")


	err := a.Run()
	out := getStdout()
	require.Error(t, err, "Run should not return an error, stdout: %v", out)
	require.Contains(t, out, filename, "Config file should be read. Output was\n", out)
}
```

and 

https://github.com/ubuntu/authd/blob/main/cmd/authd/daemon/daemon_test.go#L278-L334

ashuntu:
There is still a test failing, though I'm unsure if it's related or not.

wsl-pro-service/cmd/wsl-pro-service/service/config.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+// SetVerboseMode change ErrorFormat and logs between very, middly and non verbose.
+func setVerboseMode(level int) {
+	var reportCaller bool
+	switch level {
+	case 0:
+		logrus.SetLevel(consts.DefaultLogLevel)
+	case 1:
+		logrus.SetLevel(logrus.InfoLevel)
```
CarlosNihelton, line 84: You should be able to cover this line if you set verbosity to 1 in one of the test cases you've added (or another one specific for this if you prefer).

CarlosNihelton (approved):
LGTM :tada:

PR #739, opened 2024-04-18
maint(wsl-pro-service): Prepare changelog for wsl-pro-service v0.1.3

CarlosNihelton:
If PR #732 get's done in time, we must add a matching line in the changelog. Otherwise, this is it.


didrocks:
(I think we shouldn’t wait on 732)

CarlosNihelton:
Changes in coverage are unrelated to this.

PR #740, opened 2024-04-24
fix(agent): WSL net address selection

CarlosNihelton:
@wck0 demonstrated that the agent might choose the wrong adapter address to listen to when the setup has more than one "Hyper-V Virtual Ethernet Adapter" (description goes numbered from the second on, by default).

This PR aims to desambiguate the case when more than one adapter is found containing that string in the description (that cannot be easily changed by the user). It does so by checking the presence of "vEthernet (WSL" in the friendly name, which the user can change, but I consider unlikely due the fact that the adapter is hidden by default.

Additionally, if the network stack is configured to the brand new "mirrored" mode, that adapter address won't be available inside the WSL instances, thus we must listen on 127.0.0.1. The wsl-pro-service is alreay aware of that possibility, so no changes are required in there. A good side effect of relying on `wsl --system` for that detection is that we ensure the existence of the WSL network adapter by the time we look for it by starting a WSL distro (we use the system one because that's the only one we can assume always present).

Testing the existing functionality (plus the added one) required mocking the `wsl --system wslinfo --networking-mode` command as well as the Win32 `GetAdapterAddresses` function. The first is mocked by invoking the current process (the test runner) with special arguments to trigger it to run a specific test function designed to emulate that command (`daemon/testutils`). As shown in the conversation in this PR, we've been mocking subprocesses in a suboptimal way, fixing all other places is subject to another PR. The Win32 `GetAdapterAddresses` requires a more complex mock, because it deals with low level data structures and memory allocation. We could have wrapped it with a little bit more logic and mock the wrapper instead, with less complexity, but that would hide a significant portion of logic dealing with memory buffers, which should be exercised to ensure correctness. Thus this PR takes the hard path of dealing with the low level structures, filling pointers/sizes pairs with low level data (see `daemon/networking_mock.go` and `daemon/networking_windows.go`). In the end we get more fine grained control over the test behavior and the complexity won't surface at the call site.

---
Fix: UDENG-2712


windows-agent/internal/daemon/networking_windows.go
```diff
@@ -35,6 +65,18 @@ func getWslIP() (net.IP, error) {
...
 }
 
+// networkIsMirrored detects whether the WSL network is mirrored or not.
+func networkIsMirrored() (bool, error) {
+	// It does so by launching the system distribution.
+	cmd := exec.CommandContext(context.Background(), "wsl", "--system", "wslinfo", "--networking-mode", "-n")
+
+	out, err := cmd.Output()
+	if err != nil {
+		return false, fmt.Errorf("could not get networking mode: %w", err)
```
didrocks, line 75: The error here will not tell you much (only which error code the program exited with). You probably want to print stdout + stderr.

didrocks (changes requested):
I think code-wise, this looks good! I only have one minor command on error output when the command failed.

However, testing-wise. There is probably some work to do. I feel we cover by accident the error and non error case as one could be done on Windows, and the other on Linux, we ideally want:
- be in control of our command output with a mock, as this is an external dependency.
- have both positive and negative tests exercised on each platform (or at least in one of them), in explicit tests.
- ensure we did run the coverage manually to confirm this too.

Tell me if you need any help!

windows-agent/internal/daemon/networking_mock.go
```diff
@@ -1,18 +1,210 @@
...
+// WslSystemMock mocks commands running inside the WSL system distro.
+// Add it to your package_test with:
+//
+//	func TestWithWslSystemMock(t *testing.T) { daemon.WslSystemMock(t) }
+//
+//nolint:thelper // This is a faux test used to mock commands running via `wsl -- system`
+func WslSystemMock(t *testing.T) {
+	// Setup
+	if t.Name() != "TestWithWslSystemMock" {
+		panic("The WslSystemMock faux test must be named TestWithWslSystemMock")
```
CarlosNihelton, line 150: codecov is complaining that my mock of WSL is not covered, which was created solely for testing and is broadly used. How should we handle that? Should the command invoking the test mock be `go test -coverpkg....` matching the same directory as the main tests are running? 🤷🏽‍♂️ 

I also considered instead of running `go test` as the executable mock, run the same binary the test process is running. Since the function implementing the executable is part of the same test file, `go test` already compiled it when running the test cases, so `os.Executable()` is a valid target for the mock, with a minor adjustment to the command line invocation: instead of `go test -run ^TESTNAME$` it would be `thisbinary -test.run ^TESTNAME$`. I suspect that would speed up tests with mocks, but I'm not quite sure how much, as `go test` probably runs part of the compiler toolchain at least to figure out that there is no need to compile anything.

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -73,6 +73,7 @@ func TestStartQuit(t *testing.T) {
 			serveErr := make(chan error)
 			go func() {
 				serveErr <- d.Serve(ctx)
+				close(serveErr)
```
didrocks, line 76: nicely spotted!

windows-agent/internal/daemon/networking_mock.go
```diff
@@ -1,18 +1,210 @@
...
+)
+
+type mockIPAdaptersState int
+
+const (
+	mockError mockIPAdaptersState = iota
+	emptyList
+	noHyperVAdapterInList
+	singleHyperVAdapterInList
+	ok
```
didrocks, line 23: I feel `ok` here to not be descriptive enough. Here, you want to control the mock behaviour, right? I would prefix them all with something related to this state.

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -144,35 +145,73 @@ func TestStartQuit(t *testing.T) {
...
+
+			wsl := daemon.NewWslSystemMock(tc.netmode, nil, tc.denyWslSystemAccess)
+			ipconfig := daemon.NewHostIPConfigMock(tc.withAdapters)
+			serveErr := make(chan error)
+			go func() {
+				serveErr <- d.Serve(ctx, daemon.WithWslSystemDistro(wsl), daemon.WithHostIPConfig(&ipconfig))
+			}()
+
+			if tc.wantErr {
+				require.NotErrorIs(t, <-serveErr, grpc.ErrServerStopped, "Serve should fail when the WSL IP cannot be found")
```
didrocks, line 192: shouldn’t we quit the daemon if we fail here? If we can double quit (to be checked) without any issue, I would still `defer d.Quit(ctx, false)` after creating the daemon.

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -144,35 +145,73 @@ func TestStartQuit(t *testing.T) {
...
+			ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+			defer cancel()
+
+			d := daemon.New(ctx, registerer, addrDir)
+
+			wsl := daemon.NewWslSystemMock(tc.netmode, nil, tc.denyWslSystemAccess)
+			ipconfig := daemon.NewHostIPConfigMock(tc.withAdapters)
+			serveErr := make(chan error)
+			go func() {
+				serveErr <- d.Serve(ctx, daemon.WithWslSystemDistro(wsl), daemon.WithHostIPConfig(&ipconfig))
```
didrocks, line 188: you sohuld close the channel (only noticed because you fixed it above :))

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -144,35 +145,73 @@ func TestStartQuit(t *testing.T) {
...
+
+			serverStopped := make(chan struct{})
+			go func() {
+				time.Sleep(500 * time.Millisecond)
+				d.Quit(ctx, false)
+				close(serverStopped)
+			}()
+			<-serverStopped
+
+			require.ErrorIs(t, <-serveErr, grpc.ErrServerStopped, "Serve should not fail when the WSL IP is found")
```
didrocks, line 204: I see that you rely on the exact error for stopping now that you wrap it. Does it bring anything in the rest of the code? It seems that apart from the test, we don’t rely on this exact error (and so, don’t really care as part of our API).

Shouldn’t we just want some error being returned in both case, when the WSL IP cannot be found? And maybe in the `Serve` method, we capture the underlying error and tell "if we get ErrServerStopped", this is ok? (honest question, needs digging)

windows-agent/internal/daemon/networking.go
```diff
@@ -0,0 +1,101 @@
...
+	}
+}
+
+func findWslAdapterIP(i hostIpconfig) (net.IP, error) {
+	const targetDesc = "Hyper-V Virtual Ethernet Adapter"
+	const vEthernetName = "vEthernet (WSL"
+
+	head, err := i.getAdaptersAddresses()
+	if err != nil {
+		return nil, err
```
didrocks, line 38: Looking at `getAdaptersAddresses`, it seems unatural to try to trigger the error due to having to grow the internal buffer too much, right?

windows-agent/internal/daemon/networking.go
```diff
@@ -0,0 +1,101 @@
...
+	case "mirrored":
+		return net.IPv4(127, 0, 0, 1), nil
+	case "nat":
+		return findWslAdapterIP(i)
+	default:
+		return nil, fmt.Errorf("unknown networking mode: %s", mode)
+	}
+}
+
+func findWslAdapterIP(i hostIpconfig) (net.IP, error) {
```
didrocks, line 32: you added comments on the other private function, maybe this one deserve it too?

windows-agent/internal/daemon/networking.go
```diff
@@ -0,0 +1,101 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Filter the adapters by description.
+	var candidates []ipAdapterAddresses
+	for node := head; node != nil; node = node.Next() {
+		desc := node.Description()
+		// The adapter description could be "Hyper-V Virtual Ethernet Adapter #No"
+		if !strings.Contains(desc, targetDesc) {
```
didrocks, line 46: `targetDesc` and `vEthernetName` are only used once, so having it declared as a const force you to look up while the direct string comparison (even comparing with the method itself) is more straighforward.

What about:
`if !strings.Contains(node.Description(), "Hyper-V Virtual Ethernet Adapter") {` ?

Also, note that I directly used `node.Description()` than assigning a variable that is not going to reused. Btw, you did exactly this with `node.FriendlyName()` few lines below :p

windows-agent/internal/daemon/networking.go
```diff
@@ -0,0 +1,101 @@
...
+			continue
+		}
+
+		candidates = append(candidates, node)
+	}
+
+	if len(candidates) == 1 {
```
didrocks, line 53: what if there is one candidate only, not matching `"vEthernet (WSL"`?  Is it a situation that could happen in theory and we should then only return "could not find WSL adapter"? (in that case, you can just remove this additional "one line" check as the range will take care of it for you)

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
```diff
@@ -461,3 +462,5 @@ func captureStdout(t *testing.T) func() string {
 		return out.String()
 	}
 }
+
+func TestWithWslSystemMock(t *testing.T) { daemon.WslSystemMock(t) }
```
didrocks, line 466: With the changes to the mock, you should not need this anymore, correct?

windows-agent/internal/daemon/networking_mock.go
didrocks: This is not really how we do it. I think the issues comes from the way the you are running the subprocess.

I suggest that instead of passing this, you are passing a subcommand path (with args) to the real object, a little bit like https://github.com/ubuntu/authd/blob/main/internal/users/localgroups/localgroups.go#L19 (but attached on the struct).
In the tests, then, you pass another command to the constructor:  https://github.com/ubuntu/authd/blob/main/internal/users/localgroups/localgroups_test.go#L213C4-L213C14.

This command reference another test from the same function:
https://github.com/ubuntu/authd/blob/main/internal/users/localgroups/localgroups_test.go#L257. In most of the project, if you don’t need it for special build or integration tests, the whole implementation is there, like in https://github.com/ubuntu/adsys/blob/main/internal/policies/apparmor/apparmor_test.go#L212.

You can see in that case that this is referenced to a subpackage, which is the "mock", that will be excluded from codecov (manually or not): https://github.com/ubuntu/authd/blob/main/internal/users/localgroups/tests/tests.go (with the according assertion).

Note that contrary to this case, I would advise to ensure this is running on tests, like  with https://github.com/ubuntu/authd/blob/main/internal/testsdetection/ if you really need to put some part in production code to assert quickly that this is not allowed: https://github.com/ubuntu/authd/blob/main/internal/services/permissions/testutils.go.

Do not then hesitate to import, like from the `_linux.go` file to attach it to the ipConfig or wslSystem for instance (this, combined with the previous advice on how to attach it to those struct should give a good start!)

Finally, there is the need for integration tests (but it seems, we are not there yet!),. We can reuse then the import in the finale binary, like [https://github.com/ubuntu/authd/blob/main/internal/testsdetection/ and ](https://github.com/ubuntu/authd/blob/main/cmd/authd/integrationtests.go.)

windows-agent/internal/daemon/networking.go
```diff
@@ -0,0 +1,101 @@
...
+	err := cmd.Run()
+	out := bytes.TrimSpace(stdout.Bytes())
+	if err != nil {
+		return "", fmt.Errorf("%s: error: %v.\n    Stdout: %s\n    Stderr: %s", cmd.Path, err, out, stderr.String())
+	}
+
+	return strings.Split(string(out), "\n")[0], nil
+}
+
+type wslSystemDistro interface {
```
didrocks, line 88: I think we probably can drastivally simplify by removing the needs to define interfaces there in getting ` ipConfig` and `wslSystem` in `networking.go`. Those can embbed a mock struct as children, and the implemention in networking.go of `getAdaptersAddresses` for instance use the embedded object to drive the real or mock implementation.

See also my other comments which works in conjunction with this one. WDYT?

didrocks (changes requested):
Note that I didn’t go to the end of the review itself, as I think there are some intrusive change needed based on the recent patterns we introduced in authd. I’m happy to walk over some examples with you so that we can see together how this could work here.

We will get there! :)

windows-agent/internal/daemon/networking.go
CarlosNihelton: I didn't quite get this part. Focusing on the `ipConfig` struct, which is dummy, it only serves to allow inject a mock behavior for the Win32 API. You're suggesting to embed the mock inside the production one, but then how are we going to dispatch between the production and the mock versions of `GetAdaptersAddresses`? And embeding on that direction won't let me pass the mock as argument of type `ipConfig`. If I do the other way around, embedding the production one into the mock, then I'll be able to pass the mock where an `ipConfig` object is expected, but there is no means of dispatchig the call to the replacement method of `GetAdaptersAddresses`.

CarlosNihelton: I think it will work better if I pass the function as part of the `oprions` struct. The production implementation is a function forwarding to the Win32 API, while the mocked version can be a method bound to a specific instance of the mock type. This way I can avoid the `ipConfig` interface and related types (except for the mock one) and still retain the ability to customize the mock object state (and thus the behavior of the `getAdaptersAddresses` function.

didrocks: sounds good to me. Either that, or we have one big "system" instance where we attach all the side effects and mocks :)

windows-agent/internal/daemon/daemon_test.go
CarlosNihelton: Added a `defer d.Quit` right after creating the deamon.

windows-agent/internal/daemon/networking.go
CarlosNihelton: Yeah, I think it would be an unnecessary burden for the CI runners.

windows-agent/internal/daemon/daemon.go
```diff
@@ -38,24 +38,53 @@ func New(ctx context.Context, registerGRPCServices GRPCServiceRegisterer, addrDi
...
+}
+
+// WithWslSystemDistro is an optional argument for Serve that injects a backend to run commands inside the wsl --system distro.
+// For testing purposes only.
+func WithWslSystemDistro(wsl wslSystemDistro) Option {
+	return func(o *options) {
+		o.wsl = wsl
+	}
+}
+
```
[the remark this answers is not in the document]
CarlosNihelton: The implementation deviated considerably from this.

windows-agent/internal/daemon/daemon_test.go
```diff
@@ -85,7 +86,7 @@ func TestStartQuit(t *testing.T) {
 					addrContents, err = os.ReadFile(addrPath)
 					require.NoError(t, err, "Address file should be readable")
 					return string(addrContents) != "# Old port file"
-				}, 500*time.Millisecond, 50*time.Millisecond, "Pre-existing address file should be overwritten after dameon.New()")
+				}, 5000*time.Millisecond, 100*time.Millisecond, "Pre-existing address file should be overwritten after daemon.New()")
```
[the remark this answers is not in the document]
CarlosNihelton: I think we no longer need to increase those timeouts since we no longer rely on running `go test` as part of the test.

windows-agent/internal/daemon/networking.go
CarlosNihelton: I had to test back older versions of WSL to be sure. We can safely check for both the adapter description and friendly name on the same loop.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
CarlosNihelton: Yeah, instead of doing this, I import `daemontests`, otherwise having that test function is still needed because we need to  test the agent with the command line mock.

windows-agent/internal/daemon/daemon_test.go
CarlosNihelton: API wise we don't need to return that specific error object. And, with the latest changes, we either fail early (thus with a different error) or shouldn't fail, for which case `require.NoError` must do it.

windows-agent/internal/daemon/daemon.go
```diff
@@ -38,24 +38,53 @@ func New(ctx context.Context, registerGRPCServices GRPCServiceRegisterer, addrDi
...
-	wslIP, err := getWslIP()
+	opts := options{h: newIPConfig(), wsl: newWslSystem()}
+	for _, f := range args {
+		f(&opts)
+	}
+
+	wslIP, err := getWslIP(ctx, opts.h, opts.wsl)
 	if err != nil {
-		return fmt.Errorf("could not get the WSL adapter IP: %v", err)
+		return fmt.Errorf("could not get the WSL adapter IP: %w", err)
```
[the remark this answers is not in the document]
CarlosNihelton: API wise I don't consider it necessary, there is no place in the agent's upper level where the exact error returned would make a difference so far. I did that solely for testing, but as of now, that's no longer necessary, as we either fail early or not at all by the time we look into the errors returned by `daemon.Serve` in tests.

windows-agent/internal/daemon/networking_windows.go
```diff
@@ -1,127 +1,78 @@
...
 
-// byteCount returns the number of bytes in the buffer.
-func (b buffer[T]) byteCount() uint32 {
-	var t T
-	sizeOf := uint32(reflect.TypeOf(t).Size())
-	n := uint32(len(b.data))
-	return n * sizeOf
-}
+// fillFromTemplate fills the ipAdapterAddresses struct with the values from the mockIPAddrsTemplate struct for testing purposes.
+func (a *ipAdapterAddresses) fillFromTemplate(template *mockIPAddrsTemplate, next *ipAdapterAddresses) {
```
didrocks, line 56: It seems we only fill the public methods from that struct. Do you think this could be something that is in the `daemontests` subpackage too that we inject only when needed? (and so, it doesn’t end up in production code too?) That would cascade to `ipToRawSockaddrAny` too.

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,79 @@
+// Package testutils exports test helpers to be used in other packages that need to change internal behaviors of the daemon.
+package testutils
```
didrocks, line 2: if you create the subpackage named `daemontests` that can move there too :)

Actually, maybe testutils is a better naming for it and we should follow that convention (that forces though to use an alias when importing it). Which one do ou prefer?

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,79 @@
...
+	"syscall"
+	"testing"
+
+	"github.com/canonical/ubuntu-pro-for-wsl/common/testdetection"
+)
+
+// MockWslSystemCmd mocks commands running inside the WSL system distro.
+// Add it to your package_test with:
+//
+//	func TestWithWslSystemMock(t *testing.T) { daemontests.MockWslSystemCmd(t) }
```
didrocks, line 18: That description will change when we explicit enable it instead of running `init()`.

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	var argv []string
+	begin := slices.Index(os.Args, "--")
+	if begin != -1 {
+		argv = os.Args[begin+1:]
+	}
+
+	// Action
+	exit := func(args []string) int {
```
didrocks, line 47: I see what you did here, but as returning only do an exit, I would really just exit in the 2 cases we don’t return 0, and then don’t use a closure that you call immediatly.

That will save the finale `if` too by always printing in the "happy path"

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,79 @@
...
+	}(argv)
+
+	// Ensure we clean-exit.
+
+	if exit == 0 {
+		// testing library only prints this line when it fails
+		// Manually printing it means that we can simply remove the last two lines to get the true output
+		fmt.Fprintf(os.Stdout, "\nexit status 0\n")
+	}
+	syscall.Exit(exit)
```
didrocks, line 78: why syscall instead of os?

didrocks (changes requested):
This is definitively easier to understand that the last pass, good work!

I think though that we have too many code in the production one. Last week, we discussed the changes I wanted to do in the "user" package in authd to follow more the permissions one, which put most of the logic outside of the production code, change the defaults explicitely to the caller, in a centralized place. I think we should follow this.

However, on the subpackage name, I must admit I like the symmetry of using "testutils", it only enforce the caller side to handle it properly with an alias as we will have probably multiple testutils, but if you want to keep that name, I agree this is a way forward and will do the changes in authd.

Tell me if anything is unclear.

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: copy-pasta from the existing mocks that I didn't put thought on to change😅

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: I think testutils is less invasive, easy to filter out of coverage reports (already done actually) and can have more test-related things that we can export or not to higher levels.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
```diff
@@ -508,3 +509,7 @@ func captureStdout(t *testing.T) func() string {
 		return out.String()
 	}
 }
+
+func TestWithWslSystemMock(t *testing.T) {
```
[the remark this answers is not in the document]
CarlosNihelton: I need the function `testutils.MockWslSystemCmd()` in the binary when testing the agent. I guess I currently can't have it "automatically" because that's the only function in `testutils` and thus, if I don't explicitly call it, then it's an unused package. If instead, I have an API to override the defaults and call it, then I get that function in binary transparently? Let me test that.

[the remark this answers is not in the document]
CarlosNihelton: Couldn't figure out a clean way to allow invoking the current binary as the mock for the subprocess (`wsl --system` in this case).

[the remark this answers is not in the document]
CarlosNihelton: I did remove the inits, and properly set the `defaultOptions` while exposing an API only for tests to be able to override them, but I couldn't figure out a better way to have a way to have a test named `TestWithWslSystemMock` which allows replacing `wsl --system wslinfo ...` with `os.Args[0] -test.run '^TestWithWslSystemMock$' -- wslinfo ... ` in `agent_test`. The pattern I see in authd to mock executables is pretty much this.

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: Not sure, as though I replaced init() properly, I couldn't find a way to have that test case in other package tests, like in the `agent_test`, that allows replacing the `wsl.exe --system` subprocess without doing this.

CarlosNihelton:
Rebasing on `bump-go-1.22.3` to get CI green again. See #756 for details.

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,105 @@
...
+	testdetection.MustBeTesting()
+
+	m := NewHostIPConfigMock(MultipleHyperVAdaptersInList)
+
+	defaultOptions.wslSystemCmd = []string{
+		os.Args[0],
+		"-test.run",
+		"TestWithWslSystemMock",
+		"--",
+		"wslinfo",
```
didrocks, line 93: see my other comment on what to put in `wslSystemCmd`, that way, you don't need to duplicate this (but just think about consuming them in the mock. If you only have one variant, I don’t think you need to assert on them though)

windows-agent/internal/daemon/testutils/testutils.go
```diff
@@ -0,0 +1,105 @@
...
+	default:
+		fmt.Fprintf(os.Stderr, "Invalid argument: [%s]\n", a)
+		fmt.Fprintln(os.Stderr, errorUsage)
+		os.Exit(1)
+	}
+}
+
+var (
+	//go:linkname defaultOptions github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/daemon.defaultOptions
+	defaultOptions struct {
```
didrocks, line 75: Yeah, that’s the only thing I don’t like with this pattern: we have the repeat the struct def here (but at least, it’s a compile time error if they don’t match)

windows-agent/internal/daemon/networking_mock.go
```diff
@@ -1,18 +1,143 @@
...
+		localIP = net.IPv4(0, 0, 0, 0)
+	}
+
+	switch m.state {
+	case noHyperVAdapterInList:
+		m.addrs = adaptersList
+	case singleHyperVAdapterInList:
+		m.addrs = append(
+			adaptersList,
+			mockIPAddrsTemplate{
```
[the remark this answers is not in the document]
didrocks: agreed, I just wanted you to be aware of that and double check if we can reuse the same names by changing the order :)

windows-agent/internal/daemon/testutils/testutils.go
didrocks: Another option to ensure we have the same alias everywhere when importing it: maybe the subdirectory can be named `testutils` but the package name will still be something related to the parent package, like `daemontestutils`?
That way, when import a name that will be autoalias to the new name. Having long name there, for tests, don’t seem to undermine the gain by having this explicitely. Tell me what you think.

windows-agent/internal/daemon/export_test.go
```diff
@@ -1,20 +1,26 @@
...
-// This only works when the build tag is "linux" or "gowslmock".
-func SetWslIPErr(t *testing.T) {
-	t.Helper()
+type MockIPAdaptersState = mockIPAdaptersState
+
+var NewHostIPConfigMock = newHostIPConfigMock
 
-	if !wsl.MockAvailable() {
-		t.Skip("gowslmock not available")
+func WithWslSystemCmd(cmd []string, cmdEnv []string) Option {
```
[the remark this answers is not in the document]
CarlosNihelton: Sometimes rebasing tricks me :)

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: That's better than the alias. We could also name the directory as `daemontestutils`  and, as we `grep -hv -e "testutils"` when filtering the coverage reports it will still match. Long names for packages are not the most idiomatic way in Go, but as you said, only for tests won't harm. I'm all for it.

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: Now I got what you meant.

windows-agent/internal/daemon/testutils/testutils.go
CarlosNihelton: Indeed.

CarlosNihelton:
> ## [Codecov](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) Report
> Attention: Patch coverage is `90.34483%` with `14 lines` in your changes are missing coverage. Please review.
> 
> > Project coverage is 87.74%. Comparing base [(`0547347`)](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/commit/0547347c9b6e9d7fbbed6927dc7592d88fef9977?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical) to head [(`a519a19`)](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?dropdown=coverage&src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical).
> 
> [Files](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)	Patch %	Lines
> [windows-agent/internal/daemon/networking.go](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&filepath=windows-agent%2Finternal%2Fdaemon%2Fnetworking.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical#diff-d2luZG93cy1hZ2VudC9pbnRlcm5hbC9kYWVtb24vbmV0d29ya2luZy5nbw==)	93.33%	[6 Missing and 2 partials ⚠️ ](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)
> [common/testdetection/testdetection.go](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&filepath=common%2Ftestdetection%2Ftestdetection.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical#diff-Y29tbW9uL3Rlc3RkZXRlY3Rpb24vdGVzdGRldGVjdGlvbi5nbw==)	33.33%	[1 Missing and 1 partial ⚠️ ](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)
> [windows-agent/internal/daemon/networking_linux.go](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&filepath=windows-agent%2Finternal%2Fdaemon%2Fnetworking_linux.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical#diff-d2luZG93cy1hZ2VudC9pbnRlcm5hbC9kYWVtb24vbmV0d29ya2luZ19saW51eC5nbw==)	80.00%	[2 Missing ⚠️ ](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)
> [...indows-agent/internal/daemon/networking_windows.go](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&filepath=windows-agent%2Finternal%2Fdaemon%2Fnetworking_windows.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical#diff-d2luZG93cy1hZ2VudC9pbnRlcm5hbC9kYWVtb24vbmV0d29ya2luZ193aW5kb3dzLmdv)	80.00%	[2 Missing ⚠️ ](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical)
> Additional details and impacted files
> [☔ View full report in Codecov by Sentry](https://app.codecov.io/gh/canonical/ubuntu-pro-for-wsl/pull/740?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=canonical).

And no mentions to `daemontestutils` in codecov complains :)

didrocks (approved):
Excellent! I think now we have a solid pattern for it and I’m really happy with the code entering. Well done!

PR #746, opened 2024-04-26
feat(gui-ubuntupro): Add support for light theme

ashuntu:
Previously, a dark themed app would be used for any OS theme. Now, the theme is dynamic and changes to a light variant if the OS theme changes.

Right now, the light theme doesn't have a background image like the dark theme does, but it can be easily added.

On Windows 11, you can change your theme by going to Settings > Personalization > Colors > "Choose your mode".

*Screenshots below*

<details>
<summary>Light theme</summary>

image
image
image

</details>

<details>
<summary>Dark theme (same as before this change)</summary>

image
image
image

</details>

---

UDENG-936


jibel:
Thanks for working on this. The result is quite nice!
I didn't review the PR yet, but as you mention, could you add a background to the light theme, similar to the dark one?

gui/packages/ubuntupro/lib/pages/widgets/page_widgets.dart
```diff
@@ -91,17 +91,20 @@ class ColumnLandingPage extends StatelessWidget {
...
 
     return Pro4WSLPage(
       body: Stack(
         fit: StackFit.expand,
         children: [
-          Positioned.fill(
-            child: Image.asset(
-              'assets/05_suru2_dark_2K.jpg',
-              fit: BoxFit.fill,
+          if (dark)
```
CarlosNihelton, line 101: If you want to get 100% coverage in your PR you can easily add this line to one of the Landscape and Subscribe pages:

```dart
tester.view.platformDispatcher.platformBrightnessTestValue =
        Brightness.dark;
```

CarlosNihelton (commented):
Code wise this is great! 

But let me invoke @anasereijo  before final approval on whether we should add a background to the light theme or remove the background on the dark one.

@anasereijo the context is: the dark theme features a background taken from the Ubuntu Pro website back in the time when we started writing this application. The website changed the suru in the main page quite significantly since then by the way.
The question is: should we remove the suru in the dark mode or create a similar one for the light mode?

See screenshots above: https://github.com/canonical/ubuntu-pro-for-wsl/pull/746#issue-2266488724

ashuntu:
This is a quick and dirty version of it as a light background, just by brightening the dark version and upping the contrast a bit.

image

anasereijo:
Hey! Thanks for sharing this. We usually don't use the branded backgrounds in applications/ui. I also checked what was [designed](https://www.figma.com/file/Ur0hDHKPvjwWeBbuH9xtqC/Ubuntu-Pro-for-Windows?type=design&node-id=552%3A2390&mode=design&t=SViZiR4J7jkZ54Y7-1) previously (by Elio I think) and it doesn't have a branded background. That said, I would simply remove the branded backgrounds and keep the dark/light background colours.

@lyubomir-popov do you have any thoughts from a brand perspective?

lyubomir-popov:
Hi, it is important to ensure the Suru backgrounds are never covered by UI - if you feel confident that's achievable regardless of screen sizes, then I see no harm in adding the light version of this suru. 

There are a couple of other considerations:
- the suru background graphics should be used as is, and not lightened / ajdusted in any way. You can find the light version [here]
(https://drive.google.com/file/d/1CoteBY05XXa3yHe2SMIGZtdPPlFKOWZn/view), and the matching dark version [here](https://drive.google.com/file/d/1Cl2hAYQTmYJ4YxIuPf3_8wxiq82Aq2ta/view?usp=sharing)

- You'll need to change the UI background to #f3f3f3 for the light theme, and #262626 for the dark theme (that is probably already done). If that causes UI issues, like input or button background remaining white, I would advise that we put this on hold until a proper theming can be implemented. In that case, removign the background ffrom the dark theme would make more sense.

ashuntu:
It would be tough to fit the suru backgrounds without having the UI overlap and retaining clarity, as seen in the screenshots below, so I think we'll opt to just remove the background in all themes of the app.

image
image

gui/packages/ubuntupro/lib/pages/widgets/page_widgets.dart
ashuntu: This shouldn't be relevant now, since we're removing the background image entirely.

CarlosNihelton (approved):
LGTM  👍🏽

PR #754, opened 2024-05-06
feat(gui-ubuntupro): Make version text selectable for copying

ashuntu:
Very small change that just allows the version text in the status bar of the GUI to be selected for copying.

---

UDENG-2749


CarlosNihelton (approved):
:tada:

PR #776, opened 2024-05-28
fix(wsl-pro-service): Addressing some flakyness

CarlosNihelton:
Some of the wsl-pro-service tests frequently show up as spurious failures, such as `internal/streams/server_tests.go` `TestServe` and `TestStop`. Some observed both in GH CI as well as in Launchpad builds (autopkgtests).

This PR is an attempt to mitigate some non-deterministic failure behavior in this module. Here I focused on the test code itself, as I think I found opportunity for improvement in there. It's possible that some flaky behavior I didn't see or address now will be adressed in futures PRs where I plan to rework the gRPC client creation and reconnection logic, which will have more invasive changes in the production code anyways.

Here's a gist of the reasoning behind the changes:

- `streams.Server.Serve()` - test cases run it in a different goroutine spawn by the test code. The changes increase the slack of time in attempt to give the runtime more change to schedule starting the server. Similarly, all checks to the mock agent's all connected getter (`agent.Service.AllConnected()`) got the same slack on `require.Eventually` check calls. Its work mentioning that the condition asserted by the require calls is already guarded by mutexes, so no races.

- `streams_tests.mockService` relies on a boolean flag to decide whether the RPC's should fail or not, but as the RPCs happen in a different goroutine, that might happen to run on a different thread, the effects of the call to `s.setBlocking` might not be visible to the thread handling the RPCs. Thus a mutex is introduced into the mock service.

---
UDENG-2724

### DIFF
--- a/wsl-pro-service/internal/daemon/daemon_test.go
+++ b/wsl-pro-service/internal/daemon/daemon_test.go
@@ -179,8 +179,7 @@ func TestServe(t *testing.T) {
 					return systemd.gotState.Load() == "STATUS=Connected"
 				}, 30*time.Second, time.Second, "Systemd never switched states to 'Connected'")
 
-				require.Eventually(t, agent.Service.AllConnected,
-					30*time.Second, time.Second, "The daemon should have connected to the Windows Agent")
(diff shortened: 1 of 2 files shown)


didrocks (approved):
I didn’t spot anything suspicious! Thanks for looking at this!

PR #777, opened 2024-05-28
fix(wsl-pro-service): Find host address when NAT + DNS tunneling ON

CarlosNihelton:
WSL version 2.2.1 (pre-release) turns DNS tunneling ON by default (again). With DNS tunneling activated, the Windows host IP no longer matches the nameserver IP found in /etc/resolv.conf but still matches the default gateway. When that was the case the nameserver IP would contain a loopback address, but since that version it contains an IP address of type 10.0.0.0/8 . Attempting to connect to such address always fails. If that behavior remains, we should change the wsl-pro-service IP selection to rely on the default gateway IP address for all cases when networking mode is not “mirrored”, for which case it must remain at 127.0.0.1 .
That’s of now the preferred method to find the host IP address as documented in https://learn.microsoft.com/en-us/windows/wsl/networking#accessing-windows-networking-apps-from-linux-host-ip

The gist is: we no longer need to read `/etc/resolv.conf`.

---
UDENG-2761


didrocks (commented):
I am always thrilled to see a PR with +9 −106 :)

I have though one question: you are talking about the default change in WSL. However, it means that users can still switch it to off. What happens in those case? Shouldn’t the code detect that and fallback to `resolv.conf` in that case?

didrocks (approved):
Thanks for the rationale!

PR #778, opened 2024-05-29
docs: UP4W tutorial cleanup

edibotopic:
This is a **first pass** at improving the main UP4W tutorial.

The main changes here included:

- Edited all main text for clarity and focus
- Added more substantive and focused intro text
- Removed unnecessary and distracting syntax highlighting from CLI code blocks
- Truncated excessively long command logs
- Added shorter, more readable headings, resulting in a more concise contents sidebar
- Made the doc more consistent with the Canonical style guide (e.g., sentence case in headings, no comments in CLI codeblocks, no markup in headings)

The changes resulted in:

- 31% decrease in lines of text
- 15% decrease in page numbers (when rendered as PDF)
- 25% decrease in mouse scrolls from start to finish (estimated from manual testing)

UDENG-2814

### DIFF
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -35,7 +35,7 @@ for this tutorial.
 
 (ref::backup-warning)=
 ```{warning}
-**If you already have Ubuntu for WSL pre-installed:** 
+**If you already have Ubuntu WSL pre-installed:** 
 
 We recommend that they are exported then deleted.
 You can then install them as described in this tutorial.
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -12,7 +12,7 @@ Throughout this tutorial, commands will be prefixed by a prompt that indicates t
 
 - `PS C:\Users\me\tutorial>` is a PowerShell prompt where the current working directory is `C:\Users\me\tutorial`.
 
-- `root@hostname:~#` indicates a Linux shell prompt with login as root in the root home directory `~`.
+- `ubuntu@wsl:~/tutorial$` indicates a Linux shell prompt login as ubuntu where the current working directory is `/home/ubuntu/tutorial/`
 
 Output logs are included in this tutorial when instructive but are typically omitted to save space.
 ```
--- a/docs/tutorial/index.md
+++ b/docs/tutorial/index.md
@@ -7,42 +7,44 @@ In this tutorial you will develop an understanding of how UP4W can be installed 
 To complete this tutorial you will need a Windows 11 machine with a minimum of 16GB RAM and a 8-core processor.
 
 ```{note}
-Throughout this tutorial we'll present shell commands together with their output:
+WSL enables using a Linux shell and Windows PowerShell side-by-side on the same machine.
 
-- The commands will be prefixed by a
-prompt matching the shell it's launched in. 
+Throughout this tutorial, commands will be prefixed by a prompt that indicates the shell being used, for example:
 
-  For example, `PS C:\Users\me\tutorial>` is a PowerShell prompt where the current working directory is `C:\Users\me\tutorial`, thus running on Windows.
+- `PS C:\Users\me\tutorial>` is a PowerShell prompt where the current working directory is `C:\Users\me\tutorial`.
 
-  Similarly, a Linux shell prompt logged in as root with the current working directory as the root home directory will look like `root@hostname:~#`
-
-- The outputs are preserved to give you more context.
+- `root@hostname:~#` indicates a Linux shell prompt with login as root in the root home directory `~`.
 ```
+<!-- TODO: continue clean up -->
+<!-- focus on text before command logs -->
+<!-- keep Pro and UP4W together -->
+<!-- keep WSL/Ubuntu/Landscape instructions short -->
+<!-- landscape logs are way too much -->
+<!-- careful with links not rendering -->
 
 ## Set things up
 
-### Install the Windows Subsystem for Linux (WSL) application
+### Install WSL and Ubuntu
 
-```{warning}
-**If you already have it pre-installed:** 
+WSL can be installed directly from the [Microsoft Store](https://www.microsoft.com/store/productId/9P9TQF7MRM4R).
 
-Please reset its configuration to the default settings by (making a backup of your existing `~\.wslconfig` file, then) running:
+If you already have WSL installed you are advised to backup `~\.wslconfig` and remove it before continuing the tutorial:
 
-    PS C:\Users\me\tutorial> Remove-Item ~\.wslconfig
-
+```text
+PS C:\Users\me\tutorial> Remove-Item ~\.wslconfig
 ```
 
-See: [Microsoft Store | Windows Subsystem for Linux](https://www.microsoft.com/store/productId/9P9TQF7MRM4R)
-
-All set. From now on you can use it to launch  WSL instances.
-
-### Install the Ubuntu 22.04 LTS and Ubuntu (Preview) applications
+Ubuntu can also be installed from the Microsoft Store.
+Choose [Ubuntu 22.04 LTS](https://www.microsoft.com/store/productId/9PN20MSR04DW) or [Ubuntu (Preview)](https://www.microsoft.com/store/productId/9P7BDVKVNXZ6)
+for this tutorial.
 
 (ref::backup-warning)=
 ```{warning}
-**If you already have them pre-installed:** 
+**If you already have Ubuntu for WSL pre-installed:** 
 
-Please export them and delete them. Then proceed with the tutorial and install them fresh from the links provided below. At the end of the tutorial you can re-import them to restore your data.
+We recommend that they are exported then deleted.
+You can then install them as described in this tutorial.
+At the end of the tutorial you can re-import and restore your data.
 
  <details><summary> How do I export, delete, and re-import Ubuntu 22.04 LTS? </summary>
 
@@ -77,37 +79,13 @@ Please export them and delete them. Then proceed with the tutorial and install t
  </details>
 ```
 
-See: 
-- [Microsoft Store | Ubuntu 22.04 LTS](https://www.microsoft.com/store/productId/9PN20MSR04DW)
-- [Microsoft Store | Ubuntu (Preview)](https://www.microsoft.com/store/productId/9P7BDVKVNXZ6)
+You can now launch WSL instances of Ubuntu on your Windows machine.
 
-All set. From now on you can launch WSL instances of the Ubuntu 22.04 LTS and Ubuntu (Preview) releases.
+### Set up Landscape
 
-(tut::ensure-ubuntu-pro)=
-### Get an Ubuntu Pro subscription token
+While a Landscape server typically runs on external computers for this tutorial it will be set up on a WSL instance on your Windows machine.
 
-Visit the Ubuntu Pro page to get a subscription.
-
-> See more: [Ubuntu | Ubuntu Pro > Subscribe](https://ubuntu.com/pro/subscribe). If you choose the personal subscription option (`Myself`), the subscription is free for up to 5 machines. 
-
-Visit your Ubuntu Pro Dashboard to retrieve your subscription token.
-
-> See more: [Ubuntu | Ubuntu Pro > Dashboard](https://ubuntu.com/pro/dashboard)
-
-
-All set. From now on you can start adding this token to the Ubuntu Pro client on your WSL instances so you can enjoy Ubuntu Pro benefits on WSL as well. 
-
-### Set up a Landscape server
-
-```{warning}
-**If you already have one:**
-
-Please ignore it and proceed to set up a fresh one as shown below. As your version might differ from the one we assume here, this will guarantee the smoothest experience.
-```
-
-Landscape servers are usually on external computers. However, for the purpose of this tutorial we will set one up on a WSL instance on your Windows machine.
-
-In your Windows PowerShell, `shutdown` WSL, then install the Ubuntu 22.04 LTS instance with the `--root` option.
+In PowerShell, `shutdown` WSL then install the Ubuntu 22.04 LTS instance with the `--root` option.
 
 ```text
 # Ensure a clean WSL environment.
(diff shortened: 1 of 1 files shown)


docs/tutorial/index.md
```diff
@@ -289,24 +135,31 @@ admin account. Enter the following credentials and click the **Sign Up** button:
...
+### Get an Ubuntu Pro token
 
-Open a new terminal window and run `ubuntu2204.exe`. Landscape server and related components will start automatically.
-```
+An Ubuntu Pro subscription gives you a token that can be added to the Ubuntu Pro client on WSL instances.
+
+Visit the [Ubuntu Pro](https://ubuntu.com/pro/subscribe) page to get a subscription.
+The `Myself` option for a personal subscription is free for up to 5 machines. 
 
+Your subscription token can then be retrieved from the [Ubuntu Pro Dashboard](https://ubuntu.com/pro/dashboard).
```
CarlosNihelton, line 160: If the user already has an active subscription (personal or business) accessing the "subscribe" page will prompt them to creata yet another subscription, which might cost money or be misleading if there is already an active subscription in place. So, I'm not so sure about this, but I think it's better to guide the user to the "dashboard" first, and in case they don't have an active subscription, then a redirection to the "subscribe" page will make sense. Feel free to interject on this.

docs/tutorial/index.md
```diff
@@ -1,74 +1,48 @@
...
 
-### Install the `Windows Subsystem for Linux` (WSL) application
-
-```{warning}
-**If you already have it pre-installed:** 
-
-Please reset its configuration to the default settings by (making a backup of your existing `~\.wslconfig` file, then) running:
+### Install WSL and Ubuntu
 
+WSL can be installed directly from the [Microsoft Store](https://www.microsoft.com/store/productId/9P9TQF7MRM4R).
```
CarlosNihelton, line 24: Not exactly a problem with your PR, but I think we need to recommend upstreamm to increase the spacing between the headings, don't you think so?

image

docs/tutorial/index.md
```diff
@@ -316,32 +169,27 @@ Open a new terminal window and run `ubuntu2204.exe`. Landscape server and relate
...
 
-Time to install UP4W! Click on [this link](https://www.microsoft.com/store/productId/9PD1WZNBDXKZ), then on the big blue **Install** button.
+To install UP4W go to [this link to the Microsoft Store](https://www.microsoft.com/store/productId/9PD1WZNBDXKZ) and click **Install**.
 
 ![Install Ubuntu Pro for WSL from the Store](./assets/store.png)
 
+After installation has finished click **Start** to begin configuring UP4W.
 
-Once the installation is complete, you will see a **Start** button. Click it. That will start the UP4W Windows application. We'll use it to configure UP4W at next.
+% :NOTE: Instead of the GUI, it's possible to configure UP4W through the Windows registry, which enables you to do things at scale. This should be explained here or elsewhere.
```
CarlosNihelton, line 178: By commenting this out we remove the only place in this tutorial where we call the attention to the sysadmins that it's possible to do things at scale.

When I first wrote this tutorial, instead of using the GUI, I used the registry because I wanted to highlight that possibility.

Is there a better place, maybe earlier in this document, where we could emphasize that, even though this doc follows the GUI-workflow? Or, instead, you believe it's better to write a new document solely focused on the sysadmin? Per my understanding the ideal way to write a tutorial for sysadmins would require deploying the applications as they would (using some centralized deployment tool, such as MS Autopilot) but then we lack the tools inside Canonical to try, take screenshots etc as we are not a Windows software house :)

docs/tutorial/index.md
```diff
@@ -354,53 +202,46 @@ ssl_public_key = C:\Users\<YOUR_WINDOWS_USER_NAME>\landscape_server.pem
...
+Done! You can now close the UP4W window.
 
-This has attached your Ubuntu Pro subscription to UP4W on the Windows host; UP4W will automatically forward it to the Ubuntu Pro client on your Ubuntu WSL instances; thus, all of your Ubuntu WSL instances will be automatically added to your Ubuntu Pro subscription.
+Your Ubuntu Pro subscription is now attached to UP4W on the Windows host.
+UP4W will automatically forward the subscription to the Ubuntu Pro client on your Ubuntu WSL instances.
+This means that all Ubuntu WSL instances will be automatically added to your Ubuntu Pro subscription.
 
 This has also configured the Landscape client built into your UP4W Windows agent to know about your Landscape server; UP4W will forward this configuration to the Landscape client on your Ubuntu WSL instances as well; and all systems where the Landscape client has been configured this way are automatically registered with Landscape.
 
+### UP4W host registration
```
CarlosNihelton, line 221:
```suggestion
### UP4W host registration with Landscape
```

Stopping at the word registration feels a bit obscure to me: registration with/into what/where? The screenshots might be enough for the users acquainted with Landscape, but I don't think we should rely on that knowledge.

docs/tutorial/index.md
```diff
@@ -354,53 +202,46 @@ ssl_public_key = C:\Users\<YOUR_WINDOWS_USER_NAME>\landscape_server.pem
...
+At the top of the page, on the right-hand side of the Landscape logo, click on "Computers". You should see your host machine listed. Details such as the operating system may take some time to appear.
 
 ![Host and WSL instances in Landscape](./assets/host.png)
 
-All set. From now on you can use your Landscape server not just to manage the Ubuntu WSL instances that UP4W has pro-attached and Landscape-registered on the host, but to tell UP4W to create and provision them on the host too!
-
+Now you can leverage UP4W from your Landscape server to create and provision Ubuntu WSL instances on the host.
 
-## Watch UP4W transform your Ubuntu Pro game on WSL!
+## Deploy WSL instances
```
CarlosNihelton, line 237: No strong opinions here: Teodora's idea with this heading was to make it somewhat exciting, click-baity, more informal and less technical/cold. Did you consider this change in the documentation mood when doing this change?

docs/tutorial/index.md
```diff
@@ -455,62 +293,50 @@ Hit:5 http://ppa.launchpad.net/landscape/self-hosted-beta/ubuntu noble InRelease
...
 me@mib:~$
 ```
 
-As usual, this instance is pro-attached and Landscape-registered -- run `pro status` to verify the pro-attachment and refresh your Landscape server page to verify and accept the registration (as before, apply the  `wsl-vision` tag and click `Accept`).
-
+You can run `pro status` to verify pro-attachment and refresh your Landscape server page to verify and accept the registration.
+As before, apply the `wsl-vision` tag and click `Accept`.
 
-### Use Landscape to deploy packages to all of your Ubuntu WSL instances at once
+### Deploy packages to all Ubuntu WSL instances
```
CarlosNihelton, line 336: As before, the previous heading was meant to have some kind of marketing effect: if the reader is not familiar with Landscape, that title would immediately show off one of its core values. With the proposed changes we still show the value, but detach it from the Landscape product. No strong opinion here, but I'd like to make sure you considered that effect when doing the changes. WDYT?

CarlosNihelton (changes requested):
I like the way the text renders more concise and easier to read overall.

I have a few changes to request, though, and some questions for your consideration.

docs/tutorial/index.md
```diff
@@ -100,185 +74,57 @@ Please export them and delete them. Then proceed with the tutorial and install t
  </details>
 ```
 
+You can now launch WSL instances of Ubuntu on your Windows machine.
 
+### Set up Landscape
```
jibel, line 79: I'd remove this entire section from the tutorial and point to a Landscape tutorial if there is one.
Installing Landscape is unrelated to Ubuntu Pro for WSL.
If really we want to explain how to install Landscape in WSL, it should be a separate tutorial in WSL documentation.

docs/tutorial/index.md
```diff
@@ -1,74 +1,48 @@
...
-
-See: [Microsoft Store | Windows Subsystem for Linux](https://www.microsoft.com/store/productId/9P9TQF7MRM4R)
-
-All set. From now on you can use it to launch  WSL instances.
-
-<br />
-
-### Install the `Ubuntu 22.04 LTS` and `Ubuntu (Preview)` applications
+Ubuntu can also be installed from the Microsoft Store.
+Choose [Ubuntu 22.04 LTS](https://www.microsoft.com/store/productId/9PN20MSR04DW) or [Ubuntu (Preview)](https://www.microsoft.com/store/productId/9P7BDVKVNXZ6)
```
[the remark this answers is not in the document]
edibotopic: @jibel , I was going to look into updating aspects like Ubuntu versions in a later PR but can do this here also if it's preferable.

Question: does this tutorial require both an LTS and a Preview version to be used? It would be more simple to use 24.04 alone if possible.

docs/tutorial/index.md
edibotopic: @jibel , I agree and the Landscape section currently takes up too much space; however, I would like to leave it here for now.
The purpose of this PR was to streamline the tutorial and improve its style while leaving much of the technical content intact.
There are decisions to be made about restructuring the docs, such as moving material into separate tutorials.
It is better to do this incrementally because otherwise there are too many moving parts.

docs/tutorial/index.md
```diff
@@ -1,74 +1,48 @@
...
-
-- Set things up
-
-- Watch UP4W transform your Ubuntu Pro game on WSL!
-
-- Tear things down
-
-<br/>
+In this tutorial you will develop an understanding of how UP4W can be installed and deployed for managing multiple WSL instances.
+To complete this tutorial you will need a Windows 11 machine with a minimum of 16GB RAM and a 8-core processor.
```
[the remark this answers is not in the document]
edibotopic: I can look into this. Should not be a problem.

docs/tutorial/index.md
```diff
@@ -1,74 +1,48 @@
...
-prompt matching the shell it's launched in. 
-
-  For example, `PS C:\Users\me\tutorial>` is a PowerShell prompt where the current working directory is `C:\Users\me\tutorial`, thus running on Windows.
+WSL enables using a Linux shell and Windows PowerShell side-by-side on the same machine.
+Throughout this tutorial, commands will be prefixed by a prompt that indicates the shell being used, for example:
 
+- `PS C:\Users\me\tutorial>` is a PowerShell prompt where the current working directory is `C:\Users\me\tutorial`.
 
-  Similarly, a Linux shell prompt logged in as root with the current working directory as the root home directory will look like `root@hostname:~#`
+- `root@hostname:~#` indicates a Linux shell prompt with login as root in the root home directory `~`.
```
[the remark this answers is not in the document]
edibotopic: I will think about this and any implications for the rest of tutorial. We need to be careful about prompts and prompt symbols in general.
Typically it's not necessary to include prompts or prompt symbols and it can actually cause issues for some users who inadvertently copy/paste them.
In this tutorial it is probably necessary to distinguish between different shells and instances.

Using `/home/ubuntu/` instead of `~` could be a good option also.

docs/tutorial/index.md
edibotopic: To my eye I don't have a significant issue with the spacing; however, I agree that any spacing issues should ideally be resolved upstream.
Coming up with our own *ad hoc* solutions adds additional complexity to maintaining the doc and also could result in inconsistent styling across docs from different teams.
In general I would also suggest that we should avoid too many levels of heading. Often this only becomes a requirement when the doc becomes too complex for its own good. It also introduces additional problems with spacing of elements.

docs/tutorial/index.md
```diff
@@ -1,74 +1,48 @@
...
-
-### Install the `Ubuntu 22.04 LTS` and `Ubuntu (Preview)` applications
+Ubuntu can also be installed from the Microsoft Store.
+Choose [Ubuntu 22.04 LTS](https://www.microsoft.com/store/productId/9PN20MSR04DW) or [Ubuntu (Preview)](https://www.microsoft.com/store/productId/9P7BDVKVNXZ6)
+for this tutorial.
 
 (ref::backup-warning)=
 ```{warning}
-**If you already have them pre-installed:** 
+**If you already have Ubuntu for WSL pre-installed:**
```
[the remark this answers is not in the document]
edibotopic: Agreed.

docs/tutorial/index.md
edibotopic: Good thinking. The text can be revised to include a statement to the effect that if you already have a subscription you can go straight to the dashboard, and if you don't you can get a subscription at the link. Of course, this will add a slight bit of friction for those who don't have a subscription but - importantly - it won't be costly or misleading.

docs/tutorial/index.md
edibotopic: I completely understand your position; however, this single line did not provide enough context to warrant its inclusion. How the approach outlined in the tutorial and the one suggested in this line are different with respect to scalability is not sufficiently clear (a lot is assumed of the reader). When I revise this PR I can look for ways to edit for clarity.

The questions you raise about doing a separate tutorial are very important. We need to think about this and some of the complexities that you highlight, but that conversation is outside the scope of this PR.

A good first start might be find that earlier version of the tutorial where you used the registry.

docs/tutorial/index.md
edibotopic: Good point. In general, I was trying to shorten excessively long headers while highlighting the purpose of a given section. This tutorial also has a lot of references to different technology (UP4W, WSL, Ubuntu, Landscape,..), which could distract from what is being achieved in practice with UP4W.

In this case I agree that "with Landscape" should be preserved.

docs/tutorial/index.md
edibotopic: The basic issue with the "transform your game" phrase is that it is unclear. When someone quickly scans a table of contents they want to know where to find the information they need. This phrase assumes that the user knows that their Ubuntu Pro game needs transforming in the first place. Then it doesn't immediately suggest how the product will transform their game.

I am sympathetic for the need to add some personality to docs; however, this doc currently needs a lot of editing, clarifying and restructuring. During that process these quirks will be a distraction, will quickly become out-of-place and - worse - could be annoying to a user. A quirky/informal phrase will not make a tutorial less technical/cold if the tutorial is already highly complex and technical. There will be opportunities to add personality again but I think it's less of a priority right now.

Ultimately what I would like to aim for is personality that makes sense in the context of the product. Instead, of saying to the user that they can "transform their game" it would be good to show that we as a team are aware of the pain and difficulty faced by those doing admin for large fleets of devices and how this product can make their life easier.

docs/tutorial/index.md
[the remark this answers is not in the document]
jibel: No, this tutorial only requires 24.04 (until the solution is available on all the releases)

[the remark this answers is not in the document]
jibel: Well, there is landscape in this tutorial which makes the whole thing confusing. That's why Landscape would be better elsewhere.

docs/tutorial/index.md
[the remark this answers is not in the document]
jibel: I agree, maybe just > and $ to differentiate Windows and Linux. Or just no prompt but then it's difficult to make the difference between the OS and what is a command and an output of a command.

docs/tutorial/index.md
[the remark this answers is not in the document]
CarlosNihelton: I think we can settle to move the Landscape setup out of this in a subsequent pull request.

CarlosNihelton (approved):
Good work!

PR #780, opened 2024-06-05
fix(windows-agent): Find the rootfs checksum in the SHA256SUMS neighbour file

CarlosNihelton:
With the merge of https://github.com/canonical/landscape-hostagent-api/pull/16 Landscape will no longer send the SHA256 checksum of the rootfs image intended to be imported as part of the Install message as previously expected. Instead we rely on the assumption that the file server will to some extent mimic the structure we see in cloud-images.ubuntu.com, where one or more images are hosted alongside a SHA256SUMS file containing the checksums of all those images, as exemplified below.

``` 
SHA256SUMS                               2024-06-01 15:01  418   
SHA256SUMS.gpg                           2024-06-01 15:01  833   
ubuntu-jammy-wsl-amd64-wsl.manifest      2024-05-31 21:38   12K  Package manifest file
ubuntu-jammy-wsl-amd64-wsl.rootfs.tar.gz 2024-05-31 21:38  225M  File system image and Kernel packed
ubuntu-jammy-wsl-arm64-wsl.manifest      2024-05-31 21:47   12K  Package manifest file
ubuntu-jammy-wsl-arm64-wsl.rootfs.tar.gz 2024-05-31 21:47  215M  File system image and Kernel packed
[DIR] unpacked/                                2024-06-01 15:01    -  
```

With the SHA256SUMS file contents are the familiar structure exemplfied below:

```
1835cb92926046a640d23c4c66e72b55d28e62570045b6d0e228c5b2ec3b69aa  ubuntu-jammy-wsl-amd64-wsl.manifest
0aa21ea5b6da814bf615506e32d54d76c3f590f777cdc30048963cbac3d439fc  ubuntu-jammy-wsl-amd64-wsl.rootfs.tar.gz
7de7f1013b9bc05cde162a55784977f85f4639d4899870a476be417cce89ef9b  ubuntu-jammy-wsl-arm64-wsl.manifest
d98ff8f636c47d93c351cc16acccca72f13ba77695a8fceea88dd5f36d567bb2  ubuntu-jammy-wsl-arm64-wsl.rootfs.tar.gz
```
(The filename at the right side of the checksum value could contain an asterisk, as seen in the cloud-images, or not, as in the WSL images).

This PR updates the Landscape API dependency to the latest main to match the changes mentioned above, and adapt the agent's Landscape command executor to deal with the changes. We no longer pass the rootfs URL blindly as a string, but rather parse it as URL so we can more easily compute the URL of the neightbour SHA256SUMS file. We try to download it upfront, and bail out if that file doesn't exist. Otherwise we acquire the expected checksum so we can preserve the logic to validate the image as we write it to the disk.

Finally, for the that installation scenario, we validate the provided distro ID to avoid clashing with the ones from the applications provided via the MS Store.

---
UDENG-3055


windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,47 +148,40 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
+		"Error when the distroname is empty":         {distroName: ""},
 
-		"Error when the rootfs isn't a valid tarball":                   {sendRootfsURL: "badfile", sendRootfsChecksum: mockErrorChecksum, wantInstalled: false},
-		"Error when the checksum doesn't match":                         {sendRootfsURL: "goodfile", sendRootfsChecksum: mockMismatchChecksum, wantInstalled: false},
+		"Error when the distro ID is reserved":                          {sendRootfsURL: "goodfile", distroName: "Ubuntu-19.13", wantInstalled: false},
+		"Error when the rootfs isn't a valid tarball":                   {sendRootfsURL: "badfile", wantInstalled: false},
+		"Error when the checksum file doesn't exist":                    {missingChecksums: true, sendRootfsURL: "goodfile", wantInstalled: false},
+		"Error when the checksum doesn't match":                         {sendRootfsURL: "badchecksum", wantInstalled: false},
 		"Error when the rootfs doesn't exist":                           {sendRootfsURL: "badresponse", wantInstalled: false},
+		"Error when the rootfs URL is ill-formed":                       {sendRootfsURL: "https://?", wantInstalled: false},
```
didrocks, line 184: I am surprise this didn’t trigger url.Parse() erroring on line 162 of `executor.go`. Maybe this is an ill-formed that fails `url.Get` but not the parsing part mind checking? :)

I understand that line 346 can’t be covered (as we will fail on the rootfs url before the sha256 sums one), but I think 162 should be doable.

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -141,10 +145,6 @@ func (e executor) install(ctx context.Context, cmd *landscapeapi.Command_Install
 		return fmt.Errorf("skipped installation: %v", err)
 	}
 
-	if err := gowsl.Install(ctx, distro.Name()); err != nil {
```
didrocks, line 144: stupid question, probably hidden by the diff, but where is the distro install code now when there is no rootfs url provided? Was that code duplicated here?

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -305,6 +322,52 @@ func download(ctx context.Context, f io.Writer, url, checksum string) (err error
...
+// 103ee8b5693bdb7c23a378453c624d8605445eb07e2e550d3fad831da865f5ea *jammy-server-cloudimg-riscv64.release.20240514.20240601.image_changelog.json
+// 1eaa1df5794122e3419c963d88f043121c164936b9b828adac650c9f5e22c3e6 *jammy-server-cloudimg-amd64.img
+// 1fcd2edf4fda78e0a6f3bc0c3684286c29371e4dd7863a59b39d2cfcff79b5e1 *jammy-server-cloudimg-amd64-root.manifest
+// 1fcd2edf4fda78e0a6f3bc0c3684286c29371e4dd7863a59b39d2cfcff79b5e1 *jammy-server-cloudimg-amd64.squashfs.manifest
+// 2646292d657f4c9ef5dfce804a5a1e66d8c1324c74147b8bc9b1bf154d7feaf8 *jammy-server-cloudimg-arm64-root.tar.xz
+//
+// ...
+func wantRootfsChecksum(u *url.URL) (string, error) {
+	imageName := filepath.Base(u.Path)
+	shasRelativeLocation, err := url.Parse("../SHA256SUMS")
```
didrocks, line 344:
```suggestion
	shasRelativeURL, err := url.Parse("../SHA256SUMS")
```

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -305,6 +322,52 @@ func download(ctx context.Context, f io.Writer, url, checksum string) (err error
...
+	if err != nil {
+		return "", fmt.Errorf("could not assemble SHA256SUMS location: %v", err)
+	}
+	checksumsURL := u.ResolveReference(shasRelativeLocation)
+
+	resp, err := http.Get(checksumsURL.String())
+	if err != nil {
+		return "", fmt.Errorf("could not download checksums file <%s>: %v", checksumsURL, err)
```
didrocks, line 352: rather than <%s>, you can use %q so that it’s quoted, which is more standard I think than having it with `<>`?

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -305,6 +322,52 @@ func download(ctx context.Context, f io.Writer, url, checksum string) (err error
...
+	}
+	defer resp.Body.Close()
```
didrocks, line 354: I think we told that the checksum file is optional, but if there is one, your file should be in, didn’t we?

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,47 +148,40 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
-	emptyFileChecksum := "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
-	mockErrorChecksum := "afe55cda4210c2439b47c62c01039027522f7ed4abdb113972b3030b3359532a"
-	mockMismatchChecksum := "1234"
-
 	testCases := map[string]struct {
 		noCloudInit            bool
 		cloudInitWriteErr      bool
 		distroAlreadyInstalled bool
-		emptyDistroName        bool
+		distroName             interface{} // using interface{} to allow nil, modelling an optional value.
```
didrocks, line 155: I would rather use a pointer then. But in general, we do something else in all our projects:
```
"" is the default
"-" means we replace with the empty value, so empty string for a string
```

For instance https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/common/i18n/i18n_test.go#L115

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,47 +148,40 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
 		"Error when the distro is already installed": {distroAlreadyInstalled: true, wantInstalled: true},
 		"Error when the distro fails to install":     {wslInstallErr: true},
 		"Error when cannot write cloud-init file":    {cloudInitWriteErr: true, wantCouldInitWriteCalled: true},
+		"Error when the distroname is empty":         {distroName: ""},
 
-		"Error when the rootfs isn't a valid tarball":                   {sendRootfsURL: "badfile", sendRootfsChecksum: mockErrorChecksum, wantInstalled: false},
-		"Error when the checksum doesn't match":                         {sendRootfsURL: "goodfile", sendRootfsChecksum: mockMismatchChecksum, wantInstalled: false},
+		"Error when the distro ID is reserved":                          {sendRootfsURL: "goodfile", distroName: "Ubuntu-19.13", wantInstalled: false},
+		"Error when the rootfs isn't a valid tarball":                   {sendRootfsURL: "badfile", wantInstalled: false},
+		"Error when the checksum file doesn't exist":                    {missingChecksums: true, sendRootfsURL: "goodfile", wantInstalled: false},
```
didrocks, line 181: I think we shouldn’t require the checksum fail, just issue a warning if we couldn’t find one (maybe we should differentiate 404 with other errors).

However, there is a test case that if there is a checksum file, it should list that rootfs or error out (you can have this with no extra line by return "" with no default if not found. Then, we would compare against this and "oh, no match because it’s empty" (which really means there is no entry)

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -305,6 +322,52 @@ func download(ctx context.Context, f io.Writer, url, checksum string) (err error
...
+		fields := strings.Fields(scanner.Text())
+		if len(fields) != 2 {
+			continue
+		}
+
+		if strings.TrimPrefix(fields[1], "*") == imageName && len(fields[0]) > 0 {
+			return fields[0], nil
+		}
+	}
+
```
didrocks, line 367: you need to do a check for scanner.Err() after scanning the file (especially if we are not going to return an error for no match, see below on the testcase)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,47 +148,40 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
-		"From a rootfs URL with a checksum": {sendRootfsURL: "goodfile", sendRootfsChecksum: emptyFileChecksum, wantInstalled: true},
+		"From a rootfs URL with a checksum": {sendRootfsURL: "goodfile", wantInstalled: true},
 		"With no cloud-init":                {noCloudInit: true, wantCouldInitWriteCalled: true, wantInstalled: true},
 
-		"Error when the distroname is empty":         {emptyDistroName: true},
 		"Error when the Appx does not exist":         {appxDoesNotExist: true},
 		"Error when the distro is already installed": {distroAlreadyInstalled: true, wantInstalled: true},
 		"Error when the distro fails to install":     {wslInstallErr: true},
 		"Error when cannot write cloud-init file":    {cloudInitWriteErr: true, wantCouldInitWriteCalled: true},
+		"Error when the distroname is empty":         {distroName: ""},
```
didrocks, line 177: why moving it below?

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,47 +148,40 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
 
-		"Error when the rootfs isn't a valid tarball":                   {sendRootfsURL: "badfile", sendRootfsChecksum: mockErrorChecksum, wantInstalled: false},
-		"Error when the checksum doesn't match":                         {sendRootfsURL: "goodfile", sendRootfsChecksum: mockMismatchChecksum, wantInstalled: false},
+		"Error when the distro ID is reserved":                          {sendRootfsURL: "goodfile", distroName: "Ubuntu-19.13", wantInstalled: false},
```
didrocks, line 179: I would probably here make the 3 cases explicit (even if it’s a little bit redundant):
- Ubuntu
- Ubuntu-XX.YY
- Ubuntu-preview

And also ensure it’s case insensitive (one test for that last one should be enough).

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -205,7 +198,13 @@ func TestInstall(t *testing.T) {
 
 			if tc.appxDoesNotExist || tc.sendRootfsURL != "" {
 				// WSLMock Install only accepts ubuntu-22.04
-				settings.name = wsltestutils.RandomDistroName(t)
+				if tc.distroName == nil {
+					settings.name = wsltestutils.RandomDistroName(t)
```
didrocks, line 202: just checking, but no chance that the random name will match a reserved one?

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -205,7 +198,13 @@ func TestInstall(t *testing.T) {
...
+				} else {
+					n, ok := tc.distroName.(string)
+					require.True(t, ok, "Setup: tc.distroName should be a string")
```
didrocks, line 205: yeah, that would simplify all this to use the other pattern for empty/default I explained below.

didrocks (changes requested):
Nice work! I have a few comments on place to improve. Logically-wise, apart from the missing sha256sums file, everything is there. I have some suggestions on the testing side, and some errors/cases that are not checked.

One additional point: IIRC, we have a landscape mock for integration tests that allows to send grpc command to us. I thought that one would be updated too or did it miss this "Install" command?
It would be a great way for manual testing until landscape is ready too :)

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: Makes sense. Is there a reason for not using a ´switch` instead of if-elses since at least three different cases are expected? (`""`, `"-"` and anything else) ?

CarlosNihelton: I think a `switch` reads better:

```go

			if tc.appxDoesNotExist || tc.sendRootfsURL != "" {
				// WSLMock Install only accepts ubuntu-22.04
				switch tc.distroName {
				case "-":
					settings.name = ""
				case "":
					settings.name = wsltestutils.RandomDistroName(t)
				default:
					settings.name = tc.distroName
				}
			}
```

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: I wanted it closer to the other test cases that depend on the distro name/ID. But I think my original idea doesn't fit anymore, as they are totally different topics.

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: Indeed it didn't break the URL parsing. That's because the URL always started with a valid scheme and address (coming from the `fileServerAddr` variable (assigned from the listener address, so always valid). I had to make that variable hold invalid contents to make `url.Parse` really break..

windows-agent/internal/proservices/landscape/executor.go
CarlosNihelton: I suspect that's a leftover:

Looking at main: inside that method we have such call, and below it we have a check whether we have a URL or store ID, based on which we call installFromURL or installFromMicrosoftStore: https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/windows-agent/internal/proservices/landscape/executor.go#L159-L167

Inside the later, we have that exact same line:

https://github.com/canonical/ubuntu-pro-for-wsl/blob/main/windows-agent/internal/proservices/landscape/executor.go#L226

We don't want to unconditionally install from the Store without the URL validation.

windows-agent/internal/proservices/landscape/executor.go
CarlosNihelton: Oh, yeah:

> it then looks up for a “SHA256SUMS” file in the same directory than the one the rootfs refers to, downloads it if available, ensures the rootfs base url is in that file, and if so, checks its sum to prevent corrupted downloads.

My bad :(

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: That's the pattern returned by `wsltestutils.RandomDistroName()`:
`return fmt.Sprintf("%s_%s_%d", testDistroPrefix, name, rand.Uint64())`
So, no way to match our reserved names unless we change that drastically.
I'm adding a comment about that just in case that test helper changes in the future, we're advised. I don't think that duplicating the reserved names check is worth the trouble.

windows-agent/internal/proservices/landscape/executor_test.go
CarlosNihelton: Well, since the checksums file is optional, we should `return "", nil` if that file is not found but `return "", ERROR` if the file exists but does not contain an entry for the rootfs being checked in `wantRootfsChecksum`, this way we know and react accordingly if the checksum string is empty.

CarlosNihelton:
Rebased to the latest main. New commits start at: https://github.com/canonical/ubuntu-pro-for-wsl/pull/780/commits/a8f081c0a85e6748ea1d0d43b1512c2a0d4c8ca2 .

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -148,51 +149,50 @@ func TestReceiveCommandStartStop(t *testing.T) {
...
+		"Error when the distro ID is reserved (case sensitiveness)":       {sendRootfsURL: "goodfile", distroName: "ubuntu-preview", wantInstalled: false},
+		"Error when the distro ID is reserved (release numbers)":          {sendRootfsURL: "goodfile", distroName: "Ubuntu-19.13", wantInstalled: false},
+		"Error when the distro ID is reserved (release numbers and case)": {sendRootfsURL: "goodfile", distroName: "uBuntu-19.13", wantInstalled: false},
+		"Error when the rootfs isn't a valid tarball":                     {sendRootfsURL: "badfile", wantInstalled: false},
+		"Error when the checksum doesn't match":                           {sendRootfsURL: "badchecksum", wantInstalled: false},
+		"Error when the checksum is missing for the rootfs":               {sendRootfsURL: "rootfswithnochecksum", wantInstalled: false},
+		"Error when the rootfs doesn't exist":                             {sendRootfsURL: "badresponse", wantInstalled: false},
+		"Error when the rootfs URL is ill-formed":                         {breakURL: true, sendRootfsURL: "goodfile", wantInstalled: false},
```
didrocks, line 191: I find it puzzling you are breaking the url while stil using a `sendRootfsURL` parameter in the same test. Maybe you can just use a hardcoded value like `@:?/BROKEN_URL`, and test this to not `url.JoinPath` in the test and take it verbatim?

windows-agent/internal/proservices/landscape/executor_test.go
didrocks: exactly! I’m resolving this one but adding a comment on another approach we should take to avoid having another parameter + having to use a parameter that is ignored then.

windows-agent/internal/proservices/landscape/executor_test.go
didrocks: It depends on the context, when we have 2, this is fine. I still like the pattern in general to set the default, then handle special case(s). If there is only one, this is a no-brainer, as here there are 2 + one default, this is debatable. Happy with your solution :)

windows-agent/internal/proservices/landscape/executor_test.go
didrocks: Agreed!

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -348,9 +348,14 @@ func wantRootfsChecksum(u *url.URL) (string, error) {
 	checksumsURL := u.ResolveReference(shasRelativeURL)
 
 	resp, err := http.Get(checksumsURL.String())
+	// Errors here are protocol errors, not 404s. "A non-2xx response doesn't cause an error".
```
didrocks, line 351: Thanks for the comment! :)

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -348,9 +348,14 @@ func wantRootfsChecksum(u *url.URL) (string, error) {
...
 	if err != nil {
 		return "", fmt.Errorf("could not download checksums file %q: %v", checksumsURL, err)
 	}
+	if resp.StatusCode == http.StatusNotFound {
+		log.Warningf(ctx, "expected checksums file %q not found", checksumsURL)
```
didrocks, line 356: I would degrade that to an Infof and replace `expected` with nothing.

(if it was really expceted, that would be an error and we said we won’t enforce it)

windows-agent/internal/proservices/landscape/executor_test.go
```diff
@@ -184,8 +185,8 @@ func TestInstall(t *testing.T) {
 		"Error when the distro ID is reserved (release numbers)":          {sendRootfsURL: "goodfile", distroName: "Ubuntu-19.13", wantInstalled: false},
 		"Error when the distro ID is reserved (release numbers and case)": {sendRootfsURL: "goodfile", distroName: "uBuntu-19.13", wantInstalled: false},
 		"Error when the rootfs isn't a valid tarball":                     {sendRootfsURL: "badfile", wantInstalled: false},
-		"Error when the checksum file doesn't exist":                      {missingChecksums: true, sendRootfsURL: "goodfile", wantInstalled: false},
 		"Error when the checksum doesn't match":                           {sendRootfsURL: "badchecksum", wantInstalled: false},
+		"Error when the checksum is missing for the rootfs":               {sendRootfsURL: "rootfswithnochecksum", wantInstalled: false},
```
didrocks, line 189: `checksum entry`? (to differentiate with "checksum file" when quickly scanning the tests)

windows-agent/internal/proservices/landscape/executor.go
didrocks: Indeed, that was clearly a leftover. Good riddance :)

windows-agent/internal/proservices/landscape/executor_test.go
didrocks: Sorry, I wrote warning here, but it should be info looking at the context, just wrote a comment on it :)

Agreed on the comment!

mocks/landscape/landscaperepl/app.go
```diff
@@ -76,6 +78,31 @@ on your command line. Hosted at the specified address.`,
 			}()
 			defer server.Stop()
 
+			fileserverDir := cmd.Flag("fileserver-dir").Value.String()
```
didrocks, line 81: This is a very good idea and will save up some cycles. I wonder if we should though not just serve cwd when starting and not messing with an option for simplicity, wdyt?

didrocks (changes requested):
Excellent update! I have a couple of comments remaining on the new changes and some suggestions, but we should be really close to merge :)

Let me know what you think!

mocks/landscape/landscaperepl/app.go
CarlosNihelton: I consider the command line flag simple enough.

The alternative requires either one remembering to `cd` first and then run the landscaperepl from that location, which requires invoking if with its absolute path or store the intended files to be served alongside the code.

Finally. with the command line flag we can avoid serving files if we so desire.

didrocks: Right, but if 90% of the cases of usage requires the flag, then, it’s a little bit of a pity to have to request that.

Also, what happens in the mock if you use the API to request a rootfs but didn’t use the flag?

CarlosNihelton: Nothing would happen to the mock, but the agent would fail to download since we'd send URLs pointing to nowhere.
I'm removing the CLI based on the argument that 90% of the cases the REPL will run requires a file server.

PR #789, opened 2024-06-06
docs: Format landing page contents

edibotopic:
This adds an "In this documentation" section to the landing page that aligns with the Diataxis framework and other product documentation.

A similar content block was removed from the main tutorial and replaced with text that includes internal links to the docs.
This content block was non-functional (there were no working links) and it would have duplicated what is now on the landing page.

UDENG-3001

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -11,6 +11,37 @@ UP4W can be installed from the [Microsoft Store](https://apps.microsoft.com/home
 
 WSL is preferred by many organisations as a solution to run a fully-functioning Linux environment on a Windows machine. UP4W empowers system administrators and corporate security teams to manage these WSL instances at scale.
 
+## In this documentation
+
+````{grid} 1 1 2 2
+
+```{grid-item-card} [Tutorial](tutorial/index)
+
+**Start here** with a hands-on introduction for new users, guiding you through your first-steps
+```
+
+```{grid-item-card} [How-to guides](howto/index)
+
+**Follow step-by-step** instructions for key operations and common tasks
+```
+
+````
+
+````{grid} 1 1 2 2
+:reverse:
+
+```{grid-item-card} [Reference](reference/index)
+
+**Read technical descriptions** of important factual information relating to UP4W
+```
+
+```{grid-item-card} [UP4W Dev](dev/index)
+
+**Review guides and reference material** aimed at developers
+```
+
+````
+
 ## Project and community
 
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -11,6 +11,37 @@ UP4W can be installed from the [Microsoft Store](https://apps.microsoft.com/home
...
+:reverse:
+
+```{grid-item-card} [Reference](reference/index)
+
+**Read technical descriptions** of important factual information relating to UP4W
+```
+
+```{grid-item-card} [UP4W Dev](dev/index)
+
+**Review guides and reference material** aimed at developers
```
CarlosNihelton, line 40: We should clarify that by developers we mean contributors to this project.

docs/index.md
```diff
@@ -11,6 +11,37 @@ UP4W can be installed from the [Microsoft Store](https://apps.microsoft.com/home
 
 WSL is preferred by many organisations as a solution to run a fully-functioning Linux environment on a Windows machine. UP4W empowers system administrators and corporate security teams to manage these WSL instances at scale.
```
CarlosNihelton, line 12: It was brought to my attention an issue in the following sentence from the previous paragraph:

> When the Landscape client is installed on the host, any connected WSL instances will be auto-enrolled in Landscape.

It suggests that there is some sort of additional Landscape client we install on Windows, which is not quite the case. UP4W is such a client and the statement should clarify that once we configure UP4W with Landscape information then any connected WSL instance will be auto-enroleld in Landscape. I'd rephrase it like:

> When UP4W is configured with the necessary information about Landscape, any connected WSL instances will be auto-enrolled in Landscape.

As it's unrelated to the changes fo this PR, feel free to bring it into a new PR or augment this one.

CarlosNihelton: I didn't comment in the proper line because it's outside of the context GitHub UI allows commenting on.

jibel (approved):
LGTM, Thanks!

docs/index.md
edibotopic: Thanks Carlos -- good point. I will change to "aimed at contributors". For now, I think we should keep "UP4W Dev" for that section of the docs. Ultimately, the docs for contributors may not need to be in a separate place; if they are clearly labelled and articulated they should be able to go alongside the regular docs but that's another conversation for another day.

docs/index.md
edibotopic: That's great Carlos -- thanks. There is some important content revision work that needs to be done in line with the testing work. I will note this suggestion and incorporate it into a later PR.

PR #812, opened 2024-06-20
fix: implement startup-time fixes for landscape.conf

didrocks:
The issue is the fact that the provisioning of new WSL instances rely on cloud-init to deploy UP4W key matters: the pro token and Landscape configuration. While the first is global to all WSL instances, Landscape client config has pieces of data that are sensitive to a particular instance and subject to change between boots.

UDENG-3128


wsl-pro-service/internal/system/landscape.go
```diff
@@ -18,31 +19,65 @@ const (
...
-		return fmt.Errorf("could not enable Landscape: %v", err)
+	// No change to do, do not rewrite config
+	if !enableWithoutModification && modifiedLandscapeConfig == landscapeConfig {
+		log.Debug(ctx, "Landscape configuration is already valid")
+		return nil
 	}
 
-	return nil
-}
+	if err := s.writeConfig(modifiedLandscapeConfig); err != nil {
```
CarlosNihelton, line 73: if `enableWithoutModification` is true and `modifiedLandscapeConfig == landscapeConfig` is false we write the `modifiedLandscapeConfig` somewhat defeating the purpose of `enableWithoutModification` flag, unless it's name is not accurate. Can you double-check?

wsl-pro-service/internal/system/landscape.go
```diff
@@ -86,94 +121,60 @@ func (s *System) writeConfig(landscapeConfig string) (err error) {
...
 		return nil
 	}
 
-	k, err := sec.GetKey(key)
-	if err != nil {
-		// No certificate
+	pathWindows := k.String()
+
+	if !strings.Contains(pathWindows, `\`) {
+		// Assuming this is a linux path already
```
CarlosNihelton, line 174: This is somewhat misleading. Windows paths can be written with forward-slashes, which complicates things. I think we should invoke wslpath unconditionally. An alternative would be to detect the regex `^\w:[\\\/].+` , assuming we only deal with absolute paths (Windows relative paths written with forward-slashes are Linux relative paths - any tool would break :) )

Here I miss Python's pathlib, where I can instantiate `PurePath` independent of the OS I'm working on.

wsl-pro-service/internal/system/system_test.go
```diff
@@ -22,6 +22,41 @@ const (
...
+
+			mock := &testutils.SystemMock{
+				FsRoot: testutils.MockFilesystemRoot(t),
+			}
+
+			if tc.systemLandscapeConfigFile != "" {
+				config, err := os.ReadFile(filepath.Join("testdata", "landscape.conf.d", tc.systemLandscapeConfigFile))
+				require.NoError(t, err, "Setup: could not load fixture")
+				err = os.MkdirAll(filepath.Dir(mock.Path(system.LandscapeConfigPath)), 0700)
+				require.NoError(t, err, "Setup: could not create landscape config dir")
```
CarlosNihelton, line 48: nit: I think we should capitalize the word "Landscape" in comments and logs/strings.

wsl-pro-service/internal/system/system_test.go
```diff
@@ -22,6 +22,41 @@ const (
...
+				err = os.WriteFile(mock.Path(system.LandscapeConfigPath), config, 0600)
+				require.NoError(t, err, "Setup: could not write landscape system config file")
```
CarlosNihelton, line 50: Ditto

wsl-pro-service/internal/system/system_test.go
```diff
@@ -22,6 +22,41 @@ const (
...
 
+func TestNew(t *testing.T) {
+	t.Parallel()
+
+	testCases := map[string]struct {
+		systemLandscapeConfigFile string
+	}{
+		"Return a new system": {},
+
+		"Only prints a warning when the landscape config validation failed": {systemLandscapeConfigFile: "invalid_ini.conf"},
```
CarlosNihelton, line 33: I don't see where we assert the warning is printed, as the test case name suggests. Can you clarify?

wsl-pro-service/internal/system/system_test.go
```diff
@@ -390,23 +425,34 @@ func TestLandscapeEnable(t *testing.T) {
...
-		"Success overriding computer_title":                           {},
-		"Success overriding the SSL certficate path":                  {},
-		"Do not append wsl tag when config tag is provided":           {},
-		"Do not append wsl tag when config tag is provided but empty": {},
-
-		"Error when the file cannot be parsed":                   {wantErr: true},
-		"Error when the config file cannot be written":           {breakWriteConfig: true, wantErr: true},
-		"Error when the landscape-config command fails":          {breakLandscapeConfig: true, wantErr: true},
+		"Appends any required fields - no ssl":   {landscapeConfigFile: "minimal.conf"},
+		"Transform Windows SSL certificate path": {landscapeConfigFile: "windows_ssl_only.conf"},
```
CarlosNihelton, line 440: As said before about Windows paths with forward slashes is accepted, we should add a test case and fixture written that way, either here or in `TestEnsureValidLandscapeConfig` (maybe even beter, since it's "closer" to the call that must handle that).

wsl-pro-service/internal/system/landscape.go
CarlosNihelton: Also we should have a test case for that.

wsl-pro-service/internal/system/system_test.go
```diff
@@ -390,23 +425,34 @@ func TestLandscapeEnable(t *testing.T) {
...
+		"Refresh computer_title if changed":      {landscapeConfigFile: "old_computer_title.conf"},
+
+		"Regular with additional keys":            {landscapeConfigFile: "regular.conf"},
+		"Do not modify other sections and keys":   {landscapeConfigFile: "regular_with_extra_keys.conf"},
+		"Reformat landscape config to proper ini": {landscapeConfigFile: "regular_with_weird_format.conf"},
+
+		"Rerun landscape even without modifications": {landscapeConfigFile: "no_change_needed.conf"},
```
CarlosNihelton, line 447: I see what you mean about test case names storytelling. 👍🏽

wsl-pro-service/internal/system/system_test.go
```diff
@@ -390,23 +425,34 @@ func TestLandscapeEnable(t *testing.T) {
...
-		"Error when the file cannot be parsed":                   {wantErr: true},
-		"Error when the config file cannot be written":           {breakWriteConfig: true, wantErr: true},
-		"Error when the landscape-config command fails":          {breakLandscapeConfig: true, wantErr: true},
+		"Appends any required fields - no ssl":   {landscapeConfigFile: "minimal.conf"},
+		"Transform Windows SSL certificate path": {landscapeConfigFile: "windows_ssl_only.conf"},
+		"Refresh computer_title if changed":      {landscapeConfigFile: "old_computer_title.conf"},
+
+		"Regular with additional keys":            {landscapeConfigFile: "regular.conf"},
+		"Do not modify other sections and keys":   {landscapeConfigFile: "regular_with_extra_keys.conf"},
+		"Reformat landscape config to proper ini": {landscapeConfigFile: "regular_with_weird_format.conf"},
```
CarlosNihelton, line 445: Ditto about the capitalization of "Landscape"

wsl-pro-service/internal/system/system.go
```diff
@@ -70,9 +71,15 @@ func New(args ...Option) *System {
 		f(&opts)
 	}
 
-	return &System{
+	s := &System{
 		backend: opts.backend,
 	}
+
+	if err := s.EnsureValidLandscapeConfig(context.Background()); err != nil {
```
CarlosNihelton, line 78: So here we ensure the fixes are done at startup-time, right?

wsl-pro-service/internal/system/system_test.go
```diff
@@ -585,6 +649,119 @@ func TestLandscapeDisable(t *testing.T) {
...
+		breakWSLPath            bool
+		breakWSLDistroName      bool
+		noLandscapeGroup        bool
+
+		wantNoLandscapeConfig    bool
+		wantNoLandscapeConfigCmd bool
+		wantErr                  bool
+	}{
+		"Appends any required fields - no ssl":   {systemLandscapeConfigFile: "minimal.conf"},
+		"Transform Windows SSL certificate path": {systemLandscapeConfigFile: "windows_ssl_only.conf"},
```
CarlosNihelton, line 669: Ditto for Windows paths with forward slashes.

wsl-pro-service/internal/system/system_test.go
```diff
@@ -585,6 +649,119 @@ func TestLandscapeDisable(t *testing.T) {
...
+		"Refresh computer_title if changed":      {systemLandscapeConfigFile: "old_computer_title.conf"},
+
+		"Regular with additional keys":            {systemLandscapeConfigFile: "regular.conf"},
+		"Do not modify other sections and keys":   {systemLandscapeConfigFile: "regular_with_extra_keys.conf"},
+		"Reformat landscape config to proper ini": {systemLandscapeConfigFile: "regular_with_weird_format.conf"},
+
+		"Do not rerun landscape without modifications":                             {systemLandscapeConfigFile: "no_change_needed.conf", wantNoLandscapeConfigCmd: true},
+		"No landscape configuration means no landscape command nor config created": {systemLandscapeConfigFile: "-", wantNoLandscapeConfigCmd: true, wantNoLandscapeConfig: true},
```
CarlosNihelton, line 677: Ditto for "Landscape" capitalization :p

CarlosNihelton (changes requested):
Thanks for getting here so quickly! I have some minor issues to report. Let me know what you think.

wsl-pro-service/internal/system/landscape.go
didrocks: We had a test case with a linux path and windows path (see coverage). But ok, it seems we’ll have to call wslpath with such a case for C://something. I’ll do it.

wsl-pro-service/internal/system/system_test.go
didrocks: We can’t capture logs with the current test setup apart of not running this subtest in parallel. What I wanted to emphasize was that we don’t return an error (and we can see logs when running the tests) and describe its behaviour. I can capture it and ensure we print something but that seems overkill. WDYT?

wsl-pro-service/internal/system/system_test.go
didrocks: agreed

wsl-pro-service/internal/system/system_test.go
didrocks: :book:

wsl-pro-service/internal/system/system.go
didrocks: Exactly! We initialize the "system" and so this at startup.

wsl-pro-service/internal/system/system_test.go
CarlosNihelton: Yeap, overkill certainly. 

I'd just rephrase the title to enforce that what we are checking here is the absence of errors: "Sucess when Landscape config validation failed (only warnings)". I know "Success" here is less precise, "No errors when ..." is not the best alternative because it starts with a negation, "(only warnings)" in parenthesis (or commas) should de-emphasize it compared to have that in the beginning of the sentence. If alternatives like that don't seem to capture the message you want to pass, then we should add a comment in the end of test body so we don't think about the same question in the future  :)

wsl-pro-service/internal/system/landscape.go
CarlosNihelton: The name of that variable implied (to me) that we would take the action of enabling Landsacpe without modifying the configuration. That's why I raised the question about the intent, which now I see is not what I understood, but rather a flag to tell us whether we should run landscape-config **even when** there are no modifications. So, might `enableEvenWithoutModification` or `enableUnconditionally` be a better name?

didrocks: I like `enableUnconditionally`! Will grab this :)

wsl-pro-service/internal/system/system_test.go
didrocks: What about `Ignore errors when the landscape config validation failed (only warnings)`?

wsl-pro-service/internal/system/system_test.go
didrocks: (Note, I only did it for the configuration, I kept the user and command as `landscape` in that context)

CarlosNihelton (approved):
Great! Thanks for addressing this!

PR #815, opened 2024-06-25
feat(gui): Hide Landscape config page by default

CarlosNihelton:
The key point of this change is: Landscape config page is disabled by default and can be enabled by setting the environment variable `UP4W_ALLOW_LANDSCAPE_INTEGRATION`.

Reverting this behavior in the future is quite easy, changing the semantics of the environment variable in `app.dart` and `ui.go` is just enough, ofc changing the environment variable name is also desirable.

We set that environment variable for integration tests so we make sure to keep that feature maintained and upgraded, preventing surprises when we eventually decide to plug it in by default.
More tests were added to make sure individual components (models/views) behave well with and without that feature enabled.
I put an assertion on the agent as well, to make sure it rejects RPC calls to `ApplyLandscapeConfig` if that environment variable is not set. Notice that nothing prevents the usage of the registry, empowering advanced users.

---
UDENG-3360


CarlosNihelton:
Closing this in favor of #835 as we preferred the registry instead of environment variables. Not only this design changes, but also the store purchase allowance must be refactored.

PR #817, opened 2024-06-26
fix: prevent subprocesses to sent notification to systemd

didrocks:
Unset NOTIFY_SOCKET just after forwarding readyness to systemd. This way, subprocesses will not attempt to notify systemd that they are ready.

Unfortunately, this could only be tested in some e2e tests, grepping for subprocess warnings and we are not at that level of details yet in the e2e test infra.

UDENG-3363


CarlosNihelton (approved):
LGTM

PR #819, opened 2024-06-26
fix(windows-agent): Adds `tags=wsl` by default

CarlosNihelton:


didrocks (approved):
Excellent work! I have nothing in particular to raise.

Thanks for ensuring that everything will behave as expected (tasks, config and agent.yaml!) and changing some of the tests to golden as they were becoming complex.

Issue #801, opened 2024-06-17
Replace "FQDN" by "Landscape server address"


jibel:
Replace Landscape FQDN by Landscape server address

![Capture d'écran 2024-06-14 113129](https://github.com/canonical/ubuntu-pro-for-wsl/assets/1928546/bfa9b4bf-e0bd-4fc4-8a37-99b6e7d1b2a7)

CarlosNihelton:
Closed by #820

PR #820, opened 2024-06-27
fix: Replaces Landscape "FQDN" with "server address"

CarlosNihelton:
That name is ubiquitous in Landscape docs
But look a bit cryptic to most users.

Here's how it looks after this small change:
image

---
Fixes UDENG-3259 and #801


jibel (approved):
+1

PR #821, opened 2024-06-27
fix(windows-agent): Deprecate `ubuntu_advantage` 

CarlosNihelton:
We should use `ubuntu_pro` module instead. `cloud-init` does understand `ubuntu_advantage`. Their docs state:
> Activate only on keys: ubuntu_pro, ubuntu-advantage, ubuntu_advantage

Yet, there are about 4 months since they deprecated ubuntu_advantage: https://github.com/canonical/cloud-init/pull/4865

Docs are up-to-date, so current users are likely to add `ubuntu_pro` key in their user-data. That wouldn't clash with the agent.yaml, in fact which token would then be applied can become unpredictable as both would reach cc_ubuntu_pro.py module.

So, inline with cloud-init docs, let's migrate to the `ubuntu_pro` key :)

I suspect @didrocks and @jibel faced this exact behavior (mix of both keys) during their tests.

Ah, I also sneaked a fix in the `#cloud-config` stenza. It's not been a problem in our tests because we always tested user-provided user-data as YAML (the WSL datasource merges both and rewrites the final raw userdata that is fed into the cloud-init pipeline). Should we put a shell script instead, we'd see that cloud-init wouldn't understand the part coming from `agent.yaml`.

### DIFF
--- a/windows-agent/internal/cloudinit/cloudinit.go
+++ b/windows-agent/internal/cloudinit/cloudinit.go
@@ -129,7 +129,7 @@ func marshalConfig(conf Config) ([]byte, error) {
 
 	contents := make(map[string]interface{})
 
-	if err := ubuntuAdvantageModule(conf, contents); err != nil {
+	if err := ubuntuProModule(conf, contents); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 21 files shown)


didrocks (approved):
Didn’t test in real world yet, but anyway, the change makes sense and has a good opportunity to fix it! Thanks :)

PR #822, opened 2024-07-01
docs: Reduce landscape setup in tutorial

edibotopic:
This PR makes the tutorial shorter and more generic by reducing the emphasis on Landscape setup.

The material on Landscape server setup in WSL is moved to a separate how-to.

Screenshots of UI elements were changed to ensure they are up-to-date and in English.

Minor additional changes were made to improve agreement with style guide.

UDENG-3054


docs/tutorial/index.md
```diff
@@ -23,110 +33,44 @@ Output logs are included in this tutorial when instructive but are typically omi
...
-```
-
-Done -- your self-hosted Landscape server is now up and running! 
-Now if you configure the Landscape client on your Ubuntu WSL instances to detect this server, they will also be registered with the Landscape service included in your Ubuntu Pro subscription.
-
-The server will stay running until you close the terminal, so keep the terminal open for the rest of the tutorial.
-If you do close the terminal running `ubuntu2204.exe` in a new terminal window will start the Landscape server automatically.
+For the remainder of the tutorial you will need to have a Landscape server
+set up and you should be able access your Landscape dashboard in a browser.
+Please refer to the [Landscape documentation](https://ubuntu.com/landscape/docs)
```
jibel, line 72: This link should go directly to the install and configure instructions https://ubuntu.com/landscape/docs/install-and-configure-landscape

docs/tutorial/index.md
```diff
@@ -23,110 +33,44 @@ Output logs are included in this tutorial when instructive but are typically omi
 
 WSL can be installed directly from the [Microsoft Store](https://www.microsoft.com/store/productId/9P9TQF7MRM4R).
 
-If you already have WSL installed you are advised to backup `~\.wslconfig` and remove it before continuing the tutorial:
+If you already have WSL installed and have `~\.wslconfig` on your system you
+are advised to backup the file and remove it before continuing the tutorial:
 
 ```text
+PS C:\Users\me\tutorial> Test-Path -Path "~\.wslconfig"
```
jibel, line 40: We don't tell the user what they should do depending on the result of this command.

jibel (changes requested):
Looks fine to me. Two small comments, though.

jibel (approved):
All good! Thank you!

docs/howto/set-up-landscape-server-in-wsl.md
```diff
@@ -0,0 +1,83 @@
+# How to install Landscape server in a WSL instance
+
+While a Landscape server typically runs on external computers, here it will be
+set up on a WSL instance on a Windows machine.
```
CarlosNihelton, line 4: If we'd preserve this piece I think it'd be good to present a motivation for doing so, as hosting Landscape on a WSL instance is not a supported way per se. In my view, the motivation is to ease development and testing of UP4W for those who don't have another machine to host Landscape and find setting up VMs harder than trying to run it on WSL, thus I'd place this under `dev docs`

edibotopic:
Thanks as always @jibel !!!

CarlosNihelton (commented):
Looks like I arrived late ;)

docs/howto/set-up-landscape-server-in-wsl.md
edibotopic: I think that's a great idea @CarlosNihelton and this should definitely be considered in a future revision.

For this PR I just wanted to make sure the useful information was captured somewhere so that it did not get lost. It was an "extra" after the material had been removed from the main tutorial.

I will note your suggestions down for now. Providing a more explicit motivation should be done wherever this guide is ultimately located.

Whether it should be in `dev docs` is something I/we need to think about. This depends to an extent on what we think about the state of those dev docs, their organisation as a whole and where they will fit in the WSL docs.

PR #829, opened 2024-07-04
docs: Update diagrams

edibotopic:
This PR incorporates definitions for diagrams relating to deployment and firewall requirements into the `workspace.dsl` file. The firewall diagram was already in the repo but its source was not included. I moved the original firewall diagram to the target directory for the diagram generation and it has now been replaced by an auto-generated version.

Other minor changes include:

- Deletion of duplicate diagrams
- Updating of links to autogenerated diagrams
- Additional text to motivate the how-to for setting up landscape server in WSL instance and to reference the architecture diagram included

Note that multiple key/legend files are generated in the diagrams directory using the current approach. This is the default behaviour of the Structurizr CLI but most of the legends are the same and I don't find them very useful. I think there is a way to change the gh-action to use a different renderer, which has more options (including the exclusion of the legend files). This should be the subject of future work.

UDENG-3443


docs/diagrams/structurizr-Firewall-key.png
CarlosNihelton, line 1: I think I missed some context: what are those `*-key.png` files and what are they used for?

edibotopic: Hey @CarlosNihelton -- thanks for taking a look.

These key files function as legends for the diagrams.

In C4 there are specific objects (software system, container) and relations that hold between those objects (arrows).

The key files provide a visual indicator that distinguishes, for example, between a software system and a component, which allows someone reading the diagram to discern what an arrow means or why one box is blue and the other grey. In my opinion, this is most helpful to the people actually working on the diagrams, who understand the C4 hierarchy and must view the various diagrams and their inter-relations together at once during the design phase.

I do not believe there is a strong argument for their inclusion in user-facing docs. We do not want to explain the C4 architectural diagram philosophy to a reader. In addition, the fact that each key/legend is its own image file is inefficient and messy.

It's out of scope for this PR, but there is an alternative renderer that I can test for the automated diagram generation. This (optionally) embeds a legend directly in the architecture diagram, which is far cleaner based on my initial tests. It can also be set up to exclude the legends (not possible with current renderer afaik), which could be a legitimate option.

Ultimately, to use these diagrams effectively in the docs they will need a good accompanying explanation in text form -- the legends cannot replace that.

CarlosNihelton (approved):
LGTM

PR #834, opened 2024-07-11
maint: Upgrade Go to 1.22.5

CarlosNihelton:
Plus some dependencies.

CI is red due vulnerabilities in Go stdlib net and gRPC as well. So upgrading just Go won't be enough to get a green CI.
Let's `go get -u ./... && go mod tidy` across the board :see_no_evil:

```
Vulnerability #1: GO-2024-2978
    Private tokens could appear in logs if context containing gRPC metadata is
    logged in google.golang.org/grpc
  More info: https://pkg.go.dev/vuln/GO-2024-2978
  Module: google.golang.org/grpc
    Found in: google.golang.org/grpc@v1.64.0
    Fixed in: google.golang.org/grpc@v1.64.1
    Example traces found:
Error:       #1: grpc/logstreamer/client.go:103:22: logstreamer.logClientStream.RecvMsg calls fmt.Sprintf, which eventually calls metadata.MD.String
```

```
Vulnerability #1: GO-2024-2963
    Denial of service due to improper 100-continue handling in net/http
  More info: https://pkg.go.dev/vuln/GO-2024-2963
  Standard library
    Found in: net/http@go1.22.4
    Fixed in: net/http@go1.22.5
    Example traces found:
Error:       #1: internal/ubuntupro/contractclient/client.go:106:23: contractclient.Client.GetProToken calls http.Client.Do
Error:       #2: internal/proservices/landscape/executor.go:291:23: landscape.download calls http.Get
```

PR #835, opened 2024-07-11
feat(gui): Conditionally hide the Landscape Config page based on registry

CarlosNihelton:
(Opening this PR on top of #834 because otherwise CI won't go green ;) )

This implements a "Settings" feature on top of the Windows registry via which we can enable Landscape configuration and MS Store purchase (previously configured via an environment variable), which are both currently disabled by default.
We make sure to not depend on the Windows registry for integration testing, i.e. everything is tested as if all features are enabled.

For smaller test cases we test with both features enabled and disabled to ensure correctness.

This PR keeps the dependency on the Settings on the highest level, only propagating the decision to the implementation of the related pages.

While this is a simpler design, it has a drawback of not reacting to changes in the underlying data source (the Windows registry).

Adding a listener for registry changes is more complex and not worth in my opinion.

---
UDENG-3360


CarlosNihelton:
There is a modest coverage on the lines reported as uncovered, because we don't report coverage of Flutter tests run on Windows (to avoid more complexity in CI installing tools to install tools on Windows to be able to parse and report coverage output).

didrocks (approved):
I did review the Dart/Flutter side as best of my knowledge. I agree it doesn’t worth checking and monitoring the registry for this and this is the best outcome I can think of.

Nice work!

PR #840, opened 2024-07-16
fix(agent): Don't bail out if we cannot find a WSL adapter

CarlosNihelton:
The UP4W background agent relies on the existence of the WSL net adapter and the distro instance to properly detect on which address it should serve.

Clean installs of Windows with recent versions of WSL won’t have any of those until the user creates the first WSL instance.

The `daemon` reports the error to find the address and the agent quickly exits.

We need to detect that situation and listen on localhost, so we have time to read Landscape information from registry (or talk to the GUI), so that we may end up creating WSL instances for the current user.
The current implementation won't never allow that, because the agent exits early if the WSL adapter address is not found.

There is a secondary issue to this: once an instance is created it won’t be able to talk to the agent if the networking mode is NAT (the current default), because `wsl-pro-service` expects the agent listening on the WSL adapter.  At least the provisioning tasks are delivered via cloud-init, thus independently of the communication to the agent, so that lack of communication won’t be an apparent issue and it solves itself when the agent restarts.

No user instance means no system distro and no WSL net adapter. Without this the agent will exit quickly and never have a chance to talk to Landscape to receive an install command.

---

### DIFF
--- a/windows-agent/internal/daemon/daemon.go
+++ b/windows-agent/internal/daemon/daemon.go
@@ -49,7 +49,8 @@ func (d Daemon) Serve(ctx context.Context, args ...Option) (err error) {
 
 	wslIP, err := getWslIP(ctx, args...)
 	if err != nil {
-		return fmt.Errorf("could not get the WSL adapter IP: %v", err)
+		log.Errorf(ctx, "could not get the WSL adapter IP: %v", err)
+		wslIP = net.IPv4(127, 0, 0, 1)
 	}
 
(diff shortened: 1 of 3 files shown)


CarlosNihelton:
This patch is only for unblocking tests, not a definitive solution yet, hence the reason for being kept in draft.

CarlosNihelton:
Closed by #896 .

CarlosNihelton:
Closed by #896

PR #844, opened 2024-07-19
docs: Make refs to config file extension more consistent with UI

edibotopic:
This PR makes references to the Landscape configuration file in the docs more consistent with the Ubuntu Pro For WSL UI, which includes a placeholder for `landscape.conf` during setup.

In `reference/landscape.md` it is now mentioned that this is a plain text file and that other extensions, such as `.ini` can also be used.

UDENG-3592


CarlosNihelton (approved):
LGTM!

PR #845, opened 2024-07-26
maint(gui): Upgrade flutter version

CarlosNihelton:
And major versions of some dependencies that were previously held due inter-package constraints.

For instance grpc 4.0.0 requires minimum Dart version 3.2.0, but agentapi was holding Dart max version to 3.0.0, thus dependabot couldn't upgrade grpc.

Nothing required changes in the proto-generated Dart code.


didrocks (approved):
Let’s upgrade!

PR #847, opened 2024-07-30
docs: remove excess files from diagram autogeneration and disable titles in generated diagrams

edibotopic:
This PR aimed to (1) reduce the number of excess files that are created in the repo during autogeneration of C4 diagrams and (2) leverage config options passed to the renderer to disable excess information in the diagrams, such as titles.

Originally, switching to an alternative rendering backend was trialled for (1) but this failed during test workflow runs.
The alternative solution was to simply remove unwanted files with `rm` in the workflow after generation.
The workflow already removes junk generated during each run (.puml files), so it has now been amended to also remove key/legend files, which are not useful for our purposes. This removed 11 files of ~110 kb.

The existing plantuml rendering backend has limited config options but allows titles to be disabled in diagrams. This change was made and helps reduce visual noise in what are already already complex diagrams. This also prevents naming conflicts; for example, terms like "landscape" and "container" are included in titles (according to the C4 convention) but these could cause confusion when a technology called "Landscape" is also being referenced in a diagram and "container" has a variety of uses in software.

The alt text used for one of the images in the docs was also fixed as it wasn't fit-for-purpose.

UDENG-3731


CarlosNihelton (approved):
Sensible decision in my opinion. Thanks for addressing that!

Issue #802, opened 2024-06-17
Add a "Certificate" field


jibel:
In this UI, if the only information missing to have the equivalent of the configuration file is the certificate, then there should be a file chooser field to select the certificate. The operation is much simpler for the user than typing the configuration in a text file, then selecting the text file from this UI.

![Capture d'écran 2024-06-14 113247](https://github.com/canonical/ubuntu-pro-for-wsl/assets/1928546/d0016448-8535-4ebf-8922-e3c1164ef7da)

And remove "Account Name" and "Registration Key" which are only valid on Landscape SaaS

CarlosNihelton:
> And remove "Account Name" and "Registration Key" which are only valid on Landscape SaaS

"Registration key" is valid in both self-hosted and SaaS offerings, thus we cannot remove it.

CarlosNihelton:
This bug report is about the following page (as seen in the original Figma design):
![The original Figma design](https://github.com/user-attachments/assets/282ccd8a-b7e6-472a-932f-315d5e6993a8)

As we iterate over testing the product together with Landscape we noticed the frequent need to use the the `Custom Configuration` option (which requires writing a conf file, certainly not as easy as filling a three field form) just because our setup required informing the self-hosted Landscape SSL certiticate (aka public key). We believe that customers with self-hosted Landscape servers are likely to pass custom certificates to their internal clients, so offering an input field for that value could result in improved UX.

But that page seems already too busy with always visible input fields grouped into two subforms. Adding a 4th input field to the top subform without further design thinking seems not the best option. Additionally there are some interleaving logic around the four fields falling into two possible scenarios:

1. If the server is `landscape.canonical.com` (aka Landscape SaaS), then `Landscape server address`, `Landscape Account name` and `Registration key` must be interactive, while the `SSL public key` should be disabled.   `Registration key` is optional but `Landscape Account name` is required and must not be `standalone`.

2. otherwise `Landscape server address`, `Registration key`  and `SSL public key` must be interactive. `Landscape Account name` could be shown containing the `standalone` value but disabled so users cannot change it. Alternatively that field could be entirely removed, as the business logic would attribute the value `standalone` to it. `Registration key`  and `SSL public key` are optional in this scenario.

Below I present two alternatives ideas that describe well the interleaved logic. Can we use them as input for the Design team to come up with an actual design proposal?

1. Mutual exclusive fields being replaced with some animation hinting the change in the form

[...]

PR #849, opened 2024-07-30
feat(gui): Landscape redesign to fit an input field for the SSL key

CarlosNihelton:
Adding one single field motivated a whole UI redesign. The discussion and thought process is documented in #802 .

The key pilar to understand this PR is: we no longer show relationship between the form fields being displayed, but rather separate them into subforms so the fields that are visible are the ones relevant to the type of configuration the user is currently doing.

[This comment in #802](https://github.com/canonical/ubuntu-pro-for-wsl/issues/802#issuecomment-2237200764) summarizes what I mean by inter-field relationship, non-trivial logic that would disable or enable or change the forbidden values of a form field depending on the values of neighbour fields. The previous logic was at the limit of being fine because the form was small, adding just one extra field was enough to turn it complex enough to justify a redesign.

We now present those types of configuration as three distinct subforms: the Landscape SaaS, self-hosted Landscape and an "advanced configuration" through which the user can supply a custom Landscape client configuration file. That way input validation is much clearer now and there is no interleaved logic causing form fields to be visible in a disabled state. The end result is a clearer UX.

The backing validation logic is split in subclasses mirroring the UI split in subforms.

The design is more horizontal now, the left side of the UI is havier on static text while the right side is more interactive.

Because of the way the widgets were composed to achieve this layout, an extra care was required to achieve a coherent tab navigation.

Before (good enough logic, tab navigation just worked):

https://github.com/user-attachments/assets/6886e8b6-5f7d-4e3e-826d-45e86f897a66

After (smaller and more focused subforms, no interleaving logic, tab navigation works with a little more effort):

https://github.com/user-attachments/assets/e001c436-fa70-49ae-a09b-7e463afb919d

### DIFF
--- a/gui/packages/ubuntupro/integration_test/ubuntu_pro_for_wsl_test.dart
+++ b/gui/packages/ubuntupro/integration_test/ubuntu_pro_for_wsl_test.dart
@@ -12,12 +12,12 @@ import 'package:ubuntupro/constants.dart';
 import 'package:ubuntupro/core/environment.dart';
 import 'package:ubuntupro/launch_agent.dart';
 import 'package:ubuntupro/main.dart' as app;
-import 'package:ubuntupro/pages/landscape/landscape_model.dart';
 import 'package:ubuntupro/pages/landscape/landscape_page.dart';
 import 'package:ubuntupro/pages/startup/startup_page.dart';
 import 'package:ubuntupro/pages/subscribe_now/subscribe_now_page.dart';
 import 'package:ubuntupro/pages/subscribe_now/subscribe_now_widgets.dart';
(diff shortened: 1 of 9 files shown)


didrocks (approved):
I didn’t spot anything particular and the design matches what we discussed and is the most logical IMHO.

However, should we grey out the first option for now, as Landscape SASS doesn’t support the beta yet?

CarlosNihelton:
> However, should we grey out the first option for now, as Landscape SASS doesn’t support the beta yet?

Very good point. The last commit does it, while keeping it enabled when compiled in debug mode so tests can run with that option enabled with a TODO item to remove the condition once SaaS starts supporting WSL.

didrocks (approved):
Excellent!

PR #854, opened 2024-08-01
fix(gui): SSL public key hint must have a certificate file extension

CarlosNihelton:
Thanks @edibotopic for noticing that :)

Additional to that fix. the first commit of this PR moves a user-facing string (SSL Key input label) from the source code to the translatable resource, so it can be properly localized.


edibotopic (approved):
LGTM!

PR #855, opened 2024-08-01
fix(gui): Landscape config serialization

CarlosNihelton:
The Landscape ping service doesn't support https by default.

Additionally, wslpath can fool wsl-pro-service, when the path passed to the command line is an empty path it outputs the current working directory absolute path. Due systemd, cwd is `/`.

image

We should have wsl-pro-service checking for empty paths before invoking wslpath for added correctness (users could well assemble a config file by hand having `ssl_public_key = `), but it also makes sense to fix it here, as we're aware of which fields are optional in the UI. But, instead of treating this (or any path) field as special, I'm being more generic, **filtering out any line in the INI configuration output that could have empty values**, i.e., we won't output lines with this shape : `key =`.

More test assertions were added validating those assumptions.


didrocks (approved):
The goal is clear with this rationale, looks good to me :)

PR #858, opened 2024-08-06
fix(wsl-pro-service): Do not translate empty paths

CarlosNihelton:
I wanted this PR to be laser-focus, just a bugfix, thus I'm not refactoring the executable-mocking just yet.

The issue is: when we invoke `wslpath -au ''`  we get the absolute path of the current working directory.
That's surely not what we expect of a Windows path.

It could be problematic if the user passes a custom Landscape config file containing this line `ssl_public_key= `, as we'd modify the final client.conf into `ssl_public_key = /` (systemd makes it so our current working directory is `/`). Landscape client would tolerate the empty value in the INI sintax, but this would force it to try to read `/` as a certificate file. Clearly not what we want.

A similar thing could happen in `system.UserProfileDir()`, but that's even funnier. Having %USERPROFILE% set to an empty string is indeed an odd edge case, so I'm not 100% sure the added code is worth it, as we're likely to err out attempting to open inexistant directories. I think it worths it because we are sure on how we err out. Leaving as it was could lead us into reading files we were not supposed to.

They issues were:

`echo %USERPROFILE%` would output `ECHO is on` if that environment variable is empty. We fix it by running `echo.%USERPROFILE%` instead. It outputs an empty line for that case. Finally, if the output is empty, we bail out.

image

I made some modifications in the wslpath and cmd.exe mocks to get closer to the surprising behavior, so I could add test cases 
demonstrating the surprising behavior.

---
Closes UDENG-3788

### DIFF
--- a/wsl-pro-service/internal/testutils/mock_executables.go
+++ b/wsl-pro-service/internal/testutils/mock_executables.go
@@ -86,8 +86,9 @@ const (
 	LandscapeEnableErr  = "UP4W_LANDSCAPE_ENABLE_ERR"
 	LandscapeDisableErr = "UP4W_LANDSCAPE_DISABLE_ERR"
 
-	WslpathErr       = "UP4W_WSLPATH_ERR"
-	WslpathBadOutput = "UP4W_WSLPATH_BAD_OUTPUT"
+	WslpathErr             = "UP4W_WSLPATH_ERR"
+	WslpathBadOutput       = "UP4W_WSLPATH_BAD_OUTPUT"
+	EmptyUserprofileEnvVar = "UP4W_EMPTY_USERPROFILE_ENV_VAR"
 
 	CmdExeErr = "UP4W_CMDEXE_ERR"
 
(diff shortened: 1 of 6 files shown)


CarlosNihelton:
Also tested by hand on Ubuntu-24.04, no hickups when attempting to find the .address file.

wsl-pro-service/internal/testutils/mock_executables.go
```diff
@@ -511,12 +512,18 @@ func WslPathMock(t *testing.T) {
...
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Could not get current working directory: %v", err)
+			}
 
 			stdout, ok := map[string]string{
 				windowsUserProfileDir:                   linuxUserProfileDir,
 				`D:\Users\TestUser\certificate`:         filepath.Join(defaultWindowsMount, "Users/TestUser/certificate"),
 				`D:/Users/TestUser/certificate`:         filepath.Join(defaultWindowsMount, "Users/TestUser/certificate"),
 				`/idempotent/path/to/linux/certificate`: `/idempotent/path/to/linux/certificate`,
+				``:                                      cwd,
```
didrocks, line 526: The last 3 could have use "" actually (we used string litterals on the first one because of the \, but here we have a mix of "" on some side and not on other".

Anyway, just a nitpick, no need to change it :)

didrocks (approved):
Excellent! Thanks a lot for the detailed explanation and the research around it! Looks good to me and tests pass :)

wsl-pro-service/internal/testutils/mock_executables.go
CarlosNihelton: Indeed, let's play on the consistency side.

PR #875, opened 2024-08-19
fix(gui): Remove spaces from the Pro token input field

### LINKED ISSUE #857: Strip spaces from Pro token in Pro UI
Ubuntu Pro for WSL v0.1.4-474-g1185b81a

If I copy/paste the token from the Pro dashboard to the token field of the app, it contains a space at the end. The field is then, rightfully, considering the token too long and invalid.

It'd be convenient to strip spaces when we test the length of the string and use the stripped string as Pro token.

CarlosNihelton:
Sometimes copy-pasting a pro token comes with trailing spaces. Neither leading, trailing or even middle spaces are alloweable in a token. This PR leverages Flutter builtin mechanisms to filter input supplied to the token text field such that any kind of whitespaces (not only the ASCII ones) are discarded before any further processing.

Closes #857 

---

I'm purposefully basing this PR on top of #873 so I can stress CI to exercise the changes therein introduced. There is no dependency between the two tasks.

---
UDENG-3810

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
+++ b/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
@@ -1,5 +1,6 @@
 import 'package:dart_either/dart_either.dart';
 import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:yaru/yaru.dart';
 import '../../core/either_value_notifier.dart';
@@ -78,6 +79,10 @@ class _ProTokenInputFieldState extends State<ProTokenInputField> {
           children: [
             Expanded(
               child: TextField(
+                inputFormatters: [
+                  // This ignores all sorts of (Unicode) whitespaces (not only at the ends).
+                  FilteringTextInputFormatter.deny(RegExp(r'\s')),
+                ],
                 autofocus: false,
                 controller: _controller,
                 decoration: InputDecoration(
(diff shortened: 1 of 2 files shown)


gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
```diff
@@ -78,6 +79,10 @@ class _ProTokenInputFieldState extends State<ProTokenInputField> {
           children: [
             Expanded(
               child: TextField(
+                inputFormatters: [
+                  // This ignores all sorts of (Unicode) whitespaces (not only at the ends).
+                  FilteringTextInputFormatter.deny(RegExp(r'\s')),
```
jibel, line 84: I'm no dart coder, so forgive my ignorance. Why not use an allow() constructor instead such as `.allow(RegExp(r'[a-zA-Z0-9]'))`  and also limit the length of the field to the max length of a token?
(also note that I couldn't confirm that it is the correct pattern for a token but it must be available somewhere in the server code)

CarlosNihelton: Well, the proper filtering to achieve your proposal is a bit more evolving, but doable to a certain extent (there are preconditions about the certain element positions as well as the use of base58 encoding that won't fit well as input formatters). 

The downside I see is that by implementing a good input filtering scheme we might be helping non-behaved users to guess a valid token by inputting a whole lot of non-sense characters that we just strip out leaving only the valid candidates, reducing the guessing effort.

There is already a more complete validation mechanism for the structure of the pro token defined in `gui/lib/core/pro_token.dart`, which is what backs the error messages we see when it's too long, too short, "corrupted" (invalid structure) etc.

If the guessing game I mentioned above is not an issue, then we can leverage the `inputFormatters` to remove parts of the input value that would cause some error conditions to surface in the UI. We can use them to constrain the length as you suggested as well as filtering characters that we can be sure are invalid (such as `I`, `O`, `0`, etc).

jibel (approved):
All good! Thanks!

CarlosNihelton:
Just rebased on main to avoid merging the commits relategd to #873 together with this one.

PR #881, opened 2024-08-23
fix(windows-agent): No empty agent.yaml

CarlosNihelton:
When the config is empty, the yaml serialization outputs `{}`. We still added some commentary in the agent.yaml, resulting in a bunch of nothing that confused cloud-init, because it's enablement is based on the existence of the files, their contents are only checked later. That wouldn't be a problem per se, as the WSL datasource would recognize there's nothing to do and bail out, but that would have costed unnecesary CPU cicles, and worse than that, the current implementation of cloud-init makes it log as an error the fact that there is nothing to do at this point, which results in more noise if the user see the output of  `cloud-init status --long`.

The fix is rather simple: do not write an empty object into agent.yaml. If the config is empty we delete that file. This way, if there is nothing from the user nor the agent's perspective, cloud-init will remain trully disabled.

I updated some tests that relied on the fact that cloud-init would create the agent.yaml file unconditionally, which is no longer the case.

UDENG-4169


windows-agent/internal/cloudinit/cloudinit_test.go
```diff
@@ -172,6 +191,17 @@ hostagent_uid = landscapeUID1234
 				return
 			}
 
+			if tc.breakRemovingFile {
```
didrocks, line 194: ditto here and below on what we want to be separated from the initial state. That way, we have a clear story on the side effect and what is expected.

windows-agent/internal/proservices/proservices_test.go
```diff
@@ -106,6 +110,17 @@ func TestNew(t *testing.T) {
 	}
 }
 
+func checkFileExists(path string) func() bool {
+	return func() bool {
+		return fileExists(path)
+	}
+}
+
+func fileExists(path string) bool {
```
didrocks, line 119: This function is only called once, so you can inline it in checkFileExists if it’s not reused :)

windows-agent/internal/proservices/testdata/TestNew/golden/when_the_config_cannot_check_if_it_is_read-only
```diff
@@ -1,3 +1,8 @@
 #cloud-config
 # This file was generated automatically and must not be edited
-{}
+landscape:
```
didrocks, line 3: that was the initial state, is it expected?

windows-agent/internal/proservices/testdata/TestNew/golden/when_the_subscription_stays_empty
```diff
@@ -1,3 +1,8 @@
 #cloud-config
 # This file was generated automatically and must not be edited
-{}
+landscape:
```
didrocks, line 3: ditto

didrocks (changes requested):
Some small fixes to do and some questions!

windows-agent/internal/proservices/testdata/TestNew/golden/when_the_config_cannot_check_if_it_is_read-only
CarlosNihelton: The previous implementation couldn't assert on the updated agent.yaml, because it's intent was just to ensure the delivery of notifications happened, it didn't wait for cloudinit to actually write the agent.yaml file, as it happens asynchronously from the point of view of the test main goroutine, thus it only asserted the file would exist.
With the current implementation I give it some time to ensure cloudinit component would have written up-to-date information inside agent.yaml, thus the golden needs to reflect the desired final state.

windows-agent/internal/cloudinit/cloudinit_test.go
```diff
@@ -172,6 +191,17 @@ hostagent_uid = landscapeUID1234
...
+				require.DirExists(t, path, "There should be a directory instead of agent.yaml")
+				return
+			}
+
+			if tc.skipProToken && tc.skipLandscapeConf {
+				// empty file should not exist
+				require.NoFileExists(t, path, "There should not be cloud-init agent file without useful contents")
```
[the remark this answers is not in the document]
CarlosNihelton: Since the `LoadWith*` family of functions all `require.NoError` when opening the file, I guess the way to do this is by first calling `Path` and checking if the returned path exists.

windows-agent/internal/proservices/testdata/TestNew/golden/when_the_subscription_stays_empty
CarlosNihelton: ditto :)

windows-agent/internal/proservices/testdata/TestNew/golden/when_the_config_cannot_check_if_it_is_read-only
didrocks: Perefect! :)

windows-agent/internal/cloudinit/cloudinit_test.go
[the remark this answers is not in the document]
didrocks: makes sense to me!

didrocks (approved):
All good from my standpoint now!

PR #882, opened 2024-08-26
fix: golangci-lint warnings about unsafe integer conversions

CarlosNihelton:
Most of the warnings were not a real issue, but the one around the Win32 GetAdapterAddersses was harder to prove with the previous implementation so I refactored it a little bit to make it easy to reach the piece of mind that we're not doing nasting error-prone narrowing conversions.

This should unblock #877 .

UDENG-4234

### DIFF
--- a/storeapi/go-wrapper/microsoftstore/store.go
+++ b/storeapi/go-wrapper/microsoftstore/store.go
@@ -2,8 +2,10 @@ package microsoftstore
 
 import (
 	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 )
 
 // findWorkspaceRoot climbs up the current working directory until the Go workspace root is found.
@@ -25,4 +27,30 @@ func findWorkspaceRoot() (string, error) {
 			return parent, nil
 		}
 	}
+}
+
+// checkError inspects the values of hres and err to determine what kind of error we have, if any, according to the rules of syscall/dll_windows.go.
+func checkError(hres int64, err error) (int64, error) {
+	// From syscall/dll_windows.go (*Proc).Call doc:
+	// > Callers must inspect the primary return value to decide whether an
+	//   error occurred [...] before consulting the error.
+	if e := NewStoreAPIError(hres); e != nil {
+		return hres, fmt.Errorf("storeApi returned error code %d: %w", hres, e)
+	}
+
+	if err == nil {
+		return hres, nil
+	}
+
+	var target syscall.Errno
+	if b := errors.As(err, &target); !b {
+		// Supposedly unrechable: proc.Call must always return a syscall.Errno
+		return hres, err
+	}
+
+	if target != syscall.Errno(0) {
+		return hres, fmt.Errorf("failed syscall to storeApi: %v (syscall errno %d)", target, err)
+	}
+
+	return hres, nil
 }
--- a/storeapi/go-wrapper/microsoftstore/store.go
+++ b/storeapi/go-wrapper/microsoftstore/store.go
@@ -44,7 +44,7 @@ func checkError(hres int64, err error) (int64, error) {
 
 	var target syscall.Errno
 	if b := errors.As(err, &target); !b {
-		// Supposedly unrechable: proc.Call must always return a syscall.Errno
+		// Supposedly unreachable: proc.Call must always return a syscall.Errno
 		return hres, err
 	}
 
--- a/storeapi/go-wrapper/microsoftstore/store.go
+++ b/storeapi/go-wrapper/microsoftstore/store.go
@@ -34,20 +34,29 @@ func checkError(hres int64, err error) (int64, error) {
 	// From syscall/dll_windows.go (*Proc).Call doc:
 	// > Callers must inspect the primary return value to decide whether an
 	//   error occurred [...] before consulting the error.
+	// There is no possibility of nil  error, the `err` return value is always constructed with the
+	// result of `GetLastError()` which could have been set by something completely
+	// unrelated to our code some time in the past, as well as it could be `ERROR_SUCCESS` which is the `Errno(0)`.
+	// If the act of calling the API fails (not the function we're calling, but the attempt to call it), then we'd
+	// have a meaningful `syscall.Errno` object via the `err` parameter, related to the actual failure (like a function not found in this DLL)
+	// Since our implementation of the store API doesn't touch errno the call should return `hres`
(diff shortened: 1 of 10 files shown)


storeapi/go-wrapper/microsoftstore/store.go
```diff
@@ -26,3 +28,29 @@ func findWorkspaceRoot() (string, error) {
...
+		return hres, fmt.Errorf("storeApi returned error code %d: %w", hres, e)
+	}
+
+	if err == nil {
+		return hres, nil
+	}
+
+	var target syscall.Errno
+	if b := errors.As(err, &target); !b {
+		// Supposedly unrechable: proc.Call must always return a syscall.Errno
```
didrocks, line 47:
```suggestion
		// Supposedly unreachable: proc.Call must always return a syscall.Errno
```

storeapi/go-wrapper/microsoftstore/store.go
```diff
@@ -26,3 +28,29 @@ func findWorkspaceRoot() (string, error) {
...
+		return hres, err
+	}
+
+	if target != syscall.Errno(0) {
+		return hres, fmt.Errorf("failed syscall to storeApi: %v (syscall errno %d)", target, err)
+	}
+
+	return hres, nil
```
didrocks, line 55: This looks weird to me, so we have a non nil error, which was not an errno, but is Errno(0), so it’s a successfull error? :) It seems that the API can’t send that (at least, not from the tests), but maybe give it a good comment?

storeapi/go-wrapper/microsoftstore/store_test.go
```diff
@@ -87,6 +89,36 @@ func TestGetSubscriptionExpirationDate(t *testing.T) {
...
+	}{
+		"Success": {},
+		// If HRESULT is not in the Store API error range and err is not a syscall.Errno then we don't have an error.
+		"With an unknown value (not an error)": {hresult: 1, wantError: false},
+
+		"Upper bound of the Store API enum range": {hresult: -1, wantError: true},
+		"Lower bound of the Store API enum range": {hresult: int64(microsoftstore.ErrNotSubscribed), wantError: true},
+		"With a system error (errno)":             {hresult: 32 /*garbage*/, err: syscall.Errno(2) /*E_FILE_NOT_FOUND*/, wantError: true},
+		"With a generic (unreachable) error":      {hresult: 1, err: errors.New("test error"), wantError: true},
+	}
```
didrocks, line 108: actually, you can test here the "weird successfull error" too and mention this should not happen :)

storeapi/go-wrapper/microsoftstore/store_test.go
```diff
@@ -87,6 +89,36 @@ func TestGetSubscriptionExpirationDate(t *testing.T) {
 	require.ErrorIs(t, gotErr, wantErr, "GetSubscriptionExpirationDate should have returned code %d", wantErr)
 }
 
+func TestErrorVerification(t *testing.T) {
+	t.Parallel()
+	testcases := map[string]struct {
+		hresult int64
+		err     error
+
+		wantError bool
```
didrocks, line 98: nitpick: we usually use `wantErr`, let’s be consitent :)

didrocks (changes requested):
Looks great! I’m trusting you on the Win32 API returned values.

However, I have a couple of nitpicks and one question/remark on the growing size for the adapter address logic!

storeapi/go-wrapper/microsoftstore/store.go
CarlosNihelton: According to the syscall documentation, there is no possibility of nil error, the `err` return value is always constructed with the result of `GetLastError()` which could have been set by something completely unrelated to our code sometime in the past, as well as it could be `ERROR_SUCCESS` which is the `Errno(0)`. :)

If the act of calling the API fails (not the function we're calling, but the attempt to call it), then we'd get a meaningful `syscall.Errno` object via the `err` parameter, certainly no `Errno(0)`, related to the actual failure (like a function not found in this DLL or whatever).

Since our implementation of the store API doesn't touch errno the call should return `hres` between -128 and 0 plus garbage in the `err` argument, we only care about the `hres` in this case (the begining of the function).

`err` should never be nil nor something other than `syscall.Errno`, so I'm not sure about the actual reason to have added this `error.As` check.

windows-agent/internal/daemon/networking.go
```diff
@@ -140,14 +141,15 @@ func getAddrList(opts options) (head *ipAdapterAddresses, err error) {
...
-		size := buff.byteCount()
-		err := opts.getAdaptersAddresses(family, flags, 0, &buff.data[0], &size)
+		size, err = buff.resizeBytes(size)
+		if err != nil {
+			return nil, err
+		}
+		err = opts.getAdaptersAddresses(family, flags, 0, &buff.data[0], &size)
 		if errors.Is(err, ERROR_BUFFER_OVERFLOW) {
 			// Buffer too small, try again with the returned size.
-			buff.resizeBytes(size)
```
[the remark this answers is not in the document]
CarlosNihelton: I made resizing the buffer to fail if the resulting allocation would exceed MaxUint32, so we enforce the integer arithmetic safety. At the first glance it seems impossible, but if we pass a value close to MaxUint32 and that's not a multiple of the size of the type T, than we increment the amount of T's we'd allocate, which may lead to more than MaxUint32 bytes (that's why I'm doing my math with uint64 in the first place). We still need the size to be uint32 because that's what `GetAdapterAddresses` works with.

I'm placing the resize call back to where they were. Having it inside the loop would allow checking for its error only once.

The loop only exists for one reason: cope with the inherently racy Win32 GetAdapterAddresses syscall. It's inherently racy because:
- the caller must provide enough memory for it to write the IP adapter addresses table (15 kB in our case)
- the memory supplied by the caller might be insuficient;
- it advises the amount of memory needed when that happens;
- the caller then needs to allocate more memory;
- the table might have changed enough so the new buffer is not enough either.

It seems rather impractical condition, but that's the recommended way to call GetAdapterAddresses ( [see the MS docs](https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses#:~:text=The%20recommended%20method,function%20multiple%20times.) )
That pattern of double-syscall is quite common on the Windows world: call me once so I tell you my preconditions so you can call me the second time the right way :) They even have an open source C++ library that handles this kind of call generically and automatically (among other things :) )

image

didrocks (approved):
Thanks for addressing those!

PR #883, opened 2024-08-27
fix(test): No `os.Exit()` calls in `TestMain` and no SA3000 warning

CarlosNihelton:
When upgrading golangci-lint in #870 I introduced a bunch of os.Exit calls in test functions blindily following the SA300 warning. That's bad, because it prevents the Go runtime to perform some clean-up after `m.Run()` returns.
In fact, [Go 1.15 changelog](https://go.dev/doc/go1.15#testingpkgtesting) clearly states that m.Run() does the right thing in regards to the program's exit code. `staticcheck` running alone won't report that warning. `golangci-lint` is failing to propagate the go version, thus confusing the downstream linter.

### DIFF
--- a/common/grpc/logconnections/logconnections_test.go
+++ b/common/grpc/logconnections/logconnections_test.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"flag"
-	"os"
 	"testing"
 
 	"github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logconnections"
@@ -109,5 +108,5 @@ func TestMain(m *testing.M) {
 		logrus.StandardLogger().SetLevel(logrus.DebugLevel)
 	}
 
-	os.Exit(m.Run())
+	m.Run()
 }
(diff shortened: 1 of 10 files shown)


.golangci.yaml
```diff
@@ -71,3 +71,7 @@ linters-settings:
   # Never have naked return ever
   nakedret:
     max-func-lines: 1
+  staticcheck:
+    checks:
+      - all
+      - '-SA3000' # disable false positive related to m.Run() and os.Exit. That's not needed since Go 1.15.
```
didrocks, line 77: this shouldn’t be needed with latest golangci-lint, no?

didrocks (changes requested):
Should be tested with latest golangci-lint

.golangci.yaml
CarlosNihelton: Yeah, we upgraded to 1.60.3 yesterday, this part should no longer be needed. I just rebased over the latest main and dropped that change.

didrocks (commented):
All the `*/Go quality check` now passes with the new golangci-lint :)

didrocks (approved):

PR #886, opened 2024-08-28
fix(agent): No empty cloudinit

CarlosNihelton:
When the config is empty, the yaml serialization outputs {}. We still added some commentary in the agent.yaml, resulting in a bunch of nothing that confused cloud-init, because it's enablement is based on the existence of the files, their contents are only checked later. That wouldn't be a problem per se, as the WSL datasource would recognize there's nothing to do and bail out, but that would have costed unnecesary CPU cicles, and worse than that, the current implementation of cloud-init makes it log as an error the fact that there is nothing to do at this point, which results in more noise if the user see the output of cloud-init status --long.

The fix is rather simple: do not write an empty object into agent.yaml. If the config is empty we delete that file. This way, if there is nothing from the user nor the agent's perspective, cloud-init will remain trully disabled.

I updated some tests that relied on the fact that cloud-init would create the agent.yaml file unconditionally, which is no longer the case.

UDENG-4169

### DIFF
--- a/windows-agent/internal/cloudinit/cloudinit.go
+++ b/windows-agent/internal/cloudinit/cloudinit.go
@@ -92,10 +92,10 @@ func (c CloudInit) WriteDistroData(distroName string, cloudInit string) error {
 // removeFileInDir attempts to remove the file 'dir/file' if it exists. Missing file is not an error.
 func removeFileInDir(dir, file string) error {
 	err := os.Remove(filepath.Join(dir, file))
-	if err == nil || os.IsNotExist(err) {
-		return nil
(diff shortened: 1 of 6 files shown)


CarlosNihelton:
This is the same as #881 , but to merge on main. I merged that PR on the wrong base branch by mistake 😅

didrocks (approved):
Just looked quickly at the diff, and it seems that you did carry what was needed.

CarlosNihelton:
Indeed, I just restored the branch `no-empty-cloudinit` which had been rebased on main and opened a new PR with that.

PR #887, opened 2024-08-28
fix(gui): Enable landscape by default

CarlosNihelton:
In #815 we disabled Landscape UI elements by default. 

The new default is Landscape UI elements enabled. They go disabled if the value `LandscapeConfigVisibility` under `HKEY_CURRENT_USER\Software\Canonical\UbuntuPro\` is set to `1`. If missing, unset or set to anything else, we go with the new default. Since that PR only affected the GUI, agent remains untouched.

The value name might be a bit more mouthful but it has the property of not conveying the default behaviour, allowing us to change it without surprising new users.

Tests and dev docs are updated as well.

---

UDENG-4243


gui/packages/ubuntupro/lib/core/settings.dart
```diff
@@ -12,9 +12,9 @@ class Settings {
     final purchase = repository.readInt(kAllowStorePurchase) == 1
         ? Options.withStorePurchase
         : Options.none;
-    final landscape = repository.readInt(kShowLandscapeConfig) == 1
```
jibel, line 15: You could have kept Show and check if it's set to 0. If we decide to change the default again (which is totally possible), it woud avoid to change the name of the key again. What do you think?

jibel (commented):
one question to address that may or may not require changes depending on your answer.

gui/packages/ubuntupro/lib/core/settings.dart
CarlosNihelton: Besides I have the powers of Git, so if there comes a time we need to go back I can simply `git revert ...`.

CarlosNihelton: Found a longer, but better, name for the value. I just rebased over latest main. There is only one new commit.

jibel (approved):
Thanks for this update!

PR #901, opened 2024-09-06
fix(windows-agent): Can't find config file inside the public dir

CarlosNihelton:
`%USERPROFILE%\.ubuntupro`.

The fix I'm proposing here is the simplest but maybe less correct if we consider that the public dir could point to anywhere else other than the default (`%USERPROFILE%\.ubuntupro` when in production). 

The issue is: when configuring viper we added `$HOME/.ubuntupro` to the config paths. Since `$HOME` is not defined on Windows nor `/` is the default path separator, that addition never worked. By ensuring we use the correct path separator, we allow viper to consider `%HOMEDRIVE%` , `%HOMEPATH%` and `%USERPROFILE%`  for `$HOME` (viper's implementation details).

The test case `"When the config is in user's public directory"` would only fail on Windows.

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
@@ -2,6 +2,7 @@ package agent_test
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


didrocks (approved):
and here we go, it’s always when I accept to not have parametric testing on one test that we then have at some point to add it!

Thanks for fixing those, let’s get verbose logs!

PR #902, opened 2024-09-06
maint(windows-agent): Investigate skipping distros

CarlosNihelton:
```
time=2024-07-24T12:17:10-07:00 level=error msg=Landscape:  skipping distro "Ubuntu-24.04" from landscape info: could not get distro "Ubuntu-24.04"'s state: could not get states of distros: exit status 1. Stdout: . Stderr:
```

That happens during construction of the status message sent to Landscape. If we skip a distro, Landscape understands it doesn't exist anymore and removes it from the dashboard. But that happens sometimes when an existing distro. It seems a WSL internal error.

https://github.com/ubuntu/GoWSL/pull/120 should prevent the stdout and stderr messages to come empty as shown above.
Thus I'm updating GoWSL here.

That doesn't prevent the issue itself.

I couldn’t reproduce this outside of UP4W. I attempted building a small binary in Go doing the same steps as GoWSL does and put it to run during system startup in loop, rebooted the OS many times but no success. The few times I’ve seen this happening in the logs were early during app startup, so I suspect it may happen because some WSL internal component is not yet ready. Thus this PR: adding one-time retry if an error happens when attempting to retrieve a distro state.

Either that will prevent the issue once for all or at least give us more context about what causes that error.

---

UDENG-3722

### DIFF
--- a/windows-agent/internal/proservices/landscape/utils.go
+++ b/windows-agent/internal/proservices/landscape/utils.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	landscapeapi "github.com/canonical/landscape-hostagent-api"
 	log "github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logstreamer"
@@ -202,7 +203,12 @@ func (e newInstanceInfoMinorError) Error() string {
 func newInstanceInfo(d *distro.Distro) (info *landscapeapi.HostAgentInfo_InstanceInfo, err error) {
 	state, err := d.State()
 	if err != nil {
-		return info, err
+		// Try again
+		<-time.After(1 * time.Second)
+		state, err = d.State()
+		if err != nil {
+			return info, fmt.Errorf("WSL internal error after retry: %v", err)
+		}
 	}
 
 	var instanceState landscapeapi.InstanceState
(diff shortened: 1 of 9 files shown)


CarlosNihelton:
Now we have to upgrade the Go. I'm unsure if we should Go with the toolchain to 1.22.7 or migrate straight into 1.23.1. 1.23.0 is affected by the same vulnerability reported in ´go/parser`.

jibel:
> Now we have to upgrade the Go. I'm unsure if we should Go with the toolchain to 1.22.7 or migrate straight into 1.23.1. 1.23.0 is affected by the same vulnerability reported in ´go/parser`.

The version of Go has been bumped to 1.23, so you'd *just* need to rebase this PR

windows-agent/internal/proservices/landscape/utils.go
```diff
@@ -202,7 +203,12 @@ func (e newInstanceInfoMinorError) Error() string {
 func newInstanceInfo(d *distro.Distro) (info *landscapeapi.HostAgentInfo_InstanceInfo, err error) {
 	state, err := d.State()
```
didrocks, line 204: I would rather have this in a loop, with some timeouts, as we are uncertain and this could happen on startup, as you said, we are unsure when WSL would be "ready".

didrocks (changes requested):
Tell me what you think about my comment or if you think this is unecessary.

As a aside, you will to rebase again due to conflicts :)

windows-agent/internal/proservices/landscape/utils.go
CarlosNihelton: Yeap! That makes sense. I saw this happening recently while the everything was up and running for a while.

CarlosNihelton:
Just rebased over the latest main.

didrocks (approved):
Looking all good from my side!

PR #928, opened 2024-10-03
docs: How-to create WSL instances on multiple Windows machines via Landscape API

CarlosNihelton:
This PR makes heavy use of tabbed code blocks to show alternative ways to do things with Bash and PowerShell.

image

Both alternatives were tested. The bash one comes from the [Landscape documentation](https://ubuntu.com/landscape/docs/use-a-specific-ubuntu-image-source-for-wsl#installing-with-a-custom-image), I essentially extracted the piece I was more concerned about (custom rootfs with cloud-init data) and made a self-contained guide out of it (plus translating to PowerShell).

---

UDENG-4694

### DIFF
--- a/docs/howto/custom-rootfs-multiple-targets.md
+++ b/docs/howto/custom-rootfs-multiple-targets.md
@@ -27,16 +27,16 @@ export LANDSCAPE_USER_EMAIL=admin@mib.com
 export LANDSCAPE_USER_PASSWORD=mib
 export LANDSCAPE_URL=https://landscape.mib.com
 
-# The URL of the custom rootfs to be deployed:
+# The URL of the custom rootfs to be deployed
 export ROOTFS_URL="http://landscape.mib.com:9009/ubuntu-24.04-custom.tar.gz"
 
-# The list of IDs of the different Windows machines on which we are going to deploy WSL instances:
+# The list of IDs of the different Windows machines on which we are going to deploy WSL instances
 export PARENT_COMPUTER_IDS=(26 30 31)
 
-# The name of the WSL instance to be created:
+# The name of the WSL instance to be created
 export COMPUTER_NAME=Carbonizer
 
-# Path to the cloud-config file whose contents will be used to initialize the WSL instances:
+# Path to the cloud-config file whose contents will be used to initialize the WSL instances
 export CLOUD_INIT_FILE="~/Downloads/init.yaml"
 ```
 
@@ -49,16 +49,16 @@ $LANDSCAPE_USER_EMAIL="admin@mib.com"
 $LANDSCAPE_USER_PASSWORD="mib"
 $LANDSCAPE_URL="https://landscape.mib.com"
 
-# The URL of the custom rootfs to be deployed:
+# The URL of the custom rootfs to be deployed
 $ROOTFS_URL="http://landscape.mib.com:9009/ubuntu-24.04-custom.tar.gz"
 
-# The list of IDs of the different Windows machines on which we are going to deploy WSL instances:
+# The list of IDs of the different Windows machines on which we are going to deploy WSL instances
 $PARENT_COMPUTER_IDS=@(26, 30, 31)
 
-# The name of the WSL instance to be created:
+# The name of the WSL instance to be created
 $COMPUTER_NAME="Carbonizer"
 
-# Path to the cloud-config file whose contents will be used to initialize the WSL instances:
+# Path to the cloud-config file whose contents will be used to initialize the WSL instances
 $CLOUD_INIT_FILE="~\Downloads\init.yaml"
 ```
 
@@ -248,7 +248,7 @@ WSL_JSON=$( jq -n                           \
     --arg b64 "$BASE64_ENCODED_CLOUD_INIT"  \
     '{rootfs_url: $rf, computer_name: $cn, cloud_init: $b64}' )
 
-# Issue the command for each Windows machine:
+# Issue the command for each Windows machine
 for COMPUTER_ID in "${PARENT_COMPUTER_IDS[@]}"; do
     API_RESPONSE=$( curl -s -X POST                             \
         "$LANDSCAPE_URL/api/v2/computers/$COMPUTER_ID/children" \
@@ -292,7 +292,7 @@ $WSL_JSON = @{
  cloud_init = "$BASE64_ENCODED_CLOUD_INIT"
 } | ConvertTo-Json
 
-# Issue the command for each Windows machine:
+# Issue the command for each Windows machine
 foreach ($COMPUTER_ID in $PARENT_COMPUTER_IDS) {
     $API_RESPONSE = Invoke-WebRequest -Method POST -Body "$WSL_JSON" `
         -Uri "$LANDSCAPE_URL/api/v2/computers/$COMPUTER_ID/children" `
--- a/docs/howto/custom-rootfs-multiple-targets.md
+++ b/docs/howto/custom-rootfs-multiple-targets.md
@@ -115,7 +115,7 @@ $LOGIN_JSON = @{
 ````
 `````
 
-Issue an authenticate request and retrieve the JWT to be used in the subsequent API requests.
+Issue an authenticate request and retrieve the JSON web token (JWT) to be used in the subsequent API requests.
 
 `````{tabs}
 
--- a/docs/howto/custom-rootfs-multiple-targets.md
+++ b/docs/howto/custom-rootfs-multiple-targets.md
@@ -1,10 +1,11 @@
-# How-to deploy a custom rootfs across multiple Windows machines with Landscape API
+# How-to deploy a custom rootfs across multiple Windows machines with the Landscape API
 
-This guide shows how to use the Landscape API to deploy a custom rootfs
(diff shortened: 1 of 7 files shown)


CarlosNihelton:
CI Link check is now failing mainly due [this page about MS Store requirements](https://learn.microsoft.com/en-us/microsoft-store/prerequisites-microsoft-store-for-business) which no longer exists. I didn't find a suitable replacement for that page so far. Even asked on one MS forum.

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
+# How-to deploy a custom rootfs across multiple Windows machines with Landscape API
+
+This guide shows how to use the Landscape API to deploy a custom rootfs
+with some cloud-config user data for further initialisation of the distro
+instances via cloud-init across multiple Windows machines already registered
+with Landscape via Ubuntu Pro for WSL.
+It can be followed either on Windows (using PowerShell) or Linux (using bash).
```
edibotopic, line 7: I would put these in bullets:

> The steps outlined in this guide can be followed using
> - Bash scripting on Linux
> - PowerShell scripting on Windows

(I guess they could also use PowerShell on Linux but that's probably not relevant here...

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Prerequisites
+
+- A running self-hosted Landscape server version `24.10~beta.5` or later;
+- Multiple Windows machines already registered with Landscape via Ubuntu Pro for WSL;
+- Make sure you have installed `curl` and `jq`, if you're following this guide on a Linux machine.
+- Familiarity with Bash and/or PowerShell.
```
edibotopic, line 14: - Replace semi-colons with periods or else remove all of the semi-colons and periods
- You could convert "already registered with Landscape..." to a link that points to a relevant internal/external guide
- I would replace "on a Linux machine" with "using Bash"

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+
+## Prepare the environment
+
+To ease the subsequent commands, adjust and export the following environment variables:
```
edibotopic, line 18: Recommended change for clarity:

> For convenience when running later commands, first export the following environmental variables, modifying the values that are assigned as needed:

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+`````{tabs}
+
+````{group-tab} Bash
+```bash
+# Credentials to authenticate the API requests
+export LANDSCAPE_USER_EMAIL=admin@mib.com
+export LANDSCAPE_USER_PASSWORD=mib
+export LANDSCAPE_URL=https://landscape.mib.com
+
+# The URL of the custom rootfs to be deployed:
```
edibotopic, line 29: I would remove terminating colons in all code comments.

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+```powershell
+$LOGIN_JSON = @{
+ email = "$LANDSCAPE_USER_EMAIL"
+ password = "$LANDSCAPE_USER_PASSWORD"
+} | ConvertTo-Json
+```
+````
+`````
+
+Issue an authenticate request and retrieve the JWT to be used in the subsequent API requests.
```
edibotopic, line 117: Will JWT be understood by all readers?

Maybe, but if in doubt define on first use:

> the JSON Web Token (JWT)

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+$JWT = ConvertTo-SecureString -AsPlainText -Force $( $LOGIN_RESPONSE.Content | ConvertFrom-Json).token
+```
+````
+
+`````
+
+## Send the Install request
+
+Build the payload with information about the WSL instance to be deployed. In this case it would look like
+`{"rootfs_url": "http://landscape.mib.com:9009/ubuntu-24.04-custom.tar.gz", "computer_name": "Carbonizer", "cloud_init": <base64 encoded material>}`.
```
edibotopic, line 147: The in-line JSON is a little hard to read.
You could put it on a separate line or even in a (prettified) code block.

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+````
+`````
+
+When that completes, you'll be able to find activities in the Landscape
+dashboard about the installation of a new WSL instance for each of the Windows
+machines listed.
+
+## Summarising the steps in a single script
+
+The steps above can be made into a single script:
```
edibotopic, line 217: For the script to run successfully the environment variables would still need to be defined.
It might be worth noting before/after showing the complete script that the vars need to be defined in case anyone skips to the part of the guide with the complete script.

docs/howto/index.md
```diff
@@ -13,4 +13,5 @@ Back up and restore Ubuntu WSL instances <backup-and-restore>
 Uninstalling UP4W, Ubuntu WSL apps and WSL <uninstalling>
 Configure the Landscape client with UP4W  <set-up-landscape-client>
 Set up a Landscape server within WSL <set-up-landscape-server-in-wsl>
+Create WSL instances on Multiple Windows machines via Landscape API <custom-rootfs-multiple-targets>
```
edibotopic, line 16: no capital for "Multiple".

also, I would try to avoid _via_ (latinism) where possible and would replace it here with something like "with the".

docs/reference/firewall_requirements.md
```diff
@@ -19,7 +19,7 @@ The following table lists the default ports and protocols used by Ubuntu Pro for
 
 If the client system is behind a proxy, ensure that the proxy is configured to allow the required connections.
 
-[^1]: [Access to the Microsoft Store](https://learn.microsoft.com/en-us/microsoft-store/prerequisites-microsoft-store-for-business) is required for the online installation of WSL instances. Without it Ubuntu Pro for WSL will still be functional but it will not be possible to install WSL instances centrally from Landscape. In this case WSL instances have to be installed manually on the Windows hosts.
+[^1]: [Access to the Microsoft Store](https://learn.microsoft.com/en-us/microsoft-store/prerequisites-microsoft-store-for-business) is required for the online installation of WSL instances. Without it Ubuntu Pro for WSL will still be functional but it will only be possible to install WSL instances centrally from Landscape from custom tarballs, not from the official Ubuntu releases.
```
edibotopic, line 22: change to:

>[...] from Landscape using custom tarballs, not using the official Ubuntu releases.

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
+# How-to deploy a custom rootfs across multiple Windows machines with Landscape API
```
edibotopic, line 1: add "the" between "with" and "Landscape"

docs/howto/custom-rootfs-multiple-targets.md
```diff
@@ -0,0 +1,314 @@
...
+$ROOTFS_URL="http://landscape.mib.com:9009/ubuntu-24.04-custom.tar.gz"
+
+# The list of IDs of the different Windows machines on which we are going to deploy WSL instances:
+$PARENT_COMPUTER_IDS=@(26, 30, 31)
+
+# The name of the WSL instance to be created:
+$COMPUTER_NAME="Carbonizer"
+
+# Path to the cloud-config file whose contents will be used to initialize the WSL instances:
+$CLOUD_INIT_FILE="~/Downloads/init.yaml"
```
edibotopic, line 61: Just checking: is the unix-style path used here OK in the PS script?

edibotopic (changes requested):
This is great Carlos -- thank you.

It's also a really excellent use of the tabbed interface.

I have a few minor comments, mostly around wording but nothing major.

I could not find an alternative for that broken MS Store link either -- we may need to just remove it for now.

docs/howto/custom-rootfs-multiple-targets.md
CarlosNihelton: Yes, it it's fine. But it won't hurt to make it into Windows paths.

edibotopic: Yes, I was just thinking about someone reading this from a Windows-specific background.
In that sense, I was also wondering if the "~" is OK in the path but it's probably OK.

CarlosNihelton: Yeah, that's OK on Powershell. Not so OK for the old cmd.exe, but we don't want to attempt to write an equivalent batch script for that thing :)

edibotopic: It could have been the third tab Carlos!!! :laughing:

edibotopic:
LGTM: thanks for pulling this together!

edibotopic (approved):

PR #930, opened 2024-10-08
ci: Factor out building the MSIX

CarlosNihelton:
Making it in its own action will let us reuse that part when creating the publishing CI workflow.


didrocks (approved):
Glad we have it factored out now!

PR #941, opened 2024-10-16
feat(windows-agent): Allow users to run it via CLI

CarlosNihelton:
We modify the MSIX packaging to create an [app execution alias](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap3-appexecutionalias), which causes a dummy binary to be created in a well-known location (by default `%LOCALAPPDATA%\Microsoft\WindowsApps`) which is included in user's `%PATH%` such that it becomes possible for the user (or scripts running in user's behalf) to run the agent via CLI. That's preferrable over declaring the agent as an [Application](https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-application) because it won't be "promoted" in the start menu as an application would. Users can still find the CLI executable by searching for its exact name ( `ubuntu-pro-agent` in this case - the `.exe` is optional in the search but required in the appx manifest declaration).

Since we didn't remove the startup task declaration, starting the agent via the CLI app execution alias is also considered by the OS as a user interaction with the app, thus the startup task is enabled for that user such that the subsequent logons will have the agent starting automatically.

The agent is made into a single-instance application so we prevent a second instance to overwrite logs, replace the gRPC service bindings and corrupt the state of an already running process. We achieve that with a simple mechanism of writing a lock file in the agent's private directory, opened with exclusive access, so a process attempting to open it a second time won't succeed. Since the already-running instance could be child of the ubuntu-pro-agent-launcher.exe (which is an invisible GUI) it's better to not implement the common pattern of GUI apps of activating the already running process window when a instance detects itself as the second one.

The end result looks like the following:

Screenshot 2024-10-14 151735

image

- `ubuntu-pro-agent.exe` in PATH
- `ubuntu-pro-agent` can be found in the start menu
- A terminal window running it is and remains visible
- clean, completion, help and version commands still work on a second instance
- the default command does not work and exits early enough before touching logs, disk or network resources. 

--- 

UDENG-4623

P.S.: I sneaked one minor unrelated fix in 6c1f70ed395f70bb280d14352cea86ffc78b314e because I was passing nearby. Also 114d65600b862b5fbeac1cd6e1b4e8ec8898017f is temporary and should be fixed after I workout the documentation PR (coming next) related to this one.

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"strconv"
 	"strings"
 
 	"github.com/canonical/ubuntu-pro-for-wsl/common"
@@ -86,9 +87,18 @@ func New(o ...option) *App {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cleanup, err := a.ensureSingleInstance(o...)
+			if err != nil {
+				// We won't serve(), so let's close the ready channel right now.
+				// Otherwise callers of WaitReady will block forever.
+				close(a.ready)
+				return err
+			}
+			defer cleanup()
+
 			ctx := context.Background()
 
-			cleanup, err := a.setUpLogger(ctx)
+			cleanup, err = a.setUpLogger(ctx)
 			if err != nil {
 				log.Warningf(ctx, "could not set logger output: %v", err)
 			}
@@ -265,4 +275,30 @@ func (a *App) setUpLogger(ctx context.Context) (func(), error) {
 	log.Debug(ctx, "Debug mode is enabled")
 
 	return func() { _ = f.Close() }, nil
+}
+
+func (a *App) ensureSingleInstance(args ...option) (cleanup func(), err error) {
+	var opt options
+	for _, f := range args {
+		f(&opt)
+	}
+
+	priv, err := a.privateDir(opt)
+	if err != nil {
+		return nil, fmt.Errorf("could not access the agent's private dir: %v", err)
+	}
+
+	path := filepath.Join(priv, "ubuntu-pro-agent.lock")
+	f, err := createLockFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := f.WriteString(strconv.Itoa(os.Getpid())); err != nil {
+		return nil, fmt.Errorf("could not write PID to lock file %s: %v", path, errors.Join(err, f.Close()))
+	}
+
+	return func() {
+		log.Warningf(context.Background(), "when releasing the lock file: %v", f.Close())
+	}, nil
 }
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
@@ -293,6 +293,9 @@ func (a *App) ensureSingleInstance(opt options) (cleanup func(), err error) {
 	if _, err := f.WriteString(pid); err != nil {
 		return nil, fmt.Errorf("could not write PID to lock file %s: %v", path, errors.Join(err, f.Close()))
 	}
+	if err := f.Sync(); err != nil {
+		return nil, fmt.Errorf("could not flush lock file %s: %v", path, errors.Join(err, f.Close()))
+	}
 
 	return func() {
 		log.Warningf(context.Background(), "when releasing the lock file: %v", f.Close())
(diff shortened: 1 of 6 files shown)


windows-agent/cmd/ubuntu-pro-agent/agent/lockfile_linux.go
```diff
@@ -0,0 +1,24 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"syscall"
+)
+
+// createLockFile tries to create or open an empty file with given name with exclusive access.
+// If the file already exists AND is still locked, it will fail.
+func createLockFile(path string) (*os.File, error) {
+	file, err := os.OpenFile(path, os.O_WRONLY|os.O_TRUNC|os.O_CREATE, 0600)
```
didrocks, line 13: nitpick: I would name that just `f` as the other instances

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
```diff
@@ -266,3 +276,25 @@ func (a *App) setUpLogger(ctx context.Context) (func(), error) {
...
+	}
+
+	path := filepath.Join(priv, "ubuntu-pro-agent.lock")
+	f, err := createLockFile(path)
+	if err != nil {
+		return nil, err
+	}
+
+	pid := strconv.Itoa(os.Getpid())
+	if _, err := f.WriteString(pid); err != nil {
```
didrocks, line 293: We are writing the pid inside the lock file, you should really flush it as we are not closing the file right away. Otherwise, we are unsure when the pid will be really written on disk.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
```diff
@@ -23,6 +24,64 @@ func init() {
...
+				err2 := os.WriteFile(filepath.Join(path, "breaking-lock.txt"), []byte{}, 0600)
+				require.NoError(t, errors.Join(err, err2), "Setup: couldn't break the lock file")
+			}
+
+			a := agent.NewForTesting(t, "", private)
+			a.SetArgs(tc.args...)
+
+			err := a.Run()
+			if tc.wantError {
+				require.Error(t, err, "Run should return an error")
```
didrocks, line 77: I think you should return after this, as we don’t check for anything else and remove the else nesting.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
```diff
@@ -23,6 +24,64 @@ func init() {
...
+	testcases := map[string]struct {
+		args                   []string
+		anotherInstanceRunning bool
+		breakLockFile          bool
+
+		wantError bool
+	}{
+		"Success when single instance": {args: []string{"version"}},
+
+		// Testing the 'clean' verb cannot run in parallel because it tries to delete files still opened by other test cases.
```
didrocks, line 39: We should definitively have a test for clean, though, even if it’s a separate test case, no? I’m unsure to understand as we have separate private directory for the lock file why we can’t run them in parallel.

didrocks (changes requested):
I have few questions, but it looks good overall!

Looking at the description, I think we should either rename "clean" to something a little bit more meaningfull, or this could also be a "force" option to the command which will then bypass the single lock execution.

I understand why you didn’t pick the addresse file as this lock file, which can reconnect, recreate on close. Maybe it’s something to comment in, as this could confuse us in the future, wdyt?

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
CarlosNihelton: The 'clean' verb removes agent data from both public and private directories, but it has no knowledge of the public and private dir options, as the options are not cached inside the App struct. Thus, it goes through `os.Getenv()` to figure out those directories and end up mesing with global state affecting other tests (not only in this test function btw). Some refactoring is needed to lift computation of the options so they can be passed to `a.installClean()` to make sure that verb respects the temporary directories created by the test cases as well as changing the `TestClean` function to take into account that we should no longer solely rely on the environment variables.

So, for now, I'm taking the easy approach of adding one test case in `TestClean`  (which is already forbidden from running in parallel) that checks whether the agent proceeds with 'clean' or bails out if a lock file exists.

CarlosNihelton:
> Looking at the description, I think we should either rename "clean" to something a little bit more meaningfull, or this could also be a "force" option to the command which will then bypass the single lock execution.

The "clean" verb kills all agent processes already running , it already has this "force" semantics. I'd not consider adding a force flag because without it the "clean" verb would have nothing left to do. The current implementation bypasses the single-instance check because it's not relevant in such a context. In fact only the root command bemefits from performing that check. As for renaming it, the only suitable verb that comes to my mind is "reset": its intensity seems more appropriate, but it also gives the impression that the agent will be running after it completes, which is not the case.

windows-agent/cmd/ubuntu-pro-agent/agent/agent_test.go
didrocks: Makes sense, thanks for detailing it!

didrocks (approved):
Thanks for the changes! I’m ok with the semantic and clean was already published. We can always revisit those kinds of admin commands in the future for instance.

PR #944, opened 2024-10-17
feat(docs): How-to guides about UP4W agent startup

### LINKED ISSUE #912: Start the agent upon installation with starting the App
In corporate setup, where the Pro app is deployed at scale from InTune for instance and the configuration is done through the registry, the admin wants to start the agent in the background with having to start the app.

If due to Windows restriction it is not possible to do it directly at the packaging level, we could start by documenting how to proceed.

CarlosNihelton:
As agreed in WS037, I'm bringing two pieces of documentation about how system administrators can have UP4W background agent start without explicit end user interaction.

One guide relies on being able to start the agent via CLI (reason why this PR is based on #941 ) by deploying an one-shot PowerShell script via Intune to be executed with the current user's credentials.

The other guide relies on the registry, but leverages [Intune Remediations](https://learn.microsoft.com/en-us/mem/intune/fundamentals/remediations) to make that approach more reliable, by having that tool automatically checking and fixing non-compliant state, thus multi-shot (should end users disable the startup task, Intune would eventually find it and fix it).

---

This PR completes UDENG-4623 and closes #912 .


docs/howto/enforce-agent-startup-remotely-registry.md
```diff
@@ -0,0 +1,118 @@
...
+if you need more detailed views of the steps outlined above.
+
+At next logon users covered by the assignment will have Intune executing the detection script and the remediation one
+if the device is found non-compliant.
+
+## Remarks
+
+Careful readers would notice that there is an inherent race condition between setting the registry value and installing
+the MSIX (if remotely deployed): when the MSIX is installed the referred registry sub-key is recreated, overwriting any
+previous value the remote management solution would have deployed if that happened before the package installation.
```
edibotopic, line 112:
```suggestion
Careful readers may notice that there is an inherent race condition between setting the registry value and installing
the MSIX (if remotely deployed): when the MSIX is installed the referred registry sub-key that is referenced is recreated, overwriting any
previous value that the remote management solution would have deployed if that happened before the package installation.
```

docs/howto/enforce-agent-startup-remotely-registry.md
```diff
@@ -0,0 +1,118 @@
...
+
+One advantage of Intune Remediation scripts in this scenario is that eventually Intune would find the non-compliant
+state and fix it automatically. One disadvantage of that scenario is that the fix doesn't start the UP4W background
+agent, i.e. it only enables the startup task, the agent will start at next logon.
```
edibotopic, line 116:
```suggestion
An advantage of Remediation scripts in this scenario is that eventually Intune finds the non-compliant
state and fixes it automatically. A potential disadvantage is that the fix doesn't start the UP4W background
agent. The fix enables the startup task and the agent will then start at next user login.
```

Other material that might be candidate for explanation article.

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
+# How to start the agent remotely
+
+Ubuntu Pro for WSL, being a Microsoft Store application, cannot ship user services as of the time of this writing (late
+2024), but can deploy startup tasks instead, programs that run with user permissions when the said user logs on the
+Windows device. The UP4W background agent runs as a startup task. By design a startup task is only enabled by the
+operating system when the user interacts with the application for the first time. While this behaviour is a feature for
+end users it presents a source of friction for deployments at scale, when system administrators expect zero-touch
+deployment of UP4W to just work.
```
edibotopic, line 8: This material seems broadly duplicated from the other guide.
While I understand it is relevant in both cases, I think it would be preferable if this more explanatory material, which provides valuable context for the guide(s), instead be captured in a dedicated Explanation article.

That explanation article could then be referenced in these guides, and any future guides pertaining to this topic, to provide context. As it stands, I think this explanatory material upsets the flow of the guides.

Consider truncating this if possible and we can try to encapsulate this in an explanation section next cycle.

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+This guide shows how system administrators can leverage Windows remote management solutions to start the agent once
+with user credentials, thus interacting with the application in user's behalf. Subsequent logons will have the UP4W
```
edibotopic, line 11:
```suggestion
with user credentials, thus interacting with the application on the user's behalf. Subsequent logins will have the UP4W
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
+# How to start the agent remotely
+
+Ubuntu Pro for WSL, being a Microsoft Store application, cannot ship user services as of the time of this writing (late
+2024), but can deploy startup tasks instead, programs that run with user permissions when the said user logs on the
+Windows device. The UP4W background agent runs as a startup task. By design a startup task is only enabled by the
+operating system when the user interacts with the application for the first time. While this behaviour is a feature for
+end users it presents a source of friction for deployments at scale, when system administrators expect zero-touch
+deployment of UP4W to just work.
+
+This guide shows how system administrators can leverage Windows remote management solutions to start the agent once
```
edibotopic, line 10:
```suggestion
This guide shows how system administrators can leverage Windows remote management solutions to start the UP4W background agent once
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+with user credentials, thus interacting with the application in user's behalf. Subsequent logons will have the UP4W
+background agent automatically as a consequence of such interaction.
```
edibotopic, line 12:
```suggestion
background agent started automatically as a consequence of the interaction.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+While this guide uses Intune, readers are expected to translate the steps into any remote management solution of their
+choice, as long as the said solution can run scripts with current user's credentials.
```
edibotopic, line 15:
```suggestion
While this guide uses Intune, readers can translate the steps for the remote management solution of their
choice, as long as the said solution can run scripts with current user's credentials.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+- At least one managed Windows device.
+- A Windows remote management solution.
+
+## Key takeaways
+
+1. Running a script as user to start the UP4W background agent makes it immediately available.
+2. Remote management solutions can be used to do that.
+3. The operating system considers the startup task enabled going forward.
+4. Subsequent logons will have the operating system starting the UP4W background agent automatically as expected.
```
edibotopic, line 27:
```suggestion
1. Running a script as user to start the UP4W background agent makes it immediately available.
2. Remote management solutions can be used to start the agent.
3. The startup task is then also enabled on the operating system.
4. On subsequent logins, the UP4W background agent starts automatically, as expected.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+While this guide uses Intune, readers are expected to translate the steps into any remote management solution of their
+choice, as long as the said solution can run scripts with current user's credentials.
+
+## Pre-requisites
+
+- At least one managed Windows device.
+- A Windows remote management solution.
+
+## Key takeaways
```
edibotopic, line 22: Again, consider Overview or Summary.

I think this section will be more useful when the initial explanatory material is removed at a later date.

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+The contents of the script can be far more elaborate, but for the purposes of this guide the following is enough:
+
+```powershell
+Write-Output "Starting the UP4W background agent remotely from Intune"
+ubuntu-pro-agent.exe
+```
+**Make sure to save that as UTF-8**, as required by Intune.
+
+Follow [this section from Intune documentation](https://learn.microsoft.com/en-us/mem/intune/apps/intune-management-extension#create-a-script-policy-and-assign-it)
+if you need more detailed step-by-step guide on how to create and assign script policies.
```
edibotopic, line 40: Maybe make this a Note block. Seems like an aside.

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+Access your organisation's [Intune Admin Center](https://intune.microsoft.com) and when logged in go to **Devices > Monitor > Manage Devices > Scripts and remediations**.
+On that page, click on the **Platform scripts** tab.
+![Platform scripts revealed under Devices > Scripts and remediations](./assets/intune-platform-scripts.png)
+
+Click on the "Add" button to create a new script policy and select the platform **Windows 10 and later**.
+
+Fill in the **Basics** form with Name and description for the script being created.
```
edibotopic, line 48:
```suggestion
Fill in the **Basics** form with a name and description for the script being created.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+In the **Settings** tab browse your machine to the PowerShell script to be deployed, and select the following options:
```
edibotopic, line 50:
```suggestion
In the **Settings** tab, browse your machine to the PowerShell script to be deployed, and select the following options:
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+- Run this script using the logged on credentials: Yes (the default). UP4W must run with user credentials.
+- Enforce script signature check: No (unless required otherwise by your company's policies)
```
edibotopic, line 52:
```suggestion
- Enforce script signature check: No (unless required otherwise by your company's policies).
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+- Run script in 64-bit PowerShell host: No (the default).
+
+Apply the "Scope tags" according to your company's practices and, in the "Assignments" make sure to select one or more
```
edibotopic, line 55:
```suggestion
Apply the "Scope tags" according to your company's practices and, in "Assignments" make sure to select one or more
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+groups encompassing the users which must receive and run the script.
```
edibotopic, line 56:
```suggestion
groups encompassing the users that must receive and run the script.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+You can then monitor for this script execution in the Intune Admin Center.
```
edibotopic, line 58:
```suggestion
You can then monitor the execution of this script in the Intune Admin Center.
```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+
+When the selected users log on their devices Intune will eventually start the UP4W background agent, a terminal window
+will be visible with its regular outputs, which might be a user experience issue. But remember that it will only happen
+once.
```
edibotopic, line 62:
```suggestion
When the selected users log in to their devices, Intune will eventually start the UP4W background agent. A terminal window
will be appear to the user showing its regular outputs. This will only happen
once.
```

I don't think we need to flag whether this is a UX issue or not, but just describe what happens.

A few things to consider clarifying briefly, if  necessary: what is meant by "regular outputs"; by "only happen once" do you mean once during each log in or once ever; does the user need to exit the terminal or does it disappear automatically?

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
...
+## Remarks
+
+Careful readers might have noticed that if the script is deployed in conjunction with a policy installing the UP4W, it
+would be theoretically possible to have the script running before the application gets installed. A more elaborate
+solution is required if that scenario is possible, like looping in the script or using a more proactive solution such as
+[Intune Remediations](https://learn.microsoft.com/en-us/mem/intune/fundamentals/remediations).
+
+[This guide](howto::enforce-with-intune) shows how to use that tool to enforce the UP4W startup task state.
+
+[modeline]: # ( vim: set tw=119: )
```
edibotopic, line 73:
```suggestion

```

docs/howto/start-agent-remotely.md
```diff
@@ -0,0 +1,73 @@
+# How to start the agent remotely
```
edibotopic, line 1: Consider whether there is a way to more clearly distinguish the titles of these two guides.

One is how to do X with the registry, this is how to do X with...?

In future, we might think about whether the guides should be combined.

edibotopic (changes requested):
Thanks for doing this @CarlosNihelton -- this is very useful information!

I have included several suggestions, mainly focused on wording and clarity.

Two broader, **non-blocking** points I would like us to think about (in subsequent discussions and PRs):

- **Explanatory content:** there is quite a bit of explanatory content. While this is good material, I think it might be better placed in a dedicated explanation article in the docs (something focused on remote management). This would ensure that the how-to guides flow and mainly contain the information necessary for someone to complete the tasks involved. In addition, this would help prevent some duplication across the two guides.
- **Separation of guides:** I think the guides might need a cleaner separation in their titles. When thinking about that in the future, we should also think whether they would be better combined.

docs/howto/enforce-agent-startup-remotely-registry.md
```diff
@@ -0,0 +1,118 @@
...
+## Using Intune Remediations
+
+Remediations are script packages that can detect and fix common issues on a user's device before they even realise
+there's a problem. Each script package consists of a detection script, a remediation script, and metadata. Through
+Intune, you can deploy these script packages and see reports on their effectiveness.
+[Visit the Microsoft Intune documentation](https://learn.microsoft.com/en-us/mem/intune/fundamentals/remediations)
+to learn more about it. Those scripts run on a predefined schedule and if the detection script reports a failure (by
+`exit 1`) then the remediation script will run. That allows system administrators to watch for the specific Registry
+key value that represents the enablement of the UP4W background agent startup task. The contents of both scripts are
+presented below. **Make sure to save them encoded in UTF-8**, as required by Intune.
```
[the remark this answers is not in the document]
CarlosNihelton: TBH I don't know how obvious would it be but I considered that out of scope of this guide. We'd be in the realm of documenting how to use text editors.

docs/howto/start-agent-remotely.md
CarlosNihelton: As how-to guides I believe they must remain separate, as they show two distinct approaches to not depend on the end-user interaction with the application and their slightly distinct implications. They could be fused in a single explanation page, though, but then we'd loose the straightforwardness of a how-to guide.

So I think the ultimate solution is to have an explanation or reference page about the UP4W background agent being a startup task,  covering the reason for such design, its limitations and referencing those how-to guides. That would allow removing the explanatory verbiage as you suggested.

edibotopic: Agreed!

edibotopic (approved):
LGTM! Much thanks.

Issue #878, opened 2024-08-22
Filter *.pem files in the file chooser


jibel:
On the Landscape configuration page, the certificate field needs a pem file and even provides as example c:\landscape.pem but the file chooser is set to display *.*. It should filter only *.pem by default with *.* as another option.

![image](https://github.com/user-attachments/assets/f5c7caf8-43dc-430b-9f1b-9264959e52c2)

ashuntu:
It looks like the [`file_picker`](https://pub.dev/packages/file_picker) package we use doesn't support a "default" option and a fallback option at the same time. Either we allow any files to be picked or limit to a set of file extensions (`'cer', 'crt', 'der', 'pem'`).

CarlosNihelton:
Let's limit the set of file extensions as said above. The worse thing that can happen is a missing element in the set, for which someone will open a bug and we'll add it accordingly.

PR #947, opened 2024-11-07
fix(gui-ubuntupro): Add filters for cert files in file picker

ashuntu:
The file picker for Landscape configuration certificates now filters to only .cer, .crt, .der, and .pem file types.

image

Fixes https://github.com/canonical/ubuntu-pro-for-wsl/issues/878

---

UDENG-5110

### DIFF
--- a/gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
+++ b/gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
@@ -27,7 +27,7 @@ void main() {
   // This should be resolved so that we don't have to specify a manual text scale factor.
   // See more: https://github.com/flutter/flutter/issues/108726#issuecomment-1205035859
   binding.platformDispatcher.textScaleFactorTestValue = 0.6;
-  FilePicker.platform = FakeFilePicker([ca_cert]);
+  FilePicker.platform = FakeFilePicker([caCert]);
 
   group('input sections', () {
     testWidgets('default state', (tester) async {
@@ -156,7 +156,7 @@ void main() {
       await tester.pumpAndSettle();
 
       final fileInput = find.ancestor(
-        of: find.text(ca_cert),
+        of: find.text(caCert),
         matching: find.byType(TextField),
       );
       expect(fileInput, findsOne);
@@ -164,7 +164,7 @@ void main() {
       await tester.tap(fileInput);
       await tester.pump();
 
-      await tester.enterText(fileInput, client_cert);
+      await tester.enterText(fileInput, clientCert);
       await tester.pump();
 
       await tester.tap(continueButton);
@@ -278,7 +278,7 @@ void main() {
       await tester.pumpAndSettle();
 
       final fileInput = find.ancestor(
-        of: find.text(ca_cert),
+        of: find.text(caCert),
         matching: find.byType(TextField),
       );
       expect(fileInput, findsOne);
@@ -404,8 +404,8 @@ Widget buildApp(
 
 const customConf = './test/testdata/landscape/custom.conf';
 const notFoundPath = './test/testdata/landscape/notfound.txt';
-const ca_cert = './test/testdata/certs/ca_cert.pem';
-const client_cert = './test/testdata/certs/client_cert.pem';
+const caCert = './test/testdata/certs/ca_cert.pem';
+const clientCert = './test/testdata/certs/client_cert.pem';
 
 class FakeFilePicker extends FilePicker {
   /// Fake [FilePicker] that always returns the given `paths`.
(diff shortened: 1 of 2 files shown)


ashuntu:
> Did you try adding a simple test case overriding the `FilePicker.platform` instance with a custom implementation or a mock so you can trigger the file picker and ensure the lines you've changed have coverage?

I tried mocking this but ran into issues because of the `platform` singleton. I think maybe we'd have to wrap the picker in a service to be able to mock it and that didn't seem very worthwhile to me since we don't really test any actual behavior anyway. Let me know if you know of an easier way to do this.

CarlosNihelton:
> > Did you try adding a simple test case overriding the `FilePicker.platform` instance with a custom implementation or a mock so you can trigger the file picker and ensure the lines you've changed have coverage?
> 
> I tried mocking this but ran into issues because of the `platform` singleton. I think maybe we'd have to wrap the picker in a service to be able to mock it and that didn't seem very worthwhile to me since we don't really test any actual behavior anyway. Let me know if you know of an easier way to do this.

Yeah, the wrapper idea is usually the most appropriate one, but for this case is overkill. Since we don't rely a whole lot on the file picker you can simply add this line `  FilePicker.platform = FakeFilePicker();` early in landscape_page_test.dart's main function (ofc you then need to implement the FakeFilePicker class by extending the original FilePicker abstract class). That allows you to trigger the button click and assert the results based on the behaviour you program the fake class to have. Let's pair when you're back if you need more details on the suggested approach.

ashuntu:
I added a fake FilePicker and tested some basic functionality.

gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
```diff
@@ -360,3 +404,34 @@ Widget buildApp(
...
+    bool withReadStream = false,
+    bool lockParentWindow = false,
+    bool readSequential = false,
+  }) {
+    return Future(
+      () async => FilePickerResult(
+        paths.map((p) => PlatformFile(name: p, path: p, size: 0)).toList(),
+      ),
+    );
+  }
```
CarlosNihelton, line 436: This can be simplified into:

```suggestion
  }) async {
    return FilePickerResult(
        paths.map((p) => PlatformFile(name: p, path: p, size: 0)).toList(),
    );
  }
```

CarlosNihelton: or even

```suggestion
  }) async =>
    FilePickerResult(
        paths.map((p) => PlatformFile(name: p, path: p, size: 0)).toList(),
    )
```

CarlosNihelton: The mere fact that you declare your function (or method) `async` makes the return type a `Future` even if you just return a value without actually performing any asynchronous step.

gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
```diff
@@ -245,8 +267,30 @@ void main() {
...
 
-      final errorText = find.text(lang.landscapeFQDNError);
-      expect(errorText, findsOne);
+      final fqdnErrorText = find.text(lang.landscapeFQDNError);
+      expect(fqdnErrorText, findsOne);
+
+      final chooseFileButton = find.ancestor(
+        of: find.text(lang.landscapeFilePicker),
+        matching: find.byType(FilledButton),
+      );
```
CarlosNihelton, line 276: Unless the button semantics are strictly important for the test case, I prefer to find buttons my test will tap just by `find.text()` (or `find.button()` from `yaru_test`). It's harder to miss a change in the localizable text getter than a change in the button type, which may lead to a more likely test failure in the future.

gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
```diff
@@ -145,6 +147,26 @@ void main() {
 
       await tester.enterText(fqdnInput, 'test.l.com');
       await tester.pump();
+
+      final chooseFileButton = find.ancestor(
+        of: find.text(lang.landscapeFilePicker),
+        matching: find.byType(FilledButton),
+      );
```
CarlosNihelton, line 154: Ditto (about  `find.text()` or `find.button()`).

CarlosNihelton (changes requested):
Almost there! Just minor simplifications.

gui/packages/ubuntupro/test/pages/landscape/landscape_page_test.dart
ashuntu: That's what I get for trusting the auto stub generation, should be good now :)

CarlosNihelton (approved):
Yes! That looks good to me now! Good job!

Issue #880, opened 2024-08-23
Verify that the certificate is really a certificate


jibel:
Pro App: v0.1.4-528-g0e4126b3-dirty

## Test Case
1. Launch the Pro App and proceed to the Landscape configuration page
2. Choose "Self-hosted configuration"
3. Enter a server address 
4. In the certificate field, select a file that is not a certificate e.g. a docx file
5. click on continue

## Expected result
It shouldn't be possible to continue since it will not be possible to connect to Landscape anyway

## Actual result
Nothing blocks the user to provide a file that is not a certificate.
furthermore, any instance creation will fail with a message that it cannot connect to Landscape.

PR #949, opened 2024-11-13
fix(gui-ubuntupro): Verify certificate files

ashuntu:
Adds verification for certificate files by first checking their file extension, and, if possible, their content. If a certificate path isn't a valid certificate, an error is shown in the UI and continuing is blocked.

<details>
<summary>Example invalid certificate file</summary>

image

</details>

Fixes https://github.com/canonical/ubuntu-pro-for-wsl/issues/880

---

UDENG-5109


gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
```diff
@@ -236,24 +248,38 @@ class LandscapeSelfHostedConfig extends LandscapeConfig {
...
+      _fileError = FileError.invalidFormat;
     } else {
-      _fileError = FileError.none;
+      final content = file.readAsBytesSync();
+      try {
+        X509.fromDer(content);
+        _fileError = FileError.none;
+        // Various exception or errors can occur here when attempting a parse
+        // ignore: avoid_catches_without_on_clauses
+      } catch (_) {
```
ashuntu, line 269: This leads to `analyze` errors despite being ignored. This is needed since both Exceptions during decoding happen as well as various Errors, both meaning that the cert is invalid (the tests now include valid and invalid binary files for testing this purpose). Maybe there's a better way to do this other than error catching but I don't believe anything is built into X509 to check for cert validity.

CarlosNihelton: You can always catch by the base class (`Exception` in this case) as you did below. I quickly look at the underlying parsing routines and they seem to only throw Exception-derived objects. 

If you've seen `Error`-derived objects then we should have a pairing session to look into it more carefully. `Error` objects are by definition to not be caught.

CarlosNihelton: I'd also move this part into its own function because it's not the prettiest code we've ever written :)

CarlosNihelton (commented):
This code is more reliable for sure. Just a little more stylish touches and we'll be good to go.

gui/packages/ubuntupro/lib/pages/landscape/landscape_model.dart
ashuntu: This is an Error that can occur if I try to load the `not_a_cert.pem` from `testdata/`:

```
[+12597 ms] [ERROR:flutter/runtime/dart_vm_initializer.cc(41)] Unhandled Exception: RangeError:
Value not in range: 106
[        ] #0      _rangeCheck (dart:typed_data-patch/typed_data_patch.dart:5119:5)
[        ] #1      _ByteBuffer.asUint8List (dart:typed_data-patch/typed_data_patch.dart:1928:5)
[        ] #2      new Uint8List.view (dart:typed_data:903:19)
[        ] #3      ASN1Parser.nextObject (package:pointycastle/asn1/asn1_parser.dart:74:30)
[        ] #4      new X509.fromDer (package:pkcs7/src/x509.dart:25:25)
[        ] #5      LandscapeSelfHostedConfig._validatePath
(package:ubuntupro/pages/landscape/landscape_model.dart:264:12)
[        ] #6      LandscapeSelfHostedConfig.sslKeyPath=
(package:ubuntupro/pages/landscape/landscape_model.dart:283:9)
[        ] #7      LandscapeModel.setSslKeyPath
(package:ubuntupro/pages/landscape/landscape_model.dart:102:16)
[        ] #8      _FilePickerFieldState.build.<anonymous closure>
(package:ubuntupro/pages/landscape/landscape_page.dart:477:31)
[        ] <asynchronous suspension>
```
*(This is with the code slightly modified to actually output errors so the line numbers aren't accurate. The offending call is `X509.fromDer(content);` in `_validatePath`)*

Most other files I've tried will indeed lead to Exceptions instead of Errors, but not this one.

CarlosNihelton: Ok. I see what's going on. Your catch-all approach makes sense. Let's just move this piece into its own function and voilá!

CarlosNihelton (approved):
Good job!

PR #963, opened 2024-11-19
maint: Add additional files to ignore

ashuntu:
At the top level, .vscode/ gets generated any time a custom vscode setting is set for the workspace, leading to changes that don't necessarily need committed (for example, my spell checker that I add spelling exceptions to).

In gui/ubuntupro/, devtools_options.yaml gets generated any time you open VS Code's devtools. Since we don't use any custom options here, I think this can be ignored.


CarlosNihelton (approved):
🚢

PR #965, opened 2024-11-19
maint: Bump Flutter and Yaru major versions

ashuntu:
Updates Flutter and Yaru to the newest major versions. As a side effect, some other packages had to be bumped as well as some minor code updates.

----

UDENG-5290


CarlosNihelton (approved):
Did you notice more rounded corners?
image

I'm happy to see the analyzer is smarter about unnecessary null checks!

ashuntu:
> Did you notice more rounded corners?

I didn't notice much of a difference, but looking at it it looks to be slightly different

PR #967, opened 2024-11-19
fix(gui-ubuntupro): Remove popup feedback when applying a token

ashuntu:
Removes the popup as well as the test that asserts its existence. The tests for the pro token field that already exist should be enough to test error functionality.

---

UDENG-5286


CarlosNihelton (approved):
Yeap! Straight to the point!

PR #968, opened 2024-11-20
fix: Streamline token errors

ashuntu:
Removes specific errors like "too short" or "too long". Also, removes any error message when the token field is cleared. This change also streamlines some of the tests so we test multiple cases of different kinds of invalid tokens, but now each results in the same error.

---

UDENG-5285


gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
```diff
@@ -129,20 +129,14 @@ class ProTokenValue extends EitherValueNotifier<TokenError, ProToken?> {
 
 extension TokenErrorl10n on TokenError {
   /// Allows representing the [TokenError] enum as a String.
-  String localize(AppLocalizations lang) {
+  String? localize(AppLocalizations lang) {
```
CarlosNihelton, line 132: Smart move!

CarlosNihelton (approved):
Good job!

PR #978, opened 2024-11-21
fix: Remove accidental inclusion of an unused import

ashuntu:
This happened because #967 removed a test that was no longer needed (as well as this import), but was not removed in main before the branch for #968 was created. Since the import had to change for the relocation of common tokens, git didn't merge the changes quite correctly.


CarlosNihelton (approved):
Thanks for looking into this!

PR #980, opened 2024-11-25
feat: Always show Pro token input

ashuntu:
Shows the token input field always instead of being in a collapsible section. Also, adds messaging when a valid token or invalid token is added, with an icon.

A small detail I missed when adjusting the copy for errors previously: invalid tokens now show the string "Invalid token" instead of "Token invalid".

<details>
<summary>Screenshot of the default look</summary>

image

</details>

<details>
<summary>Screenshot when an invalid token is entered</summary>

image

</details>

<details>
<summary>Screenshot when a valid token is entered</summary>

image

</details>

Our designs in Figma include functionality that we do not have, so I took some creative liberties in adapting it to the current layout. Let me know if it needs any adjustments.

---

UDENG-5284

### DIFF
--- a/gui/packages/ubuntupro/lib/constants.dart
+++ b/gui/packages/ubuntupro/lib/constants.dart
@@ -1,3 +1,5 @@
+import 'package:flutter/material.dart';
+
 /// The name of the file where the Agent's drop its service connection information.
 const kAddrFileName = '.ubuntupro/.address';
 
@@ -12,3 +14,5 @@ const kVersion = String.fromEnvironment(
   'UP4W_FULL_VERSION',
   defaultValue: 'Dev',
 );
+
+const kConfirmColor = Color(0xFF0E8420);
--- a/gui/packages/ubuntupro/lib/constants.dart
+++ b/gui/packages/ubuntupro/lib/constants.dart
@@ -1,5 +1,3 @@
-import 'package:flutter/material.dart';
-
 /// The name of the file where the Agent's drop its service connection information.
 const kAddrFileName = '.ubuntupro/.address';
 
@@ -14,5 +12,3 @@ const kVersion = String.fromEnvironment(
   'UP4W_FULL_VERSION',
   defaultValue: 'Dev',
 );
-
-const kConfirmColor = Color(0xFF0E8420);
--- a/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
+++ b/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
@@ -3,7 +3,6 @@ import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter_gen/gen_l10n/app_localizations.dart';
 import 'package:yaru/yaru.dart';
-import '../../constants.dart';
 import '../../core/either_value_notifier.dart';
 import '../../core/pro_token.dart';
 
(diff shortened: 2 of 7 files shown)


gui/packages/ubuntupro/lib/constants.dart
```diff
@@ -12,3 +14,5 @@ const kVersion = String.fromEnvironment(
   'UP4W_FULL_VERSION',
   defaultValue: 'Dev',
 );
+
+const kConfirmColor = Color(0xFF0E8420);
```
ashuntu, line 18: Maybe we should look at replacing this custom color with a built-in color in the future. It's one of the only custom colors we use in the app (other than `Color(0xFFE6F8E8)` in one place on the subscription status page).

CarlosNihelton: Interestingly this constant is the value of `YaruColors.success`.
https://ubuntu.github.io/yaru.dart/
https://github.com/ubuntu/yaru.dart/blob/main/lib/src/themes/colors.dart#L25-L39

I think we should use the Yaru value deliberately now that's available.

CarlosNihelton (commented):
I think your proposal finds a good balance between the stuff we know we need to change and the amount we don't know yet.
One of the key benefits of the recent releases of Yaru is the consistency brought by YaruColors. Thus my questions about whether we're approaching them properly or not.

ashuntu:
I converted the colors to Yaru colors. The status page also uses a modified version of success that is extremely close to the custom value we were using.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
```diff
@@ -61,58 +61,97 @@ class _ProTokenInputFieldState extends State<ProTokenInputField> {
...
+                  decoration: InputDecoration(
+                    hintText: lang.tokenInputHint,
+                    error: _token.errorOrNull?.localize(lang) != null
+                        ? Padding(
+                            padding: const EdgeInsets.only(top: 4),
+                            child: Row(
+                              children: [
+                                Icon(
+                                  Icons.cancel,
+                                  color: YaruColors.of(context).error,
```
ashuntu, line 97: Is there a preference for `theme.colorScheme` vs `YaruColors.of`?

CarlosNihelton: For the error case they would mean the same, but `YaruColors` offers `.link`, `.warning` and `.success` which I don't think `theme.colorScheme` does. So, YaruColors is preferrable when those semantics may change. For ease, let's just stick to `YaruColors`.

CarlosNihelton:
> I converted the colors to Yaru colors. The status page also uses a modified version of `success` that is extremely close to the custom value we were using (is there a built-in version of this I don't know about?).

I suspect if you use the `YaruInfo` widget you'll get the correct colors for free ;)

ashuntu:
> I suspect if you use the `YaruInfo` widget you'll get the correct colors for free ;)

That looks promising, I think I will do that in another PR since its unrelated to the token page.

gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_widgets.dart
```diff
@@ -47,17 +47,18 @@ class SubscriptionStatus extends StatelessWidget {
         Container(
           padding: const EdgeInsets.all(16.0),
           decoration: BoxDecoration(
-            border: Border.all(color: const Color(0xFF0E8420), width: 1.0),
-            color: const Color(0xFFE6F8E8),
+            border: Border.all(color: theme.colorScheme.success, width: 1.0),
+            color: theme.colorScheme.success
+                .copyWith(lightness: 0.94, saturation: 0.56),
```
CarlosNihelton, line 52: AFAIK `.success` is an extension method brought you by yaru.dart. So, there is no benefit in not explicitely depending on YaruColors instead.

CarlosNihelton (approved):
Good job!

PR #982, opened 2024-11-26
fix(windows-agent): Delegates `agent.WaitReady` to the daemon

CarlosNihelton:
Since #896 the agent might sometimes close the `a.ready` channel too early, giving the possibility of a call to `a.WaitReady()` exiting before the daemon really starts serving, thus not really waiting.

The deamon knows best when is the most likely moment to be serving, thus this PR slightly modifies `a.WaitReady()` to check with the daemon, which had to implement its own `daemon.WaitReady()`.

The `a.ready` channel is still useful to synchronize the access to `a.daemon`, preventing nil pointer access when the deamon is not yet initialized. Checking for `nil` is still needed, as the `a.ready` channel may be closed without ever serving, such as on a failure case.

This adds a test for `daemon.WaitReady()`, but the agent tests don't need to change, as their "flakyness" reveal the problem.

---

UDENG-4900

### DIFF
--- a/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
+++ b/windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
@@ -186,6 +186,10 @@ func (a *App) Quit() {
 // Note: we need to use a pointer to not copy the App object before the daemon is ready, and thus, creates a data race.
 func (a *App) WaitReady() {
 	<-a.ready
+	if a.daemon == nil {
+		return
+	}
+	a.daemon.WaitReady()
 }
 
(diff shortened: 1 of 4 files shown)


didrocks (approved):
This looks the right way to fix it to me. Let’s hope that definitively fixes that flackyness.

I do have one request: do you mind double checking on the systemd "ready" notification side that we are still behaving as expected?

CarlosNihelton:
> do you mind double checking on the systemd "ready" notification side that we are still behaving as expected?

By the time wsl-pro-service reports a conclusive "ready" notification to systemd, it's already talking to the agent. That part of the code didn't change, as this PR only touches the Windows side. ;)

PR #984, opened 2024-11-27
fix: Move about link and fix broken link

ashuntu:
Moves the About Ubuntu Pro button functionality to a Learn More link in the above copy. Also fixes the issue of the link not opening.

Screenshot of change

---

UDENG-5282

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
+++ b/gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
@@ -36,7 +36,8 @@ class SubscribeNowPage extends StatelessWidget {
         SizedBox(
           width: 400,
           child: MarkdownBody(
-            data: lang.proHeading('[Learn more](https://ubuntu.com/pro)'),
+            data:
+                lang.proHeading('[${lang.learnMore}](https://ubuntu.com/pro)'),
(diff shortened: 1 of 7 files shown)


gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -21,12 +23,23 @@ class SubscribeNowPage extends StatelessWidget {
...
     return LandingPage(
       children: [
-        Text(
-          lang.proHeading,
-          style:
-              theme.textTheme.bodyLarge!.copyWith(fontWeight: FontWeight.w100),
+        SizedBox(
+          width: 400,
+          child: MarkdownBody(
+            data: lang.proHeading('[Learn more](https://ubuntu.com/pro)'),
```
CarlosNihelton, line 39: "Learn more" cannot be localized this way, but it should be. I think a better way to solve this problem is by having it as a separate l10n string.

```dart
data: lang.proHeading('[${lang.learnMore}](https://ubuntu.com/pro)'),
```

It's a good idea to avoid placing the markdown syntax directly in the localizable strings, as a translator could accidentally break the link for a specific language.

CarlosNihelton (changes requested):
Good approach. Just one minor concern about l10n.

CarlosNihelton (approved):
LGTM 👍🏽

PR #995, opened 2024-12-02
fix(windows-agent): Landscape TLS by default

CarlosNihelton:
Previously, if the Landscape server's SSL public key wasn't provided we'd default to insecure credentials for gRPC.
That's not nice when the system trusts the server cert. A browser would happily render the page as safe, but the agent would fail to connect to the server's hostagent messenger.
We should assume TLS by default. If the server cert path is not provided, load the system's cert pool and try to connect based on the system's chain of trust.
To avoid complicating the existing tests, we preserve the use of insecure credentials via a context value. If set, we don't even check the cert path.
Otherwise, TLS by default.

I considered but didn't implement a more global way of requesting insecure credentials. We could either pass a command line flag to the agent or use an environment variable at the top level then derive a context with the value set for insecure credentials. In my local tests, at least, I don't feel the need for that as I'm always running Landscape withs self-signed certificates.

---
UDENG-5412


ashuntu (approved):

windows-agent/internal/proservices/landscape/testdata/TestConnect/error_when_the_SSL_certificate_is_not_trusted/landscape.conf
```diff
@@ -0,0 +1,4 @@
+[client]
+
+[host]
+url = {{ .HostURL }}
```
ashuntu, line 4: Completely inconsequential, but no new line at the end of this file.

PR #996, opened 2024-12-03
fix: Vertically center window title bar controls

ashuntu:
Centers the title bar control buttons, but they will not be the full height of the title bar until Yaru is updated to include https://github.com/ubuntu/yaru.dart/pull/967 (will require a version bump on our end in the future).

| Existing | Current state of this PR | With the Yaru update |
|----------|--------------------------|----------------------|
| image | image | image |

---

UDENG-5289


CarlosNihelton (approved):
Happy to see that the fix is easier than we originally thought!

PR #997, opened 2024-12-05
feat: Skip landscape page

ashuntu:
This a fairly large PR since there are a few moving pieces to adding this page. The existing Landscape page needs its skip functionality removed, the skip page uses an updated version of copy, the skip page uses an entirely new layout, etc.

I plan on expanding the use of reusable navigation and other layout widgets since the new designs will be much more consistent across pages. Notably, when I reorganize the Landscape page, we can unify a lot of the navigation and tile widgets.

Screenshot of the new page

---

UDENG-5287


gui/packages/ubuntupro/test/pages/landscape_skip/landscape_page_test.dart
CarlosNihelton, line 1: The test file name doesn't match its counter part. Not a hard rule, but it's good for consistency.

gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -57,7 +57,12 @@
     "purchaseStatusServer":"Something went wrong with Microsoft Store. Please try again later.",
     "purchaseStatusUnknown": "Unknown error when trying to purchase the subscription. Consider restarting this app.",
 
-    "landscapeHeading": "Configure the connection to {landscapeLink} to manage your Ubuntu WSL instances remotely.",
+    "landscapeSkip": "Skip for now",
+    "landscapeSkipDescription": "You can always configure Landscape later",
+    "landscapeSkipRegister": "Register with Landscape",
+
+    "landscapeTitle": "Landscape",
```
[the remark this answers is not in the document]
ashuntu: I guess my initial goal here was more to centralize this term so we don't have to duplicate it elsewhere, risk misspelling, etc. I do agree that it shouldn't be translated, so maybe I could just make it a regular constant variable?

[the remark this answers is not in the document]
ashuntu: I committed what I suggested, but open to anything.

gui/packages/ubuntupro/lib/pages/widgets/page_widgets.dart
```diff
@@ -168,3 +169,91 @@ class _PageContent extends StatelessWidget {
...
+                                ),
+                              ),
+                              TextSpan(
+                                text: title,
+                                style: theme.textTheme.displaySmall
+                                    ?.copyWith(fontWeight: FontWeight.w100),
+                              ),
+                            ],
+                          ),
+                        ),
```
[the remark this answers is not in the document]
ashuntu: Our existing format supports arbitrary titles/images, so we'd have to drop support for that and just use full logos with titles in them, instead of text. If you think that's better I can do that.

I'm not sure, but is there an accessibility concern with making text into images, like for logos?

gui/packages/ubuntupro/lib/pages/landscape/landscape_page.dart
```diff
@@ -14,6 +14,7 @@ import 'package:yaru/yaru.dart';
 
 import '/core/agent_api_client.dart';
 import '/pages/widgets/page_widgets.dart';
+import '../../constants.dart';
```
CarlosNihelton, line 17:
```suggestion
import '/constants.dart';
```

gui/packages/ubuntupro/lib/pages/widgets/page_widgets.dart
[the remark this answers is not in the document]
CarlosNihelton: Good point about accessibility. Let's keep it as is then.

gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
```diff
@@ -0,0 +1,36 @@
...
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        OutlinedButton(
+          onPressed: onBack,
+          child: Text(backText ?? lang.buttonBack),
+        ),
+        FilledButton(
+          onPressed: onNext,
+          child: Text(nextText ?? lang.buttonNext),
+        ),
```
CarlosNihelton, line 32: I see both buttons as the same kind in the mockup for the Landscape skip page only (while all other pages have a filled button on the preferred action). Did you check with the designers on that? Because I see you're setting the button to be filled unconditionally.

gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
```diff
@@ -0,0 +1,36 @@
...
+  Widget build(BuildContext context) {
+    final lang = AppLocalizations.of(context);
+
+    return Row(
+      mainAxisAlignment: MainAxisAlignment.spaceBetween,
+      children: [
+        OutlinedButton(
+          onPressed: onBack,
+          child: Text(backText ?? lang.buttonBack),
+        ),
```
CarlosNihelton, line 28: Also, should the back button be conditionally added to the row? Consider the first page of the wizard: you either avoid using the navigation row and place the button at the same precise location on the page or you use the navigation row but then the back button should not be shown.

CarlosNihelton: BTW I'm ok if you plan to tackle this only by the time this becomes necessary, as long as you've got a plan :)

CarlosNihelton (approved):
Just some details on the navigation row and one import to make absolute, otherwise LGTM. Good job!

gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
ashuntu: The other pages in the mockups use filled buttons for the primary action, so I figured we should do the same here as well. I'll ask about this.

CarlosNihelton: I trust you to check if the Design team and create a new PR if necessary to avoid blocking this one.

PR #1006, opened 2024-12-06
fix(windows-agent): Check that IPs are convertible to ipv4

ashuntu:
If the system receives an IPv6 address, then running the agent with mocks (for example, when running integration tests) can result in panics. This fixes that issue by checking if the IP is convertible to Ipv4.

For context, this came up when I attempted to run the Flutter integration tests on a network and host that technically support IPv6.

### DIFF
--- a/windows-agent/internal/daemon/daemontestutils/networking_mock.go
+++ b/windows-agent/internal/daemon/daemontestutils/networking_mock.go
@@ -144,7 +144,7 @@ func getLocalPrivateIP() net.IP {
 		return nil
 	}
 	for _, addr := range addrs {
-		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() && ipnet.IP.IsPrivate() {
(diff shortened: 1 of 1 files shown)


windows-agent/internal/daemon/daemontestutils/networking_mock.go
```diff
@@ -144,7 +144,7 @@ func getLocalPrivateIP() net.IP {
 		return nil
 	}
 	for _, addr := range addrs {
-		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() && ipnet.IP.IsPrivate() {
+		if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() && ipnet.IP.IsPrivate() && ipnet.IP.To4() != nil {
```
CarlosNihelton, line 147: Very good finding! I'd update the function name to `getLocalPrivateIPv4` (and the comment above it) to reflect that requirement. After all the bug existed because that requirement wasn't explicit.

CarlosNihelton (approved):
Thanks for the finding!

PR #1010, opened 2024-12-10
fix: Use YaruInfoBox for notification boxes

ashuntu:
This effectively aligns the colors correctly with what is used by Yaru for info boxes, which you can see at https://ubuntu.github.io/yaru.dart under the "YaruInfo" tab. Our colors in light mode were identical to Yaru, but dark mode colors were not. This fixes that issue and simplifies the widget code significantly by just using the `YaruInfoBox`.

This page still needs more changes in the future, but this should be sufficient for notification colors for the time being.

| | Dark | Light |
|-|------|-------|
| Existing | image| image |
| New | image | image |

---

UDENG-5590

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_widgets.dart
+++ b/gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_widgets.dart
@@ -34,8 +34,9 @@ class SubscriptionStatus extends StatelessWidget {
         ),
       ),
     ).copyWith(
-      a: const TextStyle(
+      a: TextStyle(
         decoration: TextDecoration.underline,
+        color: theme.colorScheme.onSurface,
       ),
     );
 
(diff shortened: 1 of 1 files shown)


ashuntu:
We miss some opportunity for tests here, where we could mock a launch event for the success link. However, this page is changing somewhat significantly, so I don't think it would be worthwhile for a test for that right now.

gui/packages/ubuntupro/lib/pages/subscription_status/subscription_status_widgets.dart
```diff
@@ -30,62 +30,28 @@ class SubscriptionStatus extends StatelessWidget {
...
-            color: YaruColors.jet,
           ),
         ),
       ),
     ).copyWith(
-      a: const TextStyle(
+      a: TextStyle(
         decoration: TextDecoration.underline,
-        color: YaruColors.jet,
+        color: theme.colorScheme.onSurface,
```
CarlosNihelton, line 39: Is this color parameter still necessary?

CarlosNihelton (approved):
What a lovely diff! :)

PR #1011, opened 2024-12-10
fix: Fix radio background

ashuntu:
Adds a border to the radio tiles, as well as a background to unselected tiles.

| | Dark | Light |
|-|------|-------|
| Existing | image | image |
| New | image | image |

This is also based on https://github.com/canonical/ubuntu-pro-for-wsl/pull/997 and relies on it being merged first.

---

UDENG-5593


CarlosNihelton (approved):
Clever fix! Is that how other Ubuntu apps handle that quirk in `YaruSelectableContainer`?

ashuntu:
> Clever fix! Is that how other Ubuntu apps handle that quirk in `YaruSelectableContainer`?

I'm actually not sure. I feel like this is something we could propose to Yaru as a charge if we cared enough about it in the future.

PR #1015, opened 2024-12-13
fix: Reorganize subscribenow

ashuntu:
Reorganizes the subscribe now page to be in the new two-column layout.

This moves the ownership of the token to the model of the page, instead of the text field owning it as a stateful field. This allows us to submit from any button on the page, and control the state of the text field independently. This results in a surprising about of side effects, as the previous logic had the input field handling virtually everything.

Note that the diff may be misleading, a lot of it is new mock generation and fixing existing tests to pass with the new ownership model.

image

---

UDENG-5589


gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
```diff
@@ -35,8 +53,25 @@ class SubscribeNowModel {
...
-  /// it's value from outside, the value is cached so we don't check the environment more than once.
-  bool get purchaseAllowed => _isPurchaseAllowed;
+  void tokenUpdate(String token) async {
+    _token.update(token);
+    notifyListeners();
+  }
+}
+
+/// A value-notifier for the [ProToken] with validation.
+class ProTokenValue extends EitherValueNotifier<TokenError, ProToken?> {
```
CarlosNihelton, line 63: I suspect we don't need to have this as a ValueNotifier if the SubscribeNowModel is a ChangeNotifier.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_widgets.dart
```diff
@@ -1,25 +1,27 @@
...
+import 'package:provider/provider.dart';
+import 'package:url_launcher/url_launcher_string.dart';
 import 'package:yaru/yaru.dart';
-import '../../core/either_value_notifier.dart';
 import '../../core/pro_token.dart';
+import 'subscribe_now_model.dart';
 
 /// A validated text field with a submit button that calls the supplied [onApply]
 /// callback with the validated Pro Token when the submit button is clicked.
 class ProTokenInputField extends StatefulWidget {
```
CarlosNihelton, line 13: Since the text editing controller is managed elsewhere I no longer see why this has to be a stateful widget.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
```diff
@@ -35,8 +52,29 @@ class SubscribeNowModel {
...
+  bool get hasError => either.isLeft;
+
+  void update(String token) {
+    either = ProToken.create(token);
+  }
+
+  void clear() {
+    either = const Either.right(null);
+  }
 }
```
CarlosNihelton, line 80: We can get rid of this class and use the Either directly.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
```diff
@@ -2,14 +2,31 @@ import 'package:agentapi/agentapi.dart';
...
 import '/core/pro_token.dart';
 
-class SubscribeNowModel {
+class SubscribeNowModel extends ChangeNotifier {
   final AgentApiClient client;
+  P4wMsStore store;
+
+  final _token = ProTokenValue();
+  ProTokenValue get token => _token;
+  bool get canSubmit => token.token != null;
```
CarlosNihelton, line 15: We can get rid of the ProTokenValue class and use the Either directly here. Or even completely remove the usage of eithers and do like we do in the Landscape page, where we store both the value and the error simultaneously. Personally, I prefer the Either monad, because if we have an error we don't want to keep the value.

```suggestion
  final _token = Either<TokenError,ProToken?>;
  ProToken? get token => _token.orNull();
  bool get canSubmit => token.isRight();
```

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
```diff
@@ -35,8 +52,29 @@ class SubscribeNowModel {
...
   }
 
-  /// Returns true if the environment variable 'UP4W_ALLOW_STORE_PURCHASE' has been set.
-  /// Since this reading won't change during the app lifetime, even if the user changes
-  /// it's value from outside, the value is cached so we don't check the environment more than once.
-  bool get purchaseAllowed => _isPurchaseAllowed;
+  void tokenUpdate(String token) async {
+    _token.update(token);
+    notifyListeners();
+  }
```
CarlosNihelton, line 58: If you accept removing the ProTokenValue class but still use eithers you can inline this:

```suggestion
  void tokenUpdate(String raw) async {
    _token = ProToken.create(raw);
    notifyListeners();
  }
```

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
CarlosNihelton: I suspect we can also delete the EitherValueNotifier, but I'm not 100% sure. You should check if that's possible and do accordingly.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -26,83 +29,82 @@ class SubscribeNowPage extends StatelessWidget {
...
+        showBack: false,
+        onBack: null,
+        onNext: model.canSubmit ? () => trySubmit(model) : null,
+        nextText: lang.attach,
+      ),
     );
   }
 
+  void trySubmit(SubscribeNowModel model) {
+    model.applyProToken(model.token.token!).then(onSubscriptionUpdate);
```
CarlosNihelton, line 97: I'm sure you can make this look better: `model.token.token!`.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_model.dart
CarlosNihelton: Similarly, the model itself could have a `clear()` method:

```dart
void clear() {
    _token = const Either.right(null);
  }
```

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -26,83 +29,82 @@ class SubscribeNowPage extends StatelessWidget {
...
+    model.token.clear();
```
CarlosNihelton, line 98: This could be `model.clear()` if you implement the method in the model per my suggestion above.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -26,83 +29,82 @@ class SubscribeNowPage extends StatelessWidget {
...
-              ElevatedButton(
-                onPressed: model.purchaseAllowed
-                    ? () async {
-                        final subs = await model.purchaseSubscription();
+        if (model.purchaseAllowed) ...[
+          const SizedBox(height: 16.0),
+          OutlinedButton(
+            onPressed: model.purchaseAllowed
+                ? () async {
+                    final subs = await model.purchaseSubscription();
```
CarlosNihelton, line 49: Juan's design suggested to add a button that redirects to `u.c/pro/subscribe` if MS Store Purchase is not implemented/allowed. Did you notice that?

CarlosNihelton (changes requested):
Some things to improve, but overall that's a great start!

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -10,14 +10,17 @@ import 'package:wizard_router/wizard_router.dart';
 
 import '/core/agent_api_client.dart';
 import '/pages/widgets/page_widgets.dart';
-
+import '../widgets/navigation_row.dart';
```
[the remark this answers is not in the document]
ashuntu: A bit annoying that it auto-adds imports in this way.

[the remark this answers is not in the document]
CarlosNihelton: Yeah, we have lint rules for package imports vs relative imports, but this "absolute-style relative imports" doesn't seem to bother people around. I find it so nice for importing stuff outside of this particular directory and I'm totally against package imports for stuff that are part of the same package.

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
ashuntu: I didn't initially, but I think that's a good idea so I've implemented it

gui/packages/ubuntupro/lib/pages/subscribe_now/subscribe_now_page.dart
```diff
@@ -9,15 +9,18 @@ import 'package:url_launcher/url_launcher_string.dart';
...
 import 'subscribe_now_model.dart';
 import 'subscribe_now_widgets.dart';
 
 class SubscribeNowPage extends StatelessWidget {
-  const SubscribeNowPage({super.key, required this.onSubscriptionUpdate});
+  SubscribeNowPage({super.key, required this.onSubscriptionUpdate});
+
   final void Function(SubscriptionInfo) onSubscriptionUpdate;
 
+  final controller = TextEditingController();
```
CarlosNihelton, line 22: I could swear I'd have written about this: we're essentially leaking resources by not disposing this text editing controller. Objects like this one require calling `.dispose()` when they're no longer needed. When instantiated by widgets, those should be stateful ones, as their `.dispose()` method is called by the framework itself, so that's the perfect place to drop resources.

CarlosNihelton (changes requested):
We're almost there! ah, please don't forget to update the documentation subsequently.

PR #1016, opened 2024-12-17
fix(docs): Explicits the agent's requirement on the server SSL public key

CarlosNihelton:
Until version 0.1.15 (the current version tag) of Ubuntu Pro for WSL, the app explicitly requires referencing a path to the SSL certificate on a Windows host machine.

The fix was implemented in #995. Now the agent completely follows the Windows OS certificate stores, only requiring reference to that certificate if the machine running the Landscape server is not trusted on a particular network. But we didn't issue a new release since then. So, to avoid a new release so close to the end-of-year holidays, let's update the docs in an atomic way, so we can easily revert if we deem worth it after the new year and a new release comes out.

The changes herein introduced look like the following:

image

and

image

---

UDENG-5654

### DIFF
--- a/docs/howto/set-up-landscape-client.md
+++ b/docs/howto/set-up-landscape-client.md
@@ -57,25 +57,25 @@ When you continue a status screen will appear confirming that configuration is c
 
 ![Configuration is complete](../assets/status-complete.png)
 
-### Note on SSL public key
+```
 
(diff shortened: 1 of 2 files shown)


edibotopic (approved):
LGTM: thanks for doing this, Carlos!

PR #1017, opened 2024-12-17
fix: Reorganize subscription status

ashuntu:
Reorganizes the subscription status page (the last past) according to new designs. Notably, this changes some copy, moves management links to footer text links, and uses generally the same look no matter what type of management.

image

*Note that this page will change significantly in the future to support Landscape configuration statuses. This is just a UI update only.*

---

UDENG-5591


ashuntu:
This removes some of the copy about being an org managed/store managed/manually managed page. I'm not sure if we still want those elements, but I think maybe we could replace the manage subscription link at the bottom with "This subscription is managed by your organization" when org managed. Or, keep it as is. I don't believe we have new designs for this.

ashuntu:
`LandscapePage` can likely be removed after https://github.com/canonical/ubuntu-pro-for-wsl/pull/1015 is merged.

gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -35,18 +35,9 @@
...
-    "@manuallyManaged": {
-        "placeholders": {
-            "proDashboardLink": {
-                "type": "String"
-            }
-        }
-    },
+    "ubuntuProEnabled": "Ubuntu Pro is enabled on this machine.",
+    "ubuntuProEnabledInfo": "All Ubuntu WSL instances have access to Ubuntu Pro security features.",
+    "ubuntuProManage": "Manage Ubuntu Pro subscription",
```
CarlosNihelton, line 40:
```suggestion
    "manageUbuntuPro": "Manage Ubuntu Pro subscription",
```

gui/packages/ubuntupro/lib/pages/widgets/page_widgets.dart
```diff
@@ -170,6 +170,76 @@ class _PageContent extends StatelessWidget {
...
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return Pro4WSLPage(
+      body: Padding(
+        padding: const EdgeInsets.fromLTRB(32.0, 24.0, 32.0, 32.0),
+        child: Column(
+          children: [
+            Expanded(
+              child: Row(
```
CarlosNihelton, line 197: Why we need the Row here? Wouldn't a Center widget be more appropriate?

gui/packages/ubuntupro/test/pages/subscription_status/subscription_status_page_test.dart
```diff
@@ -7,15 +7,59 @@ import 'package:ubuntu_service/ubuntu_service.dart';
...
 
 void main() {
   group('subscription info:', () {
     final client = FakeAgentApiClient();
     registerServiceInstance<AgentApiClient>(client);
     final info = SubscriptionInfo();
+
+    group('footer links:', () {
+      final landscape = LandscapeSource()..ensureOrganization();
+      testWidgets('user', (tester) async {
```
CarlosNihelton, line 24: Can we have a similar test case that finds no footer links for the org case?

CarlosNihelton:
>  I think maybe we could replace the manage subscription link at the bottom with "This subscription is managed by your organization" when org managed.

I like this suggestion better. I'm missing some feedback about this. We might ask Juan about it during our next UX review, if you prefer. I honestly consider that a useful piece of information.

CarlosNihelton (changes requested):
Very clean start. I like the way you've been introducing those changes piecewise. I have few things to address.

gui/packages/ubuntupro/test/pages/subscription_status/subscription_status_page_test.dart
ashuntu: I have `expect(find.text(lang.ubuntuProManage), findsNothing);` in the other org test cases, I figured that was enough?

CarlosNihelton: Yeap, that may suffice indeed. But if we go with your suggestion for the org case it'd be good to have it more specific.

CarlosNihelton: Now that we're sure we won't get that copy, let's call the day :)

CarlosNihelton (approved):
Well done!

PR #1018, opened 2024-12-18
maint(gui-ubuntupro): Remove generated mock files

ashuntu:
Removes all generated `*.mocks.dart` files from the ubuntupro apps. These were removed since they aren't necessarily stable and can be generated locally and in CI anyway by running `dart run build_runner build`. This also means that developers no longer have to make sure to regenerate mocks when modifying the tests or models.

---

UDENG-5592


CarlosNihelton (approved):
Glad to align the practices of the Apps team.

PR #1040, opened 2025-01-06
fix: Unify casing of select words and phrases

ashuntu:
Removed and unnecessary period, as well as correctly some words that were capitalized when they shouldn't be.

---

UDENG-5683


CarlosNihelton (approved):
LGTM

PR #1041, opened 2025-01-06
fix: Set a minimum size for the window

ashuntu:
Sets a minimum window size to the current default size. The window may be resized to be larger but cannot be resized to be smaller. This also centers the window on the screen when it opens instead of being in the top left corner.

This works, but it might be worth considering an update to https://github.com/ubuntu/yaru_window.dart to include a `setMinimumSize` and `setMaximumSize` method to `YaruWindow` instead of having to use `window_manager` directly.

---

UDENG-5682

### DIFF
--- a/gui/packages/ubuntupro/lib/main.dart
+++ b/gui/packages/ubuntupro/lib/main.dart
@@ -1,7 +1,9 @@
 import 'package:flutter/material.dart';
 import 'package:ubuntu_service/ubuntu_service.dart';
+import 'package:window_manager/window_manager.dart';
 import 'package:windows_single_instance/windows_single_instance.dart';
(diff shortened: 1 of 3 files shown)


CarlosNihelton:
> This works, but it might be worth considering an update to https://github.com/ubuntu/yaru_window.dart to include a setMinimumSize and setMaximumSize method to YaruWindow instead of having to use window_manager directly.

I don't disagree in principle, but it's just easier to do what you already did. Also, since yaru_window itself depends on window_manager and yaru on yaru_window, either way we're not pulling more dependencies that what we already do. So this implementation should be just fine.

CarlosNihelton (approved):
LGTM!

PR #1042, opened 2025-01-07
fix: Set a maximum size of page contents

ashuntu:
Sets a maximum width for page content, equal to their current effective width.

Here's a comparison for this change (maximized on a 1440p screen):

| Old | New |
|-----|-----|
| image | image |

The default size of the app remains unchanged.

*This diff is really just the content being wrapped with a new widget, nothing else changed*

---

UDENG-5684


CarlosNihelton (approved):
Great!

It feels as though GH created the "Hide whitespace" feature for Flutter developers :)

image

PR #1043, opened 2025-01-08
feat: Add widget for delaying errors on text fields

ashuntu:
Adds a delay to error messages while typing in a text field. Notably, there is no delay when errors are removed.

---

UDENG-5681

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/widgets/delayed_text_field.dart
+++ b/gui/packages/ubuntupro/lib/pages/widgets/delayed_text_field.dart
@@ -1,0 +1,82 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+
+/// A [TextField] that displays error messages on a delay instead of
+/// immediately.
+class DelayedTextField extends StatefulWidget {
+  DelayedTextField({
+    this.autofocus = false,
+    this.enabled = true,
+    this.controller,
+    this.error,
+    this.errorText,
+    this.hintText,
+    this.inputFormatters,
+    this.label,
+    this.onChanged,
+    this.onSubmitted,
+    super.key,
+  });
+
+  final bool autofocus;
+  final TextEditingController? controller;
+  final bool enabled;
+  final Widget? error;
+  final String? errorText;
+  final String? hintText;
+  final List<TextInputFormatter>? inputFormatters;
+  final Widget? label;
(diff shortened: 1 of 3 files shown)


gui/packages/ubuntupro/lib/pages/widgets/delayed_text_field.dart
```diff
@@ -0,0 +1,129 @@
...
+  TimerNotifier({required this.duration, required this.vsync})
+      : assert(duration > Duration.zero, 'Duration must be greater than zero') {
+    _ticker = vsync.createTicker(_onTick);
+  }
+
+  final TickerProvider vsync;
+  final Duration duration;
+
+  Ticker? _ticker;
+  Duration _elapsedTime = Duration.zero;
```
CarlosNihelton, line 95: I just realized that, since we don't need to accumulate the duration between onTick callbacks, we no longer need this field.

gui/packages/ubuntupro/lib/pages/widgets/delayed_text_field.dart
```diff
@@ -0,0 +1,129 @@
...
+    _ticker?.start();
+  }
+
+  /// Callback executed on each tick.
+  void _onTick(Duration elapsed) {
+    _elapsedTime = elapsed;
+
+    if (_elapsedTime >= duration) {
+      _ticker?.stop();
+      notifyListeners();
```
CarlosNihelton, line 126:
```suggestion
      scheduleMicrotask(notifyListeners());
```

Just a minor performance enhancement. Scheduling a microtask allow the current frame to finish rendering before notifying listeners, which prevents jittering and/or artifacts. Not an issue for this simple app, but it's a good practice overall.

gui/packages/ubuntupro/lib/pages/widgets/delayed_text_field.dart
```diff
@@ -0,0 +1,126 @@
...
+  @override
+  void dispose() {
+    _ticker?.dispose();
+    super.dispose();
+  }
+
+  /// Starts the timer.
+  void start() {
+    _ticker?.start();
+  }
```
CarlosNihelton, line 107: We're not calling this method, so you can get 100% test coverage by removing it :)

Sorry for not noticing it earlier, just now I looked into codecov's complaint and realized it :)

CarlosNihelton (approved):
Great job!

PR #1045, opened 2025-01-09
fix(gui-ubuntupro): Add placeholder padding to text fields to prevent jittering

ashuntu:
When entering text into a text field that may display an error, the text field and other content on the page will move to accomodate the error text. This change adds a SizedBox when there is no error to prevent this movement. As a result, page content is naturally longer, so the default window height has been increased by 50 pixels.

image

*Depends on https://github.com/canonical/ubuntu-pro-for-wsl/pull/1043*

---

UDENG-5680


CarlosNihelton (approved):
Impressive that it ended up being so simple! Good job!

PR #1046, opened 2025-01-10
fix(gui-ubuntupro): Hide inactive errors

ashuntu:
Hides errors for inactive fields since they are not relevant; when reactivated the errors will reappear.

Also fixes a small bug that caused fields not to be properly disabled on the Landscape page.

---

UDENG-5732


CarlosNihelton (approved):
LGTM

PR #1053, opened 2025-01-14
docs: Update documentation screenshots to latest designs

ashuntu:
Updates documentation screenshots to latest designs. Note that some of these haven't changed much, but the default size of the window has changed so each image is a slightly different size.

---

UDENG-5775


edibotopic (approved):
LGTM: thanks for doing this.

PR #1068, opened 2025-01-16
fix(wsl-pro-service): Prevents unnecessary re-registration with Landscape 

CarlosNihelton:
This is a two-fold fix.

## Re-registration due whitespace changes

I noticed that wsl-pro-service was causing the machine to re-register with Landscape at every startup. There is an early routine that checks for the validity of the Landscape client config file contents, amending it if necessary and re-running landscape-config **if anything changed**. The current implementation also considers whitespace changes as a valid reason to re-run landscape-config, which shouldn't be the case.

An instrumented build to diff the contents before and after that routine revealed that we're taking whitespace changes too seriously.

```
n@mib01:~$ sudo /mnt/d/wsl-pro-service -vvv
INFO Starting WSL Pro Service version Dev
INFO /home/cn/Dev/1_Work/up4w/wsl-pro-service/cmd/wsl-pro-service/service/config.go:48 initViperConfig() No configuration file: Config File "wsl-pro-service" Not Found in "[/home/n /root /etc /mnt/d]"
DEBUG /home/cn/Dev/1_Work/up4w/wsl-pro-service/cmd/wsl-pro-service/service/service.go:64 New.func1() Debug mode is enabled
DEBUG /home/cn/Dev/1_Work/up4w/wsl-pro-service/internal/system/landscape.go:74 (*System).fixAndEnableLandscapeFromConfig() Landscape: something changed.
 [client]
-account_name = standalone
-url = https://lg4070.olivec.dev/message-system
-ping_url = http://lg4070.olivec.dev/ping
-log_level = info
+account_name     = standalone
+url              = https://lg4070.olivec.dev/message-system
+ping_url         = http://lg4070.olivec.dev/ping
+log_level        = info
 registration_key = 123
-tags = wsl
-hostagent_uid = 36589e24a0e24eecaaea3c04eee2afd4
-computer_title = Ubuntu-22.04
+tags             = wsl
+hostagent_uid    = 36589e24a0e24eecaaea3c04eee2afd4
+computer_title   = Ubuntu-22.04
```

## landscape-config would always create a new registration

Secondly, re-running landscape-config would always create a new registration, even if the client configuration change is minor and unrelated to the server, such as the client debug level. 

landscape-client (24.08+git6375-0ubuntu0) added the new --register-if-needed flag for landscape-config, allowing it to refresh an existing registration instead of unconditionally creating a new one for a computer already registered with Landscape. There are situations when wsl-pro-service subprocesses lansdcape-config on an already registered computer. That creates a new registration and, even though Landscape server has some logic to detect that the new registration belongs to an existing computer, it loses data that takes long to recover, such as package version information.

Since the landscape-client version in the latest-stable PPA is already over that version we should be able to add this flag to wsl-pro-service and prevent that behaviour.

---

UDENG-5761

### DIFF
--- a/wsl-pro-service/internal/system/landscape.go
+++ b/wsl-pro-service/internal/system/landscape.go
@@ -7,9 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"slices"
 	"strings"
-	"unicode"
 
 	log "github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logstreamer"
 	"github.com/ubuntu/decorate"
@@ -61,13 +59,13 @@ func (s *System) fixAndEnableLandscapeFromConfig(ctx context.Context, landscapeC
 		return fmt.Errorf("could not parse config: %v", err)
 	}
 
-	modifiedLandscapeConfig, err := normalizeLandscapeConfig(ctx, s, iniFile)
+	modifiedLandscapeConfig, didChange, err := normalizeLandscapeConfig(ctx, s, iniFile)
 	if err != nil {
 		return err
 	}
 
 	// No change to do, do not rewrite config
-	if !enableUnconditionally && areConfigsEqual(modifiedLandscapeConfig, landscapeConfig) {
+	if !enableUnconditionally && !didChange {
 		log.Debug(ctx, "Landscape configuration is already valid")
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


wsl-pro-service/internal/system/landscape.go
```diff
@@ -75,14 +77,24 @@ func (s *System) fixAndEnableLandscapeFromConfig(ctx context.Context, landscapeC
...
 	if _, err := runCommand(cmd); err != nil {
 		return fmt.Errorf("could not enable Landscape: %v", err)
 	}
 
 	return nil
 }
 
+func areConfigsEqual(a, b string) bool {
+	fieldsFunc := func(r rune) bool {
+		return unicode.IsSpace(r) || r == '='
```
ashuntu, line 90: This is a very niche edge case, but if a URL (or any other config value) contains an `=`, then the config wouldn't be correctly reregistered if the `=` was subsequently removed/added. This might not be a valid scenario in this case, but just a thought

ashuntu (approved):

wsl-pro-service/internal/system/landscape.go
CarlosNihelton: I'm not aware of that possibility, but let's not take the risk. The functions doing the changes can and should report whether anything changed or not. So let's leverage that and avoid string comparisons at all.

ashuntu (approved):

PR #1070, opened 2025-01-22
fix: Add option to use non-elevated buttons for navigation

ashuntu:
Adds an option to use elevated or outlined buttons for "next" navigation. Non-action buttons should be outlined instead of elevated.

image
*Note "Next" is outlined instead of green*

### DIFF
--- a/gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
+++ b/gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
@@ -9,6 +9,7 @@ class NavigationRow extends StatelessWidget {
     this.nextText,
     this.showBack = true,
     this.showNext = true,
+    this.nextIsElevated = true,
     super.key,
   });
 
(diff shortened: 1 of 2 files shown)


gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
```diff
@@ -32,10 +34,15 @@ class NavigationRow extends StatelessWidget {
           ),
         if (showNext) ...[
           const Spacer(),
-          ElevatedButton(
-            onPressed: onNext,
-            child: Text(nextText ?? lang.buttonNext),
-          ),
+          nextIsElevated
```
CarlosNihelton, line 37: I'd rather have this parameter name closer to the semantics, instead of reflecting an implementation detail. Something like: `nextIsSubmit` or `emphasizeNext` or swap its meaning and name it `equalEmphasis`, or come up with something better yet, can we?

CarlosNihelton (changes requested):
It's somewhat of a nitpick, naming is always hard...

gui/packages/ubuntupro/lib/pages/widgets/navigation_row.dart
ashuntu: `nextIsAction`?

CarlosNihelton (approved):
LGTM!

PR #1071, opened 2025-01-22
fix: Fix spacing between fields with errors

ashuntu:
Fixes the spacing for the Landscape page by aligning the page to the top so errors push the fields down, instead of being aligned center and pushing the entire page.

Also, increases the spacing slightly above the token input field.

New token field spacing:
image

New Landscape page spacing:
image

For reviewers: note this should work when the window is maximized as well

---

UDENG-5853


CarlosNihelton (approved):
It does what it says! Good job!

PR #1075, opened 2025-01-28
docs: Fix typo

YanisaHS:
I think this is a typo. I don't actually have UP4W set up so I can't verify, but pretty sure. (Just passing by)


CarlosNihelton (approved):
Thanks for noticing! 👍🏽

PR #1077, opened 2025-01-29
deps: Upgrading the Go toolchain

CarlosNihelton:
Due a vulnerability report:

Vulnerability #1: GO-2025-3373
    Usage of IPv6 zone IDs can bypass URI name constraints in crypto/x509
  More info: https://pkg.go.dev/vuln/GO-2025-3373
  Standard library
    Found in: crypto/x509@go1.23.1
    Fixed in: crypto/x509@go1.23.5


didrocks (approved):
This solves the workspace issue! However, for package building, you need to bump also individual `go.mod` theorically for each subprojects, as they are not workspace aware.

As we are forcing in deb packaging `GOTOOLCHAIN=local`, I am fine we skip this one.

PR #1094, opened 2025-01-31
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [ubuntu-desktop-translations/Ubuntu Pro for WSL](https://hosted.weblate.org/projects/ubuntu-desktop-translations/ubuntu-pro-for-wsl/).

Current translation status:

Weblate translation status

---

UDENG-266

PR #1095, opened 2025-01-31
feat: Add empty ARB files for translation languages

ashuntu:
This is so we don't have to manually add the languages in Weblate itself.

Purposely excludes French and Brazilian Portuguese since those are added by https://github.com/canonical/ubuntu-pro-for-wsl/pull/1094.

---

UDENG-266


CarlosNihelton (approved):
LGTM

PR #1097, opened 2025-01-31
ci: Delete certificate files after importing

ashuntu:
These were leading to a dirty git history, which made the version in the Flutter app show as "dirty"


CarlosNihelton:
Good catch! I confirmed [this CI build](https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/13080227051/artifacts/2519187200) is not dirty!

image

CarlosNihelton (approved):
🚢

PR #1098, opened 2025-01-31
ci: Do not run end to end tests for translation changes

ashuntu:
End to End tests can't run when translations are submitted via Weblate, so just skip running those workflows when the only files that change are translations

---

UDENG-266


CarlosNihelton (approved):
LGTM

PR #1104, opened 2025-02-10
deps: Upgrade go toolchain

CarlosNihelton:
Motivation:
Vulnerability https://github.com/canonical/ubuntu-pro-for-wsl/pull/1: GO-2025-3447
    Timing sidechannel for P-256 on ppc64le in crypto/internal/nistec
  More info: https://pkg.go.dev/vuln/GO-2025-3447
  Standard library
    Found in: crypto/internal/nistec@go1.23.5
    Fixed in: crypto/internal/nistec@go1.23.6
    
CI is pretty broken due that.

PR #1105, opened 2025-02-10
fix: Handle potentially long translation strings

ashuntu:
Some strings for other languages could overflow the GUI, so this accounts for vertical increases in the layout of each page by wrapping the content in a scrollable container. The CenteredPage widget used for the status page should never realistically overflow, but its scrollable just in case :) (and it will be expanded in the future).

Also allows more than 1 line for error/hint messages, and truncates long label texts.

<details>
<summary>Label truncation and multiline error example (PL translations)</summary>

image
</details>

---

UDENG-5996


CarlosNihelton:
CI Go-related failures are addressed in #1104 .

CarlosNihelton (approved):

PR #1106, opened 2025-02-11
feat(windows-agent): Landscape executor to install modern distros

CarlosNihelton:
I think this is better reviewed commit by commit, as there are some "peasant fixes" along the way which do not exactly strictly meet the core of this PR, but they were small enough to not deserve a new PR (I think :) ).

This "feature" addresses one specific problem made more evident now that the latest stable version of WSL handles "modern distros" (those whose distribution model no longer relies on MS Store nor Windows application packages). The problem is: when Landscape sends a message to install an instance, we execute `wsl --install` if that's a stock Ubuntu release. That used to install an APPX, so we'd later run the distro launcher to provision the new instance. But now, `wsl --install Ubuntu` or `wsl --install Ubuntu-24.04` would install the modern, tar-based, distro, which no longer has a launcher to subprocess. So we replace the launcher invocation with just subprocessing `cloud-init status --wait` to wait for the complete provisioning and later check if the default user was created or not, creating one behind the scenes (preserving the current behaviour).

To make the error condition of launcher not found distinguishable I create a specific error type. PowerShell always exits 1 for any error, so that's not enough to determine which error was. Instead we need to look into it's output. There is a bunch of small test-related shenanigans I had to do to exercise some error paths along the way.

---

UDENG-5929

### DIFF
--- a/windows-agent/internal/proservices/landscape/distroinstall/distroinstall.go
+++ b/windows-agent/internal/proservices/landscape/distroinstall/distroinstall.go
@@ -16,6 +16,15 @@ import (
 	"github.com/ubuntu/gowsl"
 )
 
+// CommandNotFoundError is the error reported when a command is not found.
+type CommandNotFoundError struct {
+	errMsg string
+}
+
+func (e *CommandNotFoundError) Error() string {
+	return e.errMsg
+}
+
 // InstallFromExecutable finds the executable associated with the specified distro and installs it.
 func InstallFromExecutable(ctx context.Context, d gowsl.Distro) error {
 	executable, err := common.WSLLauncher(d.Name())
@@ -24,7 +33,13 @@ func InstallFromExecutable(ctx context.Context, d gowsl.Distro) error {
 	}
 
 	if out, err := executableInstallCommand(ctx, executable); err != nil {
-		return fmt.Errorf("could not run launcher: %v. %s", err, out)
+		// executableInstallCommand returns a generic error if the command is not found (because it relies on powershell under the hood).
+		// We need to look inside the output to determine if the command was not found.
+		msg := string(out)
+		if strings.Contains(msg, "CommandNotFoundException") {
+			return &CommandNotFoundError{errMsg: fmt.Sprintf("could not find command %q: %s", executable, msg)}
+		}
+		return fmt.Errorf("could not run launcher: %v. %s", err, msg)
 	}
 
 	return err
--- a/windows-agent/internal/proservices/landscape/executor.go
+++ b/windows-agent/internal/proservices/landscape/executor.go
@@ -343,7 +343,6 @@ func installFromURL(ctx context.Context, homeDir string, downloadDir string, dis
 	if err := touchdistro.WaitForCloudInit(ctx, distro.Name()); err != nil {
 		log.Infof(ctx, "cloud-init failed: %v", err)
 	}
-	_ = distro.Terminate()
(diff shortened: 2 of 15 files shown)


windows-agent/internal/proservices/landscape/distroinstall/distroinstall.go
```diff
@@ -16,6 +16,15 @@ import (
 	"github.com/ubuntu/gowsl"
 )
 
+// CommandNotFoundError is the error reported when a command is not found.
+type CommandNotFoundError struct {
+	errMsg string
+}
+
+func (e *CommandNotFoundError) Error() string {
```
ashuntu, line 24: Could this be this instead?

```suggestion
func (e CommandNotFoundError) Error() string {
```

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -295,20 +288,59 @@ func (e executor) shutdownHost(ctx context.Context, cmd *landscapeapi.Command_Sh
...
+// createDefaultUser creates a default user whose name is derived from the current Windows user.
+func createDefaultUser(ctx context.Context, distro gowsl.Distro) (err error) {
+	userName := "ubuntu"
+
+	windowsUser, err := user.Current()
+	if err != nil {
+		return err
+	}
+	userFields := strings.Split(windowsUser.Username, "\\") // Typically "hostname\\username"
+	n := userFields[len(userFields)-1]
```
ashuntu, line 331:
```suggestion
	n := userFields[len(userFields) - 1]
```

windows-agent/internal/proservices/landscape/distroinstall/distroinstall.go
CarlosNihelton: While it could for the purposes we use here, it's a best practice to make the  `Error()` method with a pointer-receiver. That has to do with how type-assertions work. If we'd make that a value-receiver instead, as you're suggesting, both expressions would be compilable: `if var err error = SomeFun(); err.(CommandNotFoundError)` and `if var err error = SomeFun(); err.(*CommandNotFoundError)` but one of them would be broken, depending on whether `func SomeFunc() error` returned a pointer or a value (both implementations would be valid from the compiler's perspective). It's a quite obscure problem, to be honest. So sticking to the convention allows the error handler to not care about the possibility of the `err` containing a value or a pointer. It's a pointer (or nil if no error) and that's it.

By making the `Error()` method with a pointer-receiver we constrain the type system such that the `CommandNotFoundError` type doesn't satisfy the `error` interface but the pointer to it (`*CommandNotFoundError`) does. Thus, checking for the value type in a type assertion makes no sense.

This link demonstrates how that's possible. 

https://go.dev/play/p/FYiyBUdlSvZ

windows-agent/internal/proservices/landscape/executor.go
CarlosNihelton: The go formatter doesn't like your suggestion.

windows-agent/internal/proservices/landscape/distroinstall/distroinstall.go
ashuntu: Ah that makes sense, ignore my suggestion then :)

windows-agent/internal/proservices/landscape/executor.go
ashuntu: Weird since it seems like we use spaces between operands elsewhere

ashuntu (approved):

PR #1133, opened 2025-02-24
docs: Better headers and metas

edibotopic:
Search results for these docs are primarily determined by the top-level header and the meta descriptions used.
We currently don't set meta descriptions, so the first line of text after the title is used in search results.

For example, here is search result for installing Ubuntu Pro for WSL. It is possible that a user reads the title (from the title of the page) and description (defaulting to first line after title) and thinks "what is UP4W?"

Screenshot from 2025-02-24 10-08-00

In this example, it would be difficult to tell that this page even relates to WSL, unless someone looks carefully at the URL:

Screenshot from 2025-02-24 10-06-58

The changes in this PR attempt to add more descriptive titles to individual pages, which will be more informative when read as an isolated search result; in addition, meta descriptions are added so that we have more control over how pages are described in these search results.

The markdown meta information included at the top of a given markdown pages gets converted into html, like so:

```html
<meta content="A complete Ubuntu environment on Windows using WSL, with enhanced security and remote management provided by Ubuntu Pro for WSL." lang="en" name="description" xml:lang="en" />
```

Other changes:

* Dev docs notice: adds a notice when a page is primarily intended for a developer audience, to better account for a scenario where a user ends up on the installation guide for developers, for example
* Redirects: added some more redirects to account for the recent merge and based on a survey of 404 frequency for the documentation.
* Fixups: spacing, wording and other small issues

UDENG-6053

### DIFF
--- a/docs/explanation/ref-arch-explanation.md
+++ b/docs/explanation/ref-arch-explanation.md
@@ -1,4 +1,11 @@
-# UP4W architecture
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "The different components of Ubuntu Pro for WSL work together to support automatic securing of WSL instances and integration with remote management tools."
+---
+
+# Architecture of Ubuntu Pro for WSL
 
 This page describes the different components of UP4W and how they integrate
 together to form the software architecture.
--- a/docs/custom_conf.py
+++ b/docs/custom_conf.py
@@ -128,8 +128,6 @@ slug = ""
 # NOTE: If this variable is not defined, set to None, or the dictionary is empty,
 # the sphinx_reredirects extension will be disabled.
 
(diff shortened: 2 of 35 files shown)


docs/explanation/ref-arch-explanation.md
```diff
@@ -1,4 +1,11 @@
-# UP4W architecture
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "The different components of Ubuntu Pro for WSL work together to support automatic securing of WSL instances and integration with remote management tools."
+---
```
CarlosNihelton, line 6: Just learned this new syntax!

CarlosNihelton (approved):
Great job phrasing those summarised descriptions!

PR #1137, opened 2025-02-25
fix(wsl-pro-service): wait for cloud-init before fixing Landscape client.conf

CarlosNihelton:
With the current behaviour wsl-pro-service and cloud-init race for the Landscape client.conf file during first boot.
It's possible that wsl-pro-service passes by the routine supposed to fix that configuration file before it exists, thus the instance wouldn't be registered with Landscape before a reboot.

We cannot simply order them via the systemd units because the Landscape configuration happens in cloud-init's final stage. The sync point for that stage is cloud-init.target. But that systemd target `Wants=multi-user.target` so it won't be reached before the login finishes. `wsl-pro.service`on the other hand declares itself as  `WantedBy=multi-user.target` so it has a systemd target against which it's installed (otherwise it wouldn't be enabled), which makes it a dependency for the boot, thus making it impossible to make it `After=cloud-init.target` or `Wants=cloud-init.target`. That would result in a cycle.
There is no other meaningful systemd target we could make `wsl-pro.service` a dependency of to solve this cycle.

So let's sync in code. 

I'm doing the now popular snippet to check if the cloud-init.service is enabled before running `cloud-init --wait` to prevent hangs in case the unit goes disabled. I'm not checking if systemd is running because we are a child of systemd. Since I'm doing the check right before attempting to touch the client.conf file, and waiting on cloud-init could make systemd time us out, I'm moving that check to a place after we notify systemd that we are ready, thus unblocking the boot. I'm also adding a test case to make sure a broken client.conf won't interfere with wsl-pro-service's communication with the windows agent.

As a final quality-of-life measure I'm adding this line to the systemd unit:

```diff
[Service]
Type=notify
+ NotifyAccess=all
ExecStart=/usr/libexec/wsl-pro-service
Restart=always
RestartSec=2s
```
because I saw systemd logging many times complaints about wsl-pro-service notifying readiness from another PID than the one systemd originally started. This should allow any subprocess or thread of wsl-pro-service to notify it without that annoying log line.

---
UDENG-5890

### DIFF
--- a/.github/workflows/qa-azure.yaml
+++ b/.github/workflows/qa-azure.yaml
@@ -90,7 +90,9 @@ jobs:
       - name: Set up Ubuntu
         uses: Ubuntu/WSL/.github/actions/wsl-install@main
         with:
+          # TODO: Migrate this to the tar-based 24.04
           distro: "Ubuntu-Preview"
+          useStore: true
       - name: Set up Go
         # actions/setup-go is broken
         shell: powershell
--- a/wsl-pro-service/internal/daemon/daemon.go
+++ b/wsl-pro-service/internal/daemon/daemon.go
@@ -137,6 +137,12 @@ func (d *Daemon) Serve(service streams.CommandService) error {
 		return fmt.Errorf("could not notify systemd: %v", err)
 	}
 
+	// Since this function syncs with cloud-init and may take too long to run, let's do it after notifying
+	// systemd about our readiness to prevent delaying boot.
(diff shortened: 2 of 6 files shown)


ashuntu (approved):

ashuntu:
Oops, looks like I opened a PR for the Ubuntu-Preview change: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1142

Feel free to merge this one instead and I can close my PR.

Issue #1107, opened 2025-02-11
docs: TYPE YOUR QUESTION HERE


JonH130:
*Please describe the question or issue you're facing with "Get started with UP4W - Ubuntu on WSL documentation".*

"The install link below will work only if you’re logged in to the Microsoft Store with an account for which access to the app has been enabled."

Clearly I'm not special because I am logged into my Microsoft account and the link provided comes back with "⚠️ Product not found.
We couldn't find 9pd1wznbdxkz. It may be unavailable in your market."

I'm in the US and I'm making a bit of an assumption that this is your target audience. It would be helpful to know just how one can get "access to the app enabled".

![Image](https://github.com/user-attachments/assets/fa20e92e-5b0b-4266-bce0-dd958983111b)

---
*Reported from: https://documentation.ubuntu.com/wsl/en/latest/tutorials/getting-started-with-up4w/*

edibotopic:
Thanks @JonH130 

The Ubuntu Pro for WSL application is still under development and not yet generally available.
So that link only takes you to the download page if you have access to the application.

Access to the beta program is relatively closed at this point -- to maximise our ability to offer support to beta users -- but we are hoping to open up access soon.
We will provide clearer instructions as soon as that happens.

PR #1164, opened 2025-03-18
docs: Update Pro notice

edibotopic:
* Updates wording of Pro notice to explicitly state that application is in beta to better communicate its availability (https://github.com/canonical/ubuntu-pro-for-wsl/issues/1107)
* Adds custom Ubuntu styling to Pro blocks, which makes it more notable and branded

Example in Light mode:

image

Example in Dark mode:

image

**Note:** the design/web team is currently looking into the proposal for a custom Pro notice (https://github.com/canonical/vanilla-framework/issues/5491) but they have given their blessing to try my implementation out in the meantime, as it is preferential to using an info or note block, which are more generic.

UDENG-6388


CarlosNihelton (approved):
I like this brown very much!

edibotopic:
> I like this brown very much!

I think it's just grey, but now I'm questioning myself!

PR #1165, opened 2025-03-19
feat(docs): Explanation about binfmt and Ubuntu specifics

CarlosNihelton:
While https://github.com/ubuntu/WSL/issues/512 is a representative issue, I've seen similar complains in the upstream repository about binfmt_misc registration mediated by systemd not working. I've also seen a bunch of other complains in the past about WSL binary interoperability not working because of some surprising interaction with `systemd-binfmt.service`.

This explanation page aims to eliminate (or at least decrease the amount of) surprises by:
- Explaining power users why we chose to disable that service
- Giving them knowledge to do the opposite and deal with potential consequences

I have high hopes (based on conversations I already had with upstream) that we'll see strong improvements in this area still in the WSL 2.5.x release series.

---

UDENG-6366

### DIFF
--- a/docs/explanation/binfmt.md
+++ b/docs/explanation/binfmt.md
@@ -18,8 +18,9 @@ to pass that along to Windows (check `/proc/sys/fs/binfmt_misc/WSLInterop*` in a
 
 For reasons better explained in the rest of this page, we consider `systemd-binfmt.service` as a potential
 issue for most WSL users, thus that service **is intentionally disabled for Ubuntu on WSL**. Most users should
-not notice or care about that service, but those relying on emulators or interpreters may find this behaviour
-particularly annoying. If you are one of those users this page is for you.
+not notice or care about that service, as, by default, it does not affect the user's ability to run Windows
+executables. But those relying on emulators or interpreters may find this behaviour particularly annoying. If
+you are one of those users this page is for you.
 
 ## The systemd-binfmt.service and Windows binary interoperability
 
--- a/docs/explanation/binfmt.md
+++ b/docs/explanation/binfmt.md
@@ -25,16 +25,17 @@ you are one of those users this page is for you.
 ## The systemd-binfmt.service and Windows binary interoperability
 
 With systemd enabled, `systemd-binfmt.service` runs during boot, reads configuration files from specific
-directories and registers additional executable formats with the kernel. All registrations are removed when it
-stops. If that service is not aware of the WSL registration mechanism it can break Windows binary
-interoperability in different ways, such as:
+directories and registers additional executable formats with the kernel. All registrations are removed when
+the service stops. If that service is not aware of the WSL registration mechanism, Windows binary
+interoperability can break due to different factors, including:
 
-- Multiple running distro instances being affected when one stops. Since the WSL distro instances are effectively
-containers sharing the same kernel and the `binfmt_misc` file system interface is a shared mount point, when a
-distro instance stops and `systemd-binfmt.service` quits, it can break the registration for other instances.
+- `binfmt_misc` mountpoint being shared by multiple distro instances cause interoperability to be broken for
+multiple instances when one is shutdown. Since the WSL distro instances are effectively containers sharing the
+same kernel, when a distro instance stops and `systemd-binfmt.service` quits, it can break the registration
+for other instances.
 
 - Startup ordering. If WSL didn't order its own registration after that service, the service would break WSL
-interoperability.
+interoperability at boot time.
 
 - Service restart. Restarting `systemd-binfmt.service` implies unregistering and re-registering executable
 file formats. That can easily happen without the user being aware, such as when installing emulators or other
--- a/docs/explanation/binfmt.md
+++ b/docs/explanation/binfmt.md
@@ -43,17 +43,17 @@ packages that rely on `binfmt_misc`.
 
 ## Current limitations of binfmt registration protection implemented by WSL
 
-The scenarios above were reported in previous versions of WSL. Upstream implemented numerous improvements
-about that topic. As of version 2.5.1, WSL is capable of restoring its binfmt registration at startup and when
-that service is restarted. Yet, the solution is not complete enough to ensure most WSL users won't be affected
-by Windows interoperability breaks from time to time. For example, while `systemctl restart
-systemd-binfmt.service` by itself won't cause any issues, if that command runs after `systemctl daemon-reload`
-then the Windows executable format registration will be gone. It seems an edge case, but there is a
-non-trivial amount of combinations of packages that, when installed or uninstalled together, lead to such
-behaviour. Consider for instance, installing both `qemu-user-static` and `binfmt-support` (used in combination
-for example to allow ARM devices to execute x86_64 binaries). The latest needs to run `systemctl daemon-reload`
-in its post installation script and the first restarts the binfmt service. That happens exactly in the order
-that breaks Windows binary interoperability!
+The scenarios above were reported by users in previous versions of WSL. The WSL developers have since
+implemented numerous improvements. As of version 2.5.1, WSL is capable of restoring its binfmt registration at
+startup and when that service is restarted. Yet, the current solution does not guarantee that WSL users won't
+be affected by occasional breakage in Windows interoperability. For example, while `systemctl restart
+systemd-binfmt.service` by itself won't cause any problems, if that command runs after `systemctl
+daemon-reload` then the Windows executable format registration will be gone. This seems an edge case, but
+there is a non-trivial number of combinations of packages that, when installed or uninstalled together, lead
+to such behaviour. Consider for instance, installing both `qemu-user-static` and `binfmt-support` (used in
+combination for example to allow ARM devices to execute x86_64 binaries). The latter needs to run `systemctl
+daemon-reload` in its post installation script and the former restarts the binfmt service, which is exactly
+the order that breaks Windows binary interoperability!
 
 ```{warning} If you really want to understand why that happens
 
--- a/docs/explanation/binfmt.md
+++ b/docs/explanation/binfmt.md
@@ -2,16 +2,16 @@
 myst:
   html_meta:
     "description lang=en":
-      "Explains how systemd-binfmt.service affects the Ubuntu WSL experience and what can be done about that."
+      "Explains why systemd-binfmt.service affects the WSL experience, our approach to the service on the Ubuntu distro and how users can configure the behaviour."
 ---
 
-# Ubuntu on WSL and `binfmt_misc`
+# Support for miscellaneous binary formats (binfmt_misc) with Ubuntu on WSL
 
-One key feature of the WSL experience is the binary interoperability, that is, the ability to seamlessly run
-Windows binaries from inside WSL and vice-versa. Linux can run Windows binaries thanks to `binfmt_misc`, a
-capability offered by the Linux kernel via which arbitrary executable file formats can be recognised and
-passed to certain user space applications, such as interpreters, emulators and virtual machines, that know how
-to execute such format. The executable formats are registered via a file interface, usually located at
+A key feature of  WSL is binary interoperability, which allows
+Windows binaries to be run inside WSL and vice-versa. Linux can run Windows binaries thanks to `binfmt_misc`, a
+capability offered by the Linux kernel. With `binfmt_misc`, arbitrary executable file formats can be recognised and
+passed to certain user space applications, such as interpreters, emulators and virtual machines, which can then
+execute that specific format. The executable formats are registered via a file interface, usually located at
 `/proc/sys/fs/binfmt_misc/register` or using wrappers such as those offered by `binfmt-support` or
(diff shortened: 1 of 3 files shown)


docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+about that topic. As of version 2.5.1, WSL is capable of restoring its binfmt registration at startup and when
+that service is restarted. Yet, the solution is not complete enough to ensure most WSL users won't be affected
+by Windows interoperability breaks from time to time. For example, while `systemctl restart
+systemd-binfmt.service` by itself won't cause any issues, if that command runs after `systemctl daemon-reload`
+then the Windows executable format registration will be gone. It seems an edge case, but there is a
+non-trivial amount of combinations of packages that, when installed or uninstalled together, lead to such
+behaviour. Consider for instance, installing both `qemu-user-static` and `binfmt-support` (used in combination
+for example to allow ARM devices to execute x86_64 binaries). The latest needs to run `systemctl daemon-reload`
+in its post installation script and the first restarts the binfmt service. That happens exactly in the order
+that breaks Windows binary interoperability!
```
edibotopic, line 52:
```suggestion
The scenarios above were reported by users in previous versions of WSL. The WSL developers have since implemented numerous improvements
. As of version 2.5.1, WSL is capable of restoring its binfmt registration at startup and when
the service is restarted. Yet, the current solution does not guarantee that WSL users won't be affected
by occasional breakaged in Windows interoperability. For example, while `systemctl restart
systemd-binfmt.service` by itself won't cause any problems, if that command runs after `systemctl daemon-reload`
then the Windows executable format registration will be gone. This seems like an edge case, but there is a
non-trivial number of packages that, when installed or uninstalled together, lead to such
behaviour. Consider `qemu-user-static` and `binfmt-support`, which are often installed together because they are used in combination;
for example, this enables ARM devices to execute x86_64 binaries. The latest needs to run `systemctl daemon-reload`
in its post-installation script and the first restarts the binfmt service, which is exactly the order
that breaks Windows binary interoperability!
```

Query re: line below...

> The **latest** needs to run `systemctl daemon-reload` in its post-installation script and the **first** restarts the binfmt service

What does "latest" and "first" refer to here, it's a little unclear. Do you mean "latter" (binfmt-support) and "former" (qemu-user-static)? If so, I would use latter/former or just spell them out again.

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+```{warning} If you really want to understand why that happens
+
+**Doing the tests below can make binary interoperability broken until the WSL instance is restarted!**
```
edibotopic, line 56:
```suggestion
**Doing the tests below can break binary interoperability until the WSL instance is restarted.**
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+WSL writes the override file `/run/systemd/generator.early/wsl-binfmt.service` that runs some commands to
+restore the WSL interoperability registration when the systemd-binfmt.service (re)starts.
+
+Try running the following command: `sudo systemctl daemon-reload`
+Then try finding that file again. It's gone.
```
edibotopic, line 62:
```suggestion
Try running the following command: `sudo systemctl daemon-reload`.
Then try finding that file again. It's gone.
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+If `systemd-binfmt.service` was allowed to run at this point, Windows binary interoperability would break.
+```
+
+
+## The Ubuntu approach
+
+While there are other ways to solve this problem, we assumed most users wouldn't notice that service being
+disabled, thus Ubuntu WSL images are published with a file that makes systemd disable that service on WSL,
+which is:
```
edibotopic, line 77:
```suggestion
While there are other ways to solve this problem, we assumed that most users wouldn't notice if the service was
disabled. Ubuntu WSL images are therefore published with a file that makes systemd disable that service on WSL,
which is:
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+`/usr/lib/systemd/system/systemd-binfmt.service.d/wsl.conf`
+
+```ini
+[Unit]
+ConditionVirtualization=!wsl
+```
+
+Advanced users that need emulators and binfmt support managed by systemd more than the Windows binary
+interoperability or that can tolerate the need to restart WSL when the interoperability breaks are encouraged
+to remove that file.
```
edibotopic, line 88:
```suggestion
Users that need emulators and binfmt support managed by systemd more than the Windows binary
interoperability or that can tolerate the need to restart WSL when the interoperability breaks are encouraged
to remove that file.
```

I would avoid classifying users as "advanced" or otherwise, but just point out the functionality that they want.

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+```bash
+sudo rm /usr/lib/systemd/system/systemd-binfmt.service.d/wsl.conf
+sudo systemctl daemon-reload
+```
+
+Then, manually restart that service.
```
edibotopic, line 95:
```suggestion
Then, manually restart that service:
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+```bash
+sudo systemctl restart systemd-binfmt.service
+```
+
+From now on, foreign executable file format registration mediated by systemd will just work as you'd expect.
```
edibotopic, line 101:
```suggestion
After running these commands, foreign executable file format registration mediated by systemd will work.
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+```ini
+[Unit]
+ConditionVirtualization=!wsl
+```
+
+Advanced users that need emulators and binfmt support managed by systemd more than the Windows binary
+interoperability or that can tolerate the need to restart WSL when the interoperability breaks are encouraged
+to remove that file.
+
+```bash
```
edibotopic, line 90: For your consideration: this section is a good candidate for self-contained how-to guide content.

Consider a user who doesn't care about the explanation but just wants to get their emulator/interpreter working. In that case, having a concise how-to guide (cross-linked with this wonderful explanation, of course) might be preferable.

A further point. I consider there -- at a minimum -- to be three potential audiences for this content:

1. Someone who knows that they need support for interpreters/emulators and wants to configure Ubuntu on WSL to support their use-case (the "advanced" user you mention)
2. Someone who isn't affected by the binfmt configuration and is unlikely to ever need to change the default (let's say the "beginner" user)
3. Someone who isn't affected by the binfmt configuration but is curious about how it might help them achieve something (the "curious" user)

In a dedicated how-to guide, there would be scope to include a demonstrative example, with a specific emulator or interpreter. I think that this would especially benefit user of type (3), who might not know about binfmt or how it affects them but could be interested in trying it out.

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+From now on, foreign executable file format registration mediated by systemd will just work as you'd expect.
+
+## Closing out
+
+More improvements are expected in the WSL 2.5.x release series, so we're positive about removing the
+override that disables the `systemd-binfmt.service` on Ubuntu on WSL soon. That transition should be
+transparent to most users. In the meanwhile, power users that need `systemd-binfmt.service` should refer to
+this page.
```
edibotopic, line 108:
```suggestion
More improvements are expected in the WSL 2.5.x release series, so we're positive that we will soon be able to remove the
override that disables the `systemd-binfmt.service` on Ubuntu on WSL. That transition should be
transparent to most users. In the meantime, users that need `systemd-binfmt.service` can follow the configuration steps outlined in
this page.
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+`/proc/sys/fs/binfmt_misc/register` or using wrappers such as those offered by `binfmt-support` or
+`systemd-binfmt.service`. WSL registers Windows binaries to be passed to the `/init` program, which knows how
+to pass that along to Windows (check `/proc/sys/fs/binfmt_misc/WSLInterop*` in a WSL distro instance).
+
+For reasons better explained in the rest of this page, we consider `systemd-binfmt.service` as a potential
+issue for most WSL users, thus that service **is intentionally disabled for Ubuntu on WSL**. Most users should
+not notice or care about that service, but those relying on emulators or interpreters may find this behaviour
+particularly annoying. If you are one of those users this page is for you.
+
+## systemd-binfmt.service
```
edibotopic, line 24:
```suggestion
## The systemd-binfmt.service and Windows binary interoperability
```

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+distro instance stops and `systemd-binfmt.service` quits it can broken the registration for other instances.
+
+- Startup ordering. If WSL didn't order its own registration after that service, the service would break WSL
+interoperability.
+
+- Service restart. Restarting `systemd-binfmt.service` implies unregistering and re-registering executable
+file formats. That can easily happen without the user awareness, such as when installing emulators or other
+packages that rely on `binfmt_misc`.
+
+The scenarios above were reported in previous versions of WSL. Upstream implemented numerous improvements
```
edibotopic, line 42: Maybe give this a sub-heading?

The reason I say this is that this is a relatively dense, technical explanation and could benefit from more signposts for the reader.

Perhaps: `## Current limitations when restoring binfmt registration on WSL` ???

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "Explains how systemd-binfmt.service affects the Ubuntu WSL experience and what can be done about that."
```
edibotopic, line 5:
```suggestion
      "Explains why systemd-binfmt.service affects the WSL experience, our approach to the service on the Ubuntu distro and how users can configure the behaviour."
```

edibotopic (changes requested):
Excellent work @CarlosNihelton -- this was not an area I was familiar with and its was interesting to learn about it (perfect for an Explanation page).

The page explains the problem well and provides a valid justification for our choices.

I have added various small tweaks, mostly for clarity. There are some minor queries too, most notably a proposal for possibly spinning off a how-to guide, but we can pick that up another time if needs be.

Great job!!!

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+passed to certain user space applications, such as interpreters, emulators and virtual machines, that know how
+to execute such format. The executable formats are registered via a file interface, usually located at
+`/proc/sys/fs/binfmt_misc/register` or using wrappers such as those offered by `binfmt-support` or
+`systemd-binfmt.service`. WSL registers Windows binaries to be passed to the `/init` program, which knows how
+to pass that along to Windows (check `/proc/sys/fs/binfmt_misc/WSLInterop*` in a WSL distro instance).
+
+For reasons better explained in the rest of this page, we consider `systemd-binfmt.service` as a potential
+issue for most WSL users, thus that service **is intentionally disabled for Ubuntu on WSL**. Most users should
+not notice or care about that service, but those relying on emulators or interpreters may find this behaviour
+particularly annoying. If you are one of those users this page is for you.
```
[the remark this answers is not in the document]
CarlosNihelton: I think it's a good measure to emphasise it.

docs/explanation/binfmt.md
```diff
@@ -0,0 +1,113 @@
...
+
+## systemd-binfmt.service
+
+With systemd enabled, `systemd-binfmt.service` runs during boot, reads configuration files from specific
+directories and registers additional executable formats with the kernel. All registrations are removed when it
+stops. If that service is not aware of the WSL registration mechanism it can break Windows binary
+interoperability in different ways, such as:
```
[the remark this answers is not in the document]
CarlosNihelton: The user visible effects will be the same: breaking interop. Let me reuse your suggestion in a hopefully more streamlined way.

docs/explanation/binfmt.md
CarlosNihelton: Since those are measures taken by WSL rather than by the user, I'd rather name the section as:

`## Current limitations of binfmt registration protections implemented by WSL`

docs/explanation/binfmt.md
CarlosNihelton: Yeah, it should be latter/former. Thanks for flagging it.

docs/explanation/binfmt.md
CarlosNihelton: I have high hopes that the next releases of WSL will feature a more complete solution to this problem.
When that happens, I'll come to this document and flip it upside-down, turning it into a guide for those not yet running on the latest WSL versions. Having it self-contained is not only better for maintenance purposes but also for expressing its limited relevance.

edibotopic (approved):
LGTM: no further comments!

PR #1166, opened 2025-03-19
fix(windows-agent): Cleanup cloud init files for invalid distros

ashuntu:
If a distro is invalidated (the user deletes it, for example), its cloud-init files may stick around without being cleaned up. This could lead to unwanted behavior if a new distro with the same name is created.

Also, if a Landscape install falls, the corresponding cloud-init file is not cleaned up. This fixes that as well.

---

UDENG-6343

### DIFF
--- a/windows-agent/internal/distros/database/database_test.go
+++ b/windows-agent/internal/distros/database/database_test.go
@@ -501,7 +501,7 @@ func TestDatabaseCleanup(t *testing.T) {
 			databaseFromTemplate(t, dbDir, distros...)
 
 			cleanupCalled := false
-			var cleanupFunc func(string) = nil
+			var cleanupFunc func(string)
 			if tc.cleanupFunc {
 				cleanupFunc = func(d string) {
 					require.True(t, strings.EqualFold(tc.markDistroUnreachable, d), "Unexpected cleaned up distro name")
--- a/windows-agent/internal/distros/database/database_test.go
+++ b/windows-agent/internal/distros/database/database_test.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"testing"
 	"text/template"
 	"time"
@@ -500,12 +501,12 @@ func TestDatabaseCleanup(t *testing.T) {
 
 			databaseFromTemplate(t, dbDir, distros...)
 
-			cleanupCalled := false
+			var cleanupCalled atomic.Bool
 			var cleanupFunc func(string)
 			if tc.cleanupFunc {
 				cleanupFunc = func(d string) {
 					require.True(t, strings.EqualFold(tc.markDistroUnreachable, d), "Unexpected cleaned up distro name")
-					cleanupCalled = true
+					cleanupCalled.Store(true)
 				}
 			}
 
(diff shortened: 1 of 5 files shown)


windows-agent/internal/distros/database/database_test.go
```diff
@@ -497,7 +501,16 @@ func TestDatabaseCleanup(t *testing.T) {
...
 
-			db, err := database.New(ctx, dbDir)
+			var cleanupCalled atomic.Bool
+			var cleanupFunc func(string)
+			if tc.cleanupFunc {
+				cleanupFunc = func(d string) {
+					require.True(t, strings.EqualFold(tc.markDistroUnreachable, d), "Unexpected cleaned up distro name")
+					cleanupCalled.Store(true)
+				}
+			}
```
CarlosNihelton, line 511: Using require in this case is discouraged because it may panic in a goroutine other than the test main (this callback will be executed in another goroutine for sure).

```suggestion
			if tc.cleanupFunc {
				cleanupFunc = func(d string) {
					cleanupCalled.Store(strings.EqualFold(tc.markDistroUnreachable, d))
				}
			}
```

windows-agent/internal/proservices/landscape/executor.go
```diff
@@ -207,6 +207,11 @@ func (e executor) install(ctx context.Context, cmd *landscapeapi.Command_Install
 		if err != nil {
 			log.Warningf(ctx, "Landscape Install: failed to clean up %q after failed Install: %v", distro.Name(), err)
 		}
+
+		err = e.cloudInit().RemoveDistroData(distro.Name())
+		if err != nil {
+			log.Warningf(ctx, "Landscape install: failed to clean up %q cloud-init data after failed install: %v", distro.Name(), err)
+		}
```
CarlosNihelton, line 214: Since we don't really handle the errors, but log them in almost identical ways, I'd reorganize this code a little bit differently to get rid of the coverage warning:

```go
// Avoid error states by cleaning up on error
err := errors.Join(distro.Uninstall(ctx), e.cloudInit().RemoveDistroData(distro.Name()))
if err != nil {
			log.Warningf(ctx, "Landscape Install: failed to clean up %q after failed Install: %v", distro.Name(), err)
}
```

If you think it's too terse you can do

```go
e0 := distro.Uninstall(ctx)
e1 := e.cloudInit().RemoveDistroData(distro.Name()))
err := errors.Join(e0, e1)
if err != nil {
....
```

Either way communicates clearly that those actions don't have any dependency relationship between each other and we don't do much with their errors.

windows-agent/internal/proservices/proservices.go
```diff
@@ -85,7 +85,7 @@ func New(ctx context.Context, publicDir, privateDir string, args ...Option) (s M
 		return s, err
 	}
 
-	db, err := database.New(ctx, privateDir)
+	db, err := database.New(ctx, privateDir, func(d string) { _ = cloudInit.RemoveDistroData(d) })
```
CarlosNihelton, line 88: It may make your tests just a little bit harder, but I think we should log the error here. Imagine later we have a problem with an instance being surprisingly initialised from cloud-init files seating in the folder for no known reason. This log line would tell when the clean-up failed, leading to that situation. We can pass to the log call the same context used to construct the database.

CarlosNihelton (changes requested):
Good job so far. I have a few issues to report, though.

windows-agent/internal/proservices/proservices.go
ashuntu: I had misread `RemoveDistroData` and thought it was logging already, but it's only a decorate :), so good catch.

CarlosNihelton (approved):
I realize the non covered lines in this patch are hard to cover indeed because the methods to break `cloudinit.RemoveDistroData()` also break `proservices.New()` so we'd fail early.

Good job!

PR #1171, opened 2025-03-20
docs: add a security overview

edibotopic:
Adds a new security overview page to the Explanation section of the docs.

This page outlines security consideration when using Ubuntu on WSL, includes links to relevant security information found elsewhere, and briefly mentions some tips for configuring features relevant to security.

In future, this could possibly be broken down into individual pages if some of the content was to be developed further, but for now a single page is sufficient.

UDENG-6389

### DIFF
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -68,7 +68,7 @@ necessary.
 Once logged into a Windows host machine, the user can create WSL
 instances without elevated privileges.
 
-When first opening an Ubuntu on WSL terminal with `wsl.exe -d ubuntu`, the user is
+When first opening an Ubuntu on WSL terminal with `wsl.exe -d <ubuntu distro>`, the user is
 prompted for a username and password to create the default user account on Ubuntu.
 
 Even if a password is set, it can be changed by the root user; however, the
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -50,22 +50,23 @@ wsl --install --from-file ubuntu-<version number>-wsl.amd64.wsl
 
 ## Login
 
-### Host
+### Windows host
 
 Any WSL instance is only as secure as its Windows host.
 
 The Windows user should be protected by a strong password, which will — by
 extension — help secure instances of Ubuntu on WSL on the host machine.
 
-Store your passwords securely and only share them with administrators.
+Store your passwords securely and only share them with administrators when/if
+necessary.
 
-### Instance
+### WSL instance
 
 Once logged into a Windows host machine, the user can create WSL
 instances without elevated privileges.
 
 When first opening an Ubuntu on WSL terminal with `wsl.exe -d ubuntu`, the user is
-prompted for a username and password.
+prompted for a username and password to create the default user account on Ubuntu.
 
 Even if a password is set, it can be changed by the root user; however, the
 permissions of the Windows user supersede that of the Ubuntu user.
@@ -75,12 +76,12 @@ permissions of the Windows user supersede that of the Ubuntu user.
 Access to a WSL instance as the root user is possible:
 
 ```text
-wsl -d Ubuntu -u root
+wsl -d <ubuntu distro> -u root
 ```
 
 After accessing an instance as a regular (non-root) user, a password is still expected for
-commands requiring `sudo` within the instance.
-The standard Linux user account controls still apply here.
+commands requiring `sudo` within the instance:
+the standard Linux user account controls apply.
 
 Interacting with an instance using root access has no effect on the permissions
 of the Windows' user, which continues to take precedence.
@@ -137,7 +138,7 @@ apply when operating from within a WSL instance.
 The instance is therefore as secure as any arbitrary program running on the user
 account of the Windows' host.
 
-If you are concerned about the security implications of interoperability, it [can
+If you remain concerned about the security implications of interoperability, it [can
 be disabled](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#interop-settings) in `/etc/wsl.conf`:
 
 ```ini
@@ -149,7 +150,7 @@ enabled=false
 Interoperability is necessary for certain processes, including provisioning
 with cloud-init.
 
-[One approach](exp::automate-hardening) is to first provision an instance and
+[One approach](exp::disable-interoperability) is to first provision an instance and
 then subsequently disable the feature.
 ```
 
@@ -160,7 +161,8 @@ distributions. For Ubuntu on WSL, the Pro client is pre-installed.
 
 ### Manual Pro-attachment
 
-To manually attach a Pro subscription to a new instance, log in and run:
+To manually attach a Pro subscription to a new instance, run this
+command from inside the instance:
 
 ```text
 sudo pro attach
@@ -249,7 +251,7 @@ To set it:
 Set-ItemProperty -Path "HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss" -Name DefaultVersion -Value 2
 ```
 
-Intune also supports policies for WSL, which include toggling the availability of WSL1:
+Intune also supports policies for WSL, which include toggling the availability of WSL1 on client machines:
 
 > [Intune configuration options for WSL](https://learn.microsoft.com/en-us/windows/wsl/intune?source=recommendations)
 
@@ -257,15 +259,15 @@ Intune also supports policies for WSL, which include toggling the availability o
 
 ### Configuring WSL features
 
-WSL features can be toggled if they present a security concern.
+WSL features can be controlled if they present a security concern.
 
 For example, root login can be disabled, WSL1 availability toggled and network
 access configured.
 
 There are various options to configure WSL instances, including:
 
-* The `.wslconfig` file for [global settings](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#wslconfig)
-* [WSL policies with Intune](https://learn.microsoft.com/en-us/windows/wsl/intune?source=recommendations)
+* The `.wslconfig` file can be edited to configure [global settings](https://learn.microsoft.com/en-us/windows/wsl/wsl-config#wslconfig) for instances
+* [WSL policies for Intune](https://learn.microsoft.com/en-us/windows/wsl/intune?source=recommendations) enable remote management of WSL features
 * Registry entries for features like [WSL1 availability](exp::wsl1-incompatibility) can be changed in the registry editor or with PowerShell scripts 
 
 (exp::automate-hardening)=
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -105,7 +105,7 @@ Ubuntu on WSL also supports the installation of `snaps`, which are a more secure
 
 > [Read more about third-party packages in the Ubuntu Server documentation](https://documentation.ubuntu.com/server/explanation/software/third-party-repository-usage/)
 
-If an instance is running, security updates are installated automatically.
+If an instance is running, security updates are installed automatically.
 This is because `unattended-upgrades` are enabled by default.
 
 ### AppArmor
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -19,6 +19,9 @@ This page assumes a WSL version of 2.4.10 or later.
 Always use a [supported LTS version](reference::distros) of Ubuntu on WSL to
 ensure that you receive regular updates and bug fixes.
 
+> For our latest installation instructions, read the [install Ubuntu on
+WSL](howto::install-ubuntu-wsl) guide.
+
 ### Verifying the download (automatic)
 
 When installing an Ubuntu image directly from the terminal using `wsl --install
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -1,0 +1,330 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "TODO — update me"
+---
+
+# Security overview for Ubuntu on WSL
+
+```{note}
+For simplicity, we refer on this page to the default "Ubuntu" distro and not
+any numbered versions when discussing the installation and creation of
+instances.
+```
+
+## Download and installation
+
+When an Ubuntu image is installed using `wsl --install ubuntu`, the SHA-256
+checksum is automatically verified to ensure that it is secure.
+
+After downloading an image as a tarball with `.wsl` extension from an online
+archive, it can then be installed by double-clicking or running:
+
+```text
+wsl --install --from-file ubuntu.tar.wsl
+```
+
+If installing a custom tarball, it is recommended that you manually verify the
+checksums file, which is included with images on the
+[releases.ubuntu.com](https://releases.ubuntu.com) page.
+
+> [Read more about verifying an Ubuntu download](https://ubuntu.com/tutorials/how-to-verify-ubuntu#1-overview)
+
+
+## Login
+
+### Host
+
+Any WSL instance is only as secure as its Windows host.
+
+The Windows user should be protected by a strong password, which will — by
+extension — help secure instances of Ubuntu on WSL on the host machine.
+
+Store your passwords securely and only share them with administrators.
+
+### Instance
+
+Once logged into a Windows host machine, the user can create WSL
+instances without elevated privileges.
+
+When first opening an Ubuntu on WSL terminal with `ubuntu.exe`, the user is
+prompted for a username and password.
+
+Even if a password is set, it can be changed by the root user; however, the
+permissions of the Windows user supersede that of the Ubuntu user.
+
+### Root access
+
+Access to a WSL instance as the root user is possible:
+
+```text
+wsl -d Ubuntu -u root
+```
+
+After accessing an instance as a root user, a password is still expected for
+commands requiring `sudo` within the instance.
+
+Interacting with an instance using root access has no effect on the permissions
+of the Windows' user, which still take precedence.
+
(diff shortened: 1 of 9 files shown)


docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+## Login
+
+### Host
+
+Any WSL instance is only as secure as its Windows host.
+
+The Windows user should be protected by a strong password, which will — by
+extension — help secure instances of Ubuntu on WSL on the host machine.
+
+Store your passwords securely and only share them with administrators.
```
CarlosNihelton, line 44: For your consideration:


```suggestion
Store your passwords securely and only share them with administrators when/if necessary.
```

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+
+### Instance
+
+Once logged into a Windows host machine, the user can create WSL
+instances without elevated privileges.
+
+When first opening an Ubuntu on WSL terminal with `ubuntu.exe`, the user is
```
CarlosNihelton, line 51: If we assume WSL 2.4.10 or later as I suggested earlier, than there won't be a `.exe` anymore, thus `wsl.exe -d Ubuntu` will be the command line to start the distro instance.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+prompted for a username and password.
```
CarlosNihelton, line 52:
```suggestion
prompted for a username and password to create the default user account on Ubuntu.
```

I wanted to emphasise that we're creating a user account at that step instead of authenticating with an existing one. You may have a better wording for that (or deem it unnecessary).

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+### Root access
+
+Access to a WSL instance as the root user is possible:
+
+```text
+wsl -d Ubuntu -u root
+```
+
+After accessing an instance as a root user, a password is still expected for
+commands requiring `sudo` within the instance.
```
CarlosNihelton, line 66: `root` doesn't need `sudo`, it's already the superuser. You may want to rephrase that like:

```suggestion
After accessing an instance as a regular (non-root) user, a password is still expected for
commands requiring `sudo` within the instance. The standard Linux user account 
controls still apply here.
```

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+
+Interacting with an instance using root access has no effect on the permissions
+of the Windows' user, which still take precedence.
```
CarlosNihelton, line 69: We may want to emphasise it further:

> Running Windows binaries as root inside WSL won't make them elevated on the Windows side, they still run with the Windows user permissions only.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+```
+
+When inside an instance, check that AppArmor is loaded and active with:
+
+```text
+aa-status
+```
+
+> [To learn more about how AppArmor contributes to Snap security, read the Snapcraft documentation](https://snapcraft.io/docs/security-policies)
+
```
CarlosNihelton, line 109: While it's true that it will work and AppArmor will be enabled with the latest versions of WSL, I didn't yet evaluated how complete is the AppArmor support with the latest Linux kernel shipped by Microsoft. I know that snapd requires patches to the kernel about AppArmor that are not yet merged upstream, but I'm not fully aware of where we are and what are those patches for.

That being said, I'm not entirely sure we should document this option now. Maybe we should add this section next cycle if we commit to do a more in-depth investigation on AppArmor support.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+> [Read more about third-party packages in the Ubuntu Server documentation](https://documentation.ubuntu.com/server/explanation/software/third-party-repository-usage/)
+
+### AppArmor
+
+AppArmor is a Linux Security Module implementation that controls the
+capabilities and permissions of applications.
+
+By default, AppArmor is installed in Ubuntu on WSL but not enabled.
+
+To enable AppArmor, add the following lines to `.wslconfig`:
```
CarlosNihelton, line 95: Less knowledgeable users will be wondering where this `.wslconfig` file is located. I like to refer to it as `%USERPROFILE%\.wslconfig` to make it look like Windows stuff.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+  - path: /etc/ssh/sshd_config
+    content: |
+      HostKey /etc/ssh/ssh_host_rsa_key
+      MaxAuthTries 3
+      PermitRootLogin no
+      PermitEmptyPasswords no
+      AllowUsers u
+runcmd:
+  - echo "[interop]" | sudo tee -a /etc/wsl.conf
+  - echo "enabled = false" | sudo tee -a /etc/wsl.conf
```
CarlosNihelton, line 290: Since we're here, you may want to use the runcmd feature to replace the root line in `/etc/passwd` to make  `wsl -d Ubuntu -u root` effectively fail to log in.

Something like:

```
- sed -i 's/^root.*$/root:x:0:0:root:\/root:\/usr\/sbin\/nologin/' /etc/passwd
```
(replace the complete line that starts with `root` with the contents `root:x:0:0:root:/root:/usr/sbin/nologin`. The backslash scaping is a bit noisy, thus the explanation)

CarlosNihelton: The ofc add a bullet point below about it.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+* Registry entries for features like [WSL1 availability](exp::wsl1-incompatibility) can be changed in the registry editor or with PowerShell scripts 
+
+(exp::automate-hardening)=
+### Automate hardening
+
+Provisioning of WSL instances can be automated with cloud-init.
+
+> [Read about automatic setup of Ubuntu on WSL with cloud-init](howto::cloud-init)
+
+Cloud-init can be used to secure your instances before first login:
```
CarlosNihelton, line 259:
```suggestion
Cloud-init can be used to initialise your instances in a more secure way (depending on your needs) before first login:
```

I'm afraid users would just copy and paste it without really considering what pieces of that are really necessary for their use cases. Also, I'd like to avoid the risk users thinking cloud-init per se is a security tool.

I considered splitting the entire cloud-config below in small snippets, with an explainer paragraph before, like:

> If you need to disable root login, then add this line to your runcmd module

then a snippet would follow that line showing what to do.

I reckon it'd make the guide a bit harder to understand. My main reasoning is just not insinuating to the user that this is the holy grail recipe for security of WSL instances.

CarlosNihelton (commented):
It's a very thoughtful guide. Congrats! I have comments, no real hard rules, but things to consider.

docs/explanation/security-overview.md
edibotopic: Much better.

docs/explanation/security-overview.md
edibotopic: Yes. Let's keep this until next cycle. That's a far and correct assessment.

In the meantime, however, should we perhaps indicate that AppArmor is currently not enabled by default?

My reason to include that would be to offset the mistaken assumption that the security features on Snaps in a WSL instance are as one might expect for some other platforms.

docs/explanation/security-overview.md
edibotopic: Nice suggestion.

docs/explanation/security-overview.md
edibotopic: Good idea!

docs/explanation/security-overview.md
edibotopic: You both mentioned this in our meeting: thanks for reminding me!

I think you were right before and now. It's better to include as discreet snippets for the reasons you suggest.

edibotopic:
Thanks @CarlosNihelton -- your comments are fantastic and will really strengthen this guide. 

I will apply your changes and improvements next week.

docs/explanation/security-overview.md
CarlosNihelton: Indeed, snap users could be surprised by that. 

```
### AppArmor

AppArmor is a Linux Security Module implementation that controls the
capabilities and permissions of applications.

By default, AppArmor is installed in Ubuntu on WSL but not enabled, because
it requires certain features and patches not yet available in the WSL kernel.

Because of that, snaps cannot be fully confined on WSL.
```

Or something along those lines.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+any numbered versions when discussing the installation and creation of
+instances.
+```
+
+## Download and installation
+
+When an Ubuntu image is installed using `wsl --install ubuntu`, the SHA-256
+checksum is automatically verified to ensure that it is secure.
+
+After downloading an image as a tarball with `.wsl` extension from an online
```
didrocks, line 21: Just some notes visually: I don’t feel this is conveying those are separate methods of installation, but that could be just me!

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+
+```text
+wsl --install --from-file ubuntu.tar.wsl
+```
+
+If installing a custom tarball, it is recommended that you manually verify the
+checksums file, which is included with images on the
+[releases.ubuntu.com](https://releases.ubuntu.com) page.
+
+> [Read more about verifying an Ubuntu download](https://ubuntu.com/tutorials/how-to-verify-ubuntu#1-overview)
```
didrocks, line 32: Remember that to do the verification, we are on Windows.

https://ubuntu.com/tutorials/how-to-verify-ubuntu#2-necessary-software has a Windows section (using the old name of WSL btw "Bash on Windows", which refers to the doc for… "Ubuntu on WSL", but you didn’t install your distro yet! We have some kind of a chicken and egg issue here, right? :)

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+commands requiring `sudo` within the instance.
+
+Interacting with an instance using root access has no effect on the permissions
+of the Windows' user, which still take precedence.
+
+## Package management
+
+### Updates and upgrades
+
+As with any distribution, packages should be routinely updated and upgraded:
```
didrocks, line 75: @CarlosNihelton: `unattended-uprades` is enabled on WSL, right? If so, security updates are installed automatically (as long as the instance is running) and I think that should be mentioned.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+By default, AppArmor is installed in Ubuntu on WSL but not enabled.
+
+To enable AppArmor, add the following lines to `.wslconfig`:
+
+```ini
+[wsl2]
+kernelCommandLine=apparmor=1 security=apparmor lsm=apparmor
+```
+
+When inside an instance, check that AppArmor is loaded and active with:
```
didrocks, line 102: Also, as this is a kernel command line, it should be mentioned that:
* This impacts all WSL distributions (even non ubuntu ones)
* For this to take effect the first time, you need to shutdown WSL to ensure there is no running instances before the parameter takes effect

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+    [user]
+    default=u
+  - path: /etc/ssh/sshd_config
+    content: |
+      HostKey /etc/ssh/ssh_host_rsa_key
+      MaxAuthTries 3
+      PermitRootLogin no
+      PermitEmptyPasswords no
+      AllowUsers u
+runcmd:
```
didrocks, line 288: nitpick: you separate the previous sections with an empty line before, you should do the same here.

didrocks (commented):
Excellent work! Very detailed and to the point!

In addition to Carlo's feedback, I have some small comments too that I think we should precise/fix. Otherwise, perfect!

docs/explanation/security-overview.md
edibotopic: You're right!

docs/explanation/security-overview.md
edibotopic: I just did a quick look online and I think this can be done in PowerShell using `Get-FileHash`:

https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/get-filehash?view=powershell-7.5

If that seems sensible (and if it works!) then we could mention that briefly here.

docs/explanation/security-overview.md
edibotopic: Excellent points -- thank you!

docs/explanation/security-overview.md
didrocks: Let’s get some confirmation from @CarlosNihelton but yeah, sounds better to me.

CarlosNihelton: Yeap, Get-FileHash works on both powershell.exe and pwsh.exe just fine and it's even recommended by Microsoft in the [build a custom distro](https://learn.microsoft.com/en-us/windows/wsl/build-custom-distro#test-the-distribution-locally) page.

docs/explanation/security-overview.md
CarlosNihelton: Yeap, enabled by default.

docs/explanation/security-overview.md
edibotopic: Perfect! I will add some info on that.

docs/explanation/security-overview.md
edibotopic: Thanks both.

docs/explanation/security-overview.md
```diff
@@ -0,0 +1,330 @@
...
+After downloading an image as a tarball with `.wsl` extension from an online
+archive, it can then be installed by double-clicking or running:
+
+```text
+wsl --install --from-file ubuntu.tar.wsl
+```
+
+If installing a custom tarball, it is recommended that you manually verify the
+checksums file, which is included with images on the
+[releases.ubuntu.com](https://releases.ubuntu.com) page.
```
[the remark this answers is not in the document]
edibotopic: resolved in: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/2940def15704d215f774d4c83ed144d2b48a0c5d

docs/explanation/security-overview.md
edibotopic: applied in: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/44e1a0a68554bdbb53a0ee1bc9aa47d68012ddf9

docs/explanation/security-overview.md
edibotopic: `.exe` has been removed in this section: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/32964dd1a1cf36638aa3ad558601f3f8ddcb9907

docs/explanation/security-overview.md
edibotopic: Applied here: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/44e1a0a68554bdbb53a0ee1bc9aa47d68012ddf9

docs/explanation/security-overview.md
edibotopic: AppArmor suggestion now provides info on AppArmor limitations in WSL but does not provide config instructions: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/273c6f383bb795f953ca860cd6991aff627eab41

docs/explanation/security-overview.md
edibotopic: Line no longer included here since changes in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/273c6f383bb795f953ca860cd6991aff627eab41

`.wslconfig` is mentioned later but only when user is linked to MS docs re: same.

docs/explanation/security-overview.md
edibotopic: I made a series of changes in this section, including trying to make these parts more distinct: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/6311b25deb42dc6f099c14f7dedbcf5502ce61df

docs/explanation/security-overview.md
edibotopic: I tested that PS command with a WSL image and it works perfectly: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/6311b25deb42dc6f099c14f7dedbcf5502ce61df

docs/explanation/security-overview.md
edibotopic: Now mentioned: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/7572b2a6cb0dad63ccf07ba5f8899af4d28a7395

docs/explanation/security-overview.md
edibotopic: AppArmor section consolidated in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/273c6f383bb795f953ca860cd6991aff627eab41 so this no longer applies, but will be noted for future development of AppArmor section.

docs/explanation/security-overview.md
edibotopic: Cloud-init section is now divided into snippets, as you and Carlos originally suggested: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1171/commits/8fa2e2c2483e2e4f9d52ddb5ac9ca97bf37efd9f

didrocks (approved):
Thanks for addressing all my concerns and comments!

PR #1182, opened 2025-03-31
deps: Upgrade protobuf

CarlosNihelton:


ashuntu (approved):
Always annoying when such a small change waterfalls into a bunch of side effects :). This looks good to me!

PR #1184, opened 2025-03-31
fix(wsl-pro-service): Less logging

CarlosNihelton:
The current implementation of wsl-pro-service's `daemon.Serve()` keeps trying to connect to the Windows agent indefinitely, with exponential back-off starting at 1 second and saturating at 60 min with a power of 2 growth factor. That means it will roughly take 2 minutes to saturate the back-off wait time and then keep retrying at every minute. Since most WSL users don't have the Windows agent running, that means one log line being added to the journal by default every minute. That's not a good default.

With this PR I aim to make wsl-pro-service quit gracefully after a predefined number of consecutive retry attempts to connect to the Windows agent. Since systemd is set to restart it always, I'm changing the restart interval from 2 seconds to 20 minutes. So we'd be trying to connect to the Windows agent for roughly 10 min, bailing out and staying quiet and silent for 20 minutes before trying again.

We could consider alternatively asking systemd to only restart us on failure, and staying quiet for the entire duration of a boot instead. Or even establishing specific error codes for when systemd should and should not restart us.

Those alternatives as well as the magic numbers sprinkled here and there are pretty much open for debate.

One final note: I did a major refactoring of `daemon.Serve()` because otherwise adding a test case to verify that method would return no error when the maximum number of retrial attempts is reached would require having such test case running for 10min or more, just waiting. Decoupling the retrial logic allows us to infer that indirectly by covering all possible retry scenarios in half a second, and trusting that `daemon.Serve()` won't error out if `retryConfig.Run()` doesn't. I could have stopped the changes in Go code at the first commit but that would be suboptimal IMHO.

---

UDENG-6571

### DIFF
--- a/wsl-pro-service/internal/daemon/daemon.go
+++ b/wsl-pro-service/internal/daemon/daemon.go
@@ -128,6 +128,7 @@ func (d *Daemon) Serve(service streams.CommandService) error {
 		maxWait      = time.Minute
 		growthFactor = 2
 	)
+	retries := 0
 	wait := 0 * time.Second
 
 	// Signal systemd before dialing for the first time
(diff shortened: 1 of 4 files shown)


ashuntu (approved):
Looks good as is to me

I think the magic numbers you chose seem reasonable. I'm not sure how likely it is that the windows agent is unreachable, but I think it is worth keeping the retry during a boot so that the connection can potentially be reestablished.

Is there a reason you chose to eventually cancel retries, instead of a larger max retry time (e.g. 30 minutes)?

PR #1186, opened 2025-04-01
fix(wsl-pro-service): Raise systemd restrictions again

CarlosNihelton:
In #814  we lowered systemd confinement a little bit to permit the pro-client to execute livepatch, even though it's currently not applicable for WSL (and even undesirable as it delays pro attaching to install that snap).

With ubuntu-pro-client version 35, currently in `-proposed` for the interesting LTSes, that behaviour was fixed, pro client correctly detects WSL and no longer attempts to run livepatch, not even when subprocessed through a confined systemd service.

That fix allows us to raise the systemd restrictions again.

--- 

UDENG-3358


ashuntu (approved):
LGTM :)

PR #1195, opened 2025-04-09
fix: Upgrade go toolchain and fixes CI

CarlosNihelton:
Ci is pretty broken due new vulnerability disclosures in the Go toolchain, such as:

```
 Checking known vulnerabilities
=== Symbol Results ===

Vulnerability #1: GO-2025-3563
    Request smuggling due to acceptance of invalid chunked data in net/http
  More info: https://pkg.go.dev/vuln/GO-2025-3563
  Standard library
    Found in: net/http/internal@go1.23.6
    Fixed in: net/http/internal@go1.23.8
    Example traces found:
Error:       #1: internal/system/system.go:264:13: system.System.findCmdExe calls bufio.Scanner.Scan, which eventually calls internal.chunkedReader.Read

Your code is affected by 1 vulnerability from the Go standard library.
This scan found no other vulnerabilities in packages you import or modules you
require.
```

Also, I introduced a racy test in my previous PR. `TestRetryLogic/With_a_precancelled_context` is inherently racy a because a pre-cancelled context makes the Run() method race wit time.After(). It's the case because the initial wait is 0, so which of ctx.Done() or time.After(0) will be preferred is unknown.

This fixes the test to assume that race condition and change the assertions slightly.

```go
		select {
		case <-ctx.Done():
			return nil
		case <-time.After(wait):
		}

```

---
UDENG-6698


wsl-pro-service/internal/daemon/daemon_test.go
```diff
@@ -367,19 +367,22 @@ func TestRetryLogic(t *testing.T) {
...
 			tooManyAttempts := false
 			rc := daemon.NewRetryConfig(minWait, maxWait, maxRetries)
+			// All functions passed below run in the same goroutine, thus no need for
+			// synchronisation.
 			err := rc.Run(ctx, func() (bool, error) {
 				return tc.succeedWithoutRetries, tc.actionError
 			}, func(wait time.Duration) {
 				if tc.wantNoRetries {
-					require.Fail(t, "Unexpected Retry attempt")
+					require.LessOrEqual(t, tryCount, 1, "Unexpected Retry attempt")
```
ashuntu, line 383: Should this be `require.Less` (or just assert 0) instead of `require.LessOrEqual`? For example, if we get here after the first attempt, `tryCount` is 1, but it should fail here since we don't want to have called the wait function again.

ashuntu (changes requested):

ashuntu (approved):

PR #1197, opened 2025-04-09
deps(gui): Upgrade Flutter to version 3.29

CarlosNihelton:
Will be current with the latest stable, better do it now than later. There is a deprecation warning about synthetic packages, which affect how we use the app l10n, but I'm deferring that task for its own PR, as it touches Dart code pretty much everywhere.

This will unblock two more dependabot's PRs.


ashuntu (approved):
LGTM

PR #1220, opened 2025-05-21
fix: Ensure Built-Using is populated

didrocks:
Using golang build package system is not enough, we need to ensure that golang debhelper is hooked in too to populate that field.


CarlosNihelton (approved):
Thanks for figuring this out!

PR #1227, opened 2025-06-05
docs: new distro format

edibotopic:
Updates the documentation to include WSL's new distribution format and recommend different wsl commands for common tasks.

# Main changes

* Defaulting to `wsl -d <Distro>` to run instances instead of `<Distro>.exe`
* Referencing use of `wsl ~` to start a distro in Ubuntu home dir, as opening in mounted C drive could be unexpected for users
* Changing to `wsl --install <Distro>` instead of `wsl --install -d <Distro>`
* Adding section on tar-based Ubuntu distros to install guide
* Adding explanatory notes on new distro format in relevant pages

As we are still in a transitionary phase between the old and new formats, we shouldn't remove information on the old format yet; for example, the reference page on Ubuntu distributions still talks about the AppxPackage, etc. There, I added an explanatory note about the two formats.

# Future work

* When other Ubuntu on WSL releases are migrated, we will need to revisit the documentation again to reflect the increased availability of Ubuntu in the new format.
* We could offer guidance on customising and deploying Ubuntu tar-based distros, in addition to a dedicated explanation of the new format and its advantages, but that is outside scope for this PR.

# Other minor changes

* Some additions to the tutorial to better meet the needs of beginner users (installing, running, exiting)
* Fixed some code blocks that didn't use our regex for excluding prompt icons or our no-copy-button for output logs
* General edits to improve flow and consistency

UDENG-6995

### DIFF
--- a/docs/howto/cloud-init.md
+++ b/docs/howto/cloud-init.md
@@ -8,12 +8,10 @@ myst:
 (howto::cloud-init)=
 # Automatic setup of Ubuntu on WSL with cloud-init
 
-Cloud-init is an industry-standard multi-distribution method for cross-platform cloud instance initialisation.
-Ubuntu WSL users can now leverage it to perform an automatic setup to get a working instance with minimal touch.
+Cloud-init is a cross-platform tool for provisioning cloud instances.
+It is an industry standard and can now also be used to automatically setup instances of Ubuntu on WSL.
 
 > See more:  [cloud-init official documentation](https://cloudinit.readthedocs.io/en/latest/index.html).
-
-The latest release of Ubuntu 24.04 LTS (Noble Numbat) comes with cloud-init already preinstalled, so you'll need that specific application to follow this guide. Ubuntu 24.04 LTS can be installed from [this link to the Microsoft Store](https://apps.microsoft.com/detail/9nz3klhxdjp5?hl=en-us&gl=US). A previous version of this guide used Ubuntu (Preview), because that comes with the latest in-development features. You can still use it to follow the instructions below, if you prefer. This feature is now available in the default Ubuntu application as well as Ubuntu 22.04 LTS.
 
 ## What you will learn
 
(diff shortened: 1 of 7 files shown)


docs/howto/cloud-init.md
hk21702, line 1: These instructions might not work for certain situations.

On newer versions of WSL, after installing a new instance, by default the distribution is launched immediately. As a consequence, users are prompted through setup right at install time. From what I can tell, if the user doesn't immediately exit out of the setup script (and in some cases even if they do), cloud init won't trigger if setup afterward. 

I'd suggest that instead of starting with a WSL2 distribution already installed, have the user setup cloud init *before* installation. That way, cloud init is guaranteed to run.

docs/howto/cloud-init.md
```diff
@@ -77,16 +78,12 @@ Save it and close it.
 In PowerShell, run:
 
 ```{code-block} text
-> ubuntu2404.exe
+> wsl ~ -d Ubuntu-24.04
 ```
 
-This command will register a new Ubuntu-24.04 instance that will be configured automatically by cloud-init.
+This command registers a new Ubuntu-24.04 instance that will be configured automatically by cloud-init.
```
hk21702, line 84: To be precise with language, the command shown here runs an already registered instance that is named `Ubuntu-24.04`. If there isn't a distribution registered with a matching name, the command will fail.

hk21702 (changes requested):
Looks good overall! I would like to see a change to how the cloud init section is handled, to reduce user error. See the comment I made for more details.

For the installation section, I also wonder if it would be better to change the order of methods presented so that the store method is presented after the more modern methods, but I'll leave that up to you.

docs/howto/cloud-init.md
edibotopic: You were correct @hk21702 . I tested a few different scenarios and although it's possible to quickly exit from the install/auto-launch and have cloud-init provisioning work, we can't really expect users to do that.

So, the two options are:

1. Your suggestion: set up cloud-init and then install the distro
2. Alternative: install with `wsl --install --no-launch` and then set up cloud-init before running the distro for the first time

(1) is probably the cleanest but maybe (2) can be specified in a note.

edibotopic:
Thanks again @hk21702 .

Based on your comments:

* Changed order of presentation for installation methods: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1227/commits/1682f60b2566daec9e42882387963db5a1029ec2
* Emphasised installing of WSL distro _after_ setting up cloud-init: https://github.com/canonical/ubuntu-pro-for-wsl/pull/1227/commits/d4b39a895d7c747ec48cdaca8d75568e77e08719

PR #1235, opened 2025-06-18
deps: Upgrade the Go toolchain

CarlosNihelton:
The motivation:

image

Upgrading will unblock many PRs currently open in a broken state.


hk21702 (approved):
I can't say that much about the Flutter changes, but overall LGTM!

CarlosNihelton:
Those are generated by the protoc Dart plugin, you can relax about them because CI would fail if a clean re-generation in CI had any diffs from what is in the PR ;)

PR #1254, opened 2025-07-17
Regenerates dart bindings

CarlosNihelton:
A new version of the protoc plugin for dart was released hours ago. It looks like the main changes were just formatting, but that's enough to break CI :(


hk21702 (approved):
LGTM!

PR #1259, opened 2025-07-22
feat(windows-agent): Handles permission denied errors sent by the server

cnihelton:
The Landscape server is adding a capability to terminate connection attempts from UP4W with status code PermissionDenied if the calling account doesn’t have a certain feature flag set. The current behaviour of UP4W agent is to spam the server with new connection attempts, respecting some exponential backoff. Instead, we should detect that particular error code and stop calling Landscape for now. 

--

UDENG-7290


CarlosNihelton:
Just rebased over latest main, only the last commit is  new.

hk21702 (approved):
LGTM

I don't think the two missed lines are critical and I don't think it is that trivial to get coverage of them, so it is probably something we can let pass.

PR #1267, opened 2025-07-28
fix(CI): `needs` is child of `job`, not `step` [plus some go deps]

cnihelton:
But steps can refer to each other by ID anyways.

I wonder why we only see that once the error lands in main: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/16577445433/workflow

---

Go vulncheck won't let this complete without upgrading `github.com/go-viper/mapstructure/v2` and I'm not too excited to open another PR just for that.


hk21702:
The error is only showing up in main since the workflow isn't being dispatched for PRs is my guess.

PR #1285, opened 2025-08-14
feat: Surface agent early errors when connecting to Landscape

cnihelton:
Now it's time for a real UI proposal:

A few error samples (now with the suggested UI enhancements ❤️ ):

https://github.com/user-attachments/assets/67baf15f-edf8-45bc-a16c-4cbee44de1f4

https://github.com/user-attachments/assets/7bf47e96-21d5-434a-8ce2-5d7957264063

And the success case

https://github.com/user-attachments/assets/635779e8-8402-445b-848a-c19d21be2701

----

The Flutter side of this PR is rather simple. A small refactoring in the `NavigationRow` to allow us to customise the "next" button. I'm overlaying the `CiruclarProgressIndicator` and zeroing the text opacity to keep the button size stable. While just replacing the Text widget with the spinner would do it, the button resizes and that's a bit disturbing. Despite that, a change in the `LandscapeModel`'s `tryApply` method enforces the return of a `GrpcError` object in all cases, which allows the UI to `switch(err.code)` and select a few interesting cases to customize the title and contents of the modal alert dialog shown containing feedback about the error, especially the error sent by the agent is placed in a `SelectableText` so one can easily copy that and paste in a bug report. The fact that the alert is modal is important to me because there is some serious acknowledgement and maybe action taking going on, being barrier-dismissible makes it too easy to unintentionally dismiss the alert. The user needs to read the alert and decide what's next. They can proceed anyway, storing the apparently wrong configuration if they have reasons for doing so (let's not block the user) or they can go back to where they were and fix potential mistakes.

The Go side is orders of magnitude more complicated because of the way we track and restart the command stream. At the transport layer, gRPC handles autoreconnection as needed when RPCs are initiated, so a healthy connection can go `Idle` and that's normal and expected. But the reconnection doesn't restart streams, that's why we have the intricated arrangement of the Landscape component methods `keepConnected()`, `connectOnce` (that sets up a connectivity monitoring goroutine) and the `receiveCommands()` which blocks on receiving commands from the gRPC stream. It's especially tricky to declare that we have a good connection, because it means the `receiveCommands` goroutine is blocked in `grpcClient.Recv()`. The fact that the Landscape server sends us a hostagent UID before validating our credentials also complicates things because we cannot assume everything is right when the first command is received and executed, as we discussed in my previous PR. We need to track changes to the hostagent UID and consider that in addition to being blocked in the `Recv` plus having a good connectivity state (`Idle` or `Ready`). 

I added extensive comments around those methods (which to me is already a code smell) because there are non-local effects affecting their behaviour. For the near future I think we'll need to rewrite this component to make this look more local and easier to reason about.

There were also a caveat about the `Config` component not propagating the data submitted when it's not new (which is right, no need to wake up WSL instances to reconnect to the same Landscape server with the same credentials) but that would cause the UI RPC to block forever waiting a completion signal, so I changed the semantics of this part to become an error, thus unblocking the UI RPC early.

---
UDENG-7432

### DIFF
--- a/gui/packages/ubuntupro/lib/l10n/app_en.arb
+++ b/gui/packages/ubuntupro/lib/l10n/app_en.arb
@@ -52,8 +52,8 @@
     "purchaseStatusSuccess": "Thank you for subscribing to Ubuntu Pro.",
     "purchaseStatusAlreadyPurchased": "You are already subscribed to Pro.",
     "purchaseStatusUserGaveUp": "Purchase attempt was canceled.",
-    "purchaseStatusNetwork": "A network issue prevented the purchase to proceed. Please try again once connected.",
-    "purchaseStatusServer": "Something went wrong with Microsoft Store. Please try again later.",
+    "purchaseStatusNetwork": "A network issue prevented the purchase from proceeding. Please try again once connected.",
+    "purchaseStatusServer": "Something went wrong with the Microsoft Store. Please try again later.",
     "purchaseStatusUnknown": "Unknown error when trying to purchase the subscription. Consider restarting this app.",
 
     "landscapeSkip": "Skip for now",


gui/packages/ubuntupro/lib/pages/landscape/landscape_page.dart
```diff
@@ -113,6 +179,56 @@ class LandscapePage extends StatelessWidget {
...
+                    Expanded(
+                      child: SelectableText(details),
+                    ),
+                  ],
+                ),
+              ),
+              const SizedBox(height: 16),
+              Text(
+                lang.landscapeProceedAnywayHint,
+                style: Theme.of(context).textTheme.bodySmall,
```
ashuntu, line 223: Also, in my opinion, this should be the same size as the other text since generally I think that's what has been recommended in the past for other areas.

gui/packages/ubuntupro/lib/pages/landscape/landscape_page.dart
```diff
@@ -113,6 +179,56 @@ class LandscapePage extends StatelessWidget {
...
+            children: [
+              Text(content),
+              const SizedBox(height: 8),
+              YaruExpandable(
+                expandButtonPosition: YaruExpandableButtonPosition.start,
+                header: Text(lang.landscapeDetails),
+                child: Row(
+                  children: [
+                    Expanded(
+                      child: SelectableText(details),
```
ashuntu, line 215: What happens if this text is excessively long? Does it scroll?

windows-agent/internal/proservices/ui/ui.go
```diff
@@ -76,8 +125,21 @@ func (s *Service) ApplyProToken(ctx context.Context, info *agentapi.ProAttachInf
...
+	} else if err != nil {
+		return nil, err
+	}
+
+	// Blocks until the Landscape service receives an interesting response from the server,
+	// thus preventing premature response to the UI and properly propagates any errors.
+	if err = <-s.landscapeListener; err != nil {
+		log.Warningf(ctx, "UI service: ApplyLandscapeConfig: %v", err)
 		return nil, err
 	}
```
hk21702, line 144: Similar thing here, technically possible risk of blocking forever should `Stop()` be called in a different thread, and the channel be set to nil before recieving. We also aren't checking the results of s.drainLandscapeListener, so this would be problematic if `Stop()` is called at any point, and SetUserLandscapeConfig for whatever reason doesn't error (or if somehow the channel is closed and set to nil between setting the config and reading the listener).

windows-agent/internal/proservices/landscape/executor_test.go
hk21702, line 1: Nitpick: From what I can tell, it is fine, but just a note that static code analysis is complaining about `SA4009` in `testReceiveCommand`. You could add a comment to ignore the analysis warning, or make `testReceiveCommand` pass in pointers for `homedir` and `downloaddir` to fix it probably.

hk21702 (changes requested):
Looks okay overall.

I have a few nitpicks and also some concerns about the life cycle management of the `landscapeListener` channel in the UI service possibly resulting in perpetual blocks or even panics if used improperly and if the stars align. It is probably incredibly rare even if used improperly, but best to make a 0.001% chance into a 0% chance.

gui/packages/ubuntupro/lib/l10n/app_en.arb
```diff
@@ -95,6 +95,20 @@
         }
     },
     "landscapeRegister": "Register",
+    "landscapeDetails": "Details:",
+    "landscapeProceedAnywayHint": "If you are sure the configuration is correct you can click on \"Proceed anyway\".",
+    "landscapeEditYourConfig": "Edit your config",
+    "landscapeProceedAnyway": "Proceed anyway",
+    "landscapeInvalidConfig": "Invalid Landscape configuration",
```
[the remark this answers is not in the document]
CarlosNihelton: We use regular sentence case, Landscape only goes capitalized because it's a proper name.

CarlosNihelton (commented):
Just rebased over the latest main. New commits start at f5433052cc74891cfd36ce059bdc2d0f4c44c82c

gui/packages/ubuntupro/pubspec.lock
```diff
@@ -12,11 +12,9 @@ packages:
   agentapi:
```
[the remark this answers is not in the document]
CarlosNihelton: Definitely not :)

CarlosNihelton:
@edibotopic you might like to see those screens and provide opinion about their contents and design.

ashuntu (approved):
Flutter side looks good

windows-agent/internal/proservices/ui/ui.go
hk21702: I've realized that the issue with the mutex solution is that now if you block waiting to receive a signal, a call to `Stop()` wouldn't be able to acquire the mutex and you could also end up blocking forever...

Maybe have the receiving from the listener channel be done in a switch statement that allows you to exit on context cancel? We'd probably want this context to be cancellable by `Stop()` without it needing to acquire the mutex just yet. Using the one passed into `ApplyLandscapeConfig` would still be iffy since we still can't guarantee anything about that context and how it relates to `Stop()`. 

Things get a bit messy because of a lack of guarantees about the channel's lifecycle.

windows-agent/internal/proservices/ui/ui.go
```diff
@@ -37,13 +40,59 @@ type Service struct {
...
+	// Drain the channel to prevent blocking on write.
+	if !s.drainLandscapeListener(ctx) {
+		return
+	}
+
+	select {
+	case s.landscapeListener <- err:
+	case ctxErr := <-ctx.Done():
+		log.Warningf(ctx, "UI service: When notifying about Landscape connection: %v", ctxErr)
 	}
```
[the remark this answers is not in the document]
hk21702: Similar thing here, though it is a little better since there is a select case using the context passed in. We don't have a guarantee right now that the context passed in will be canceled should `Stop()` be called, so we could block forever at `Stop()` trying to acquire the mutex should we be blocked here due to a full buffer.

hk21702 (changes requested):
I noticed an unfortunate flaw with my earlier suggestion of using mutexes. We've mostly just moved the source of a perpetual block to the `Stop()` method. It is a little better since panics should be impossible without significant misuse, but we may need to have a context that is controlled by `Stop()` that is cancelled before trying to acquire the mutex. We'd then use this context to signal to users of the `landscapeListener` channel that we should stop blocking on it since we're initiating a shutdown.

We could get away with not having this in `ApplyLandscapeConfig` if we don't set the channel to nil after closing it and we guarantee that it is always non-nil even if closed. But the precaution still needs to happen in `LandscapeConnectionListener`

edibotopic (changes requested):
I can't really comment on the implementation and I was not able to fully test the GUI with a live Landscape server, so I will just make some observations and ask some questions.

* **Bad FQDN**: if someone enters an incorrect value for FQDN, should there be some text that reminds the user that it must be an FQDN and not a full URL, or prompts the user to ask their admin for the FQDN, or points them to info about what an FQDN is? I say this assuming that it might be a common issue.

* **Hint about proceeding**: if the entered config is correct (e.g., a valid FQDN), but the server cannot be connected  for a different, possibly transient reasons (server down, network issues), will the connection be attempted before and/or after the user "proceeds anyway"? If so:
        - should that be made more obvious?
        e.g., UI: "Re-attempting connection in 5 seconds..."
        e.g., text: "Your Landscape server may be down. If you continue, repeated attempts to connection will be attempted until success."

* I would suggest **making the hint less generic**. For example:
        - if the user is being asked to check that the address they entered is correct, then they should be told to proceed "if you are sure that the address you entered is correct", not the more general "configuration"
        - if the user is being told that Landscape WSL features are disabled on the server, then it's not obvious what relevance their configuration has to their choice about whether to proceed
        - if the user enters the wrong registration key, it's not clear how they could determine that it is correct before proceeding (it's right or wrong?). If so, they should probably be asked to enter a valid key or contact their admin?

* **Proceed anyway button text**: is it clear what this implies?
        - does "proceed" here mean, proceed with this configuration and Landscape connection might work eventually, or...
        - abandon the landscape configuration because it didn't work and proceed to the status page showing Pro-attachment?
* **Proceed anyway button colour**: The red indicates that this is an action that should be taken with caution but the user is asked to take the action only if they are sure that they have configured things correctly.
        - it doesn't seem clear why they should be cautious?

We've discussed this before, and I'm not sure what's possible currently, but if Landscape has been misconfigured in some way, and the user proceeds, then perhaps there should be some indicator on the status page to that effect ( e.g., Ubuntu Pro is enabled (top), there was a problem with your Landscape configuration (bottom) ).

edit: some of these comments may well be outside of scope for this PR but I just wanted to note them for discussion.

edit: removed comment about orange colour of spinner as out-of-scope.

windows-agent/internal/proservices/ui/ui.go
CarlosNihelton: I think a context for the service can be a reliable way to manage this without the need for assigning the channel to `nil`.

hk21702: Indeed, a context would work for this; the only caveat is that the context needs to be controlled by `Stop()`.

CarlosNihelton:
Regarding @edibotopic's comments, let me elaborate on some:

**Proceed anyway**: means store whatever the user submitted and eventually attempt to use that configuration again. That option exists to acknowledge cases like transient networking issues like you described, the server might be down for some reason etc. If the user is sure that configuration is correct and will eventually work, they can proceed anyway, ignoring the current error state. That requires being cautious because if the user's assumptions are not correct, the connection will never work until they come back to this screen manually and submit a correct configuration.

**Making the hint about proceeding less generic**: good point! I'm working next on that topic still in this PR. I think it's worth it.

**Bad FQDN**: we already have validation in the form itself to refuse input that doesn't represent an FQDN. The actual error message shown as a form field error is: `Invalid FQDN. Format should be a hostname or IP address.` It's indeed more common than what we originally thought it would be.

CarlosNihelton:
@ashuntu @edibotopic , about making the hint less generic, my latest two commits produce a more verbose screen.

image

image

Thoughts?

PR #1288, opened 2025-08-20
docs: add google analytics

edibotopic:
This adds a cookie consent banner to the documentation.

If a user consents, Google Analytics will begin collecting data.

A "manage your tracking settings" option is also added to the footer of the documentation,
allowing users to change their preferences. 

Screenshot from 2025-08-20 14-57-38


hk21702 (approved):
Looks good overall and works in Firefox!

The CSS breaks with my ad blocker settings, but it appears to be otherwise fine. This could probably be considered to just be a case of users breaking things, but it may also be something to consider. 
image


Another note is that I'm still unsure about the lack of a "decline all non-essential" button on the root level, but that would be something to take up with design.

edibotopic:
Thanks @hk21702 !

Do you have specific ad-blocker settings that you could share?
I/you could report the problem to the people maintaining the cookie banner.

When it comes to the "decline all non-essential" option --- I agree --- and I believe the web team are planning to implement it.

hk21702:
@edibotopic uBlock Origin on Firefox with these settings:

image

From a quick test, it is the "Cookie Notices" block that is half-breaking things.

For user experience, it would probably be best to let the ad blocker do its thing and not half-break and not engage in an arms race with the adblock/filter authors.

PR #1302, opened 2025-08-22
docs: install guide fixes

edibotopic:
Several changes are made here to improve the install guide:

* Clarification of what the `wsl --install` command does (installs and enables WSL on a fresh system)
* Clarification that a reboot is required after installing and enabling WSL
* Clarification that the default WSL distro is Ubuntu but that it can be changed
* Simplification of section on install from Microsoft Store
* Deletion of image assets that we don't need
* Some minor fixes for style and flow

UDENG-7611


edibotopic:
All checks are passing locally.
The inclusive language check is handing in the CI for some reason, so marking this for review in the meantime.

docs/howto/install-ubuntu-wsl2.md
```diff
@@ -151,12 +168,15 @@ the working directory, run:
...
 ````{tip}
-For the **default Ubuntu distro only**, this command can be shortened further to:
+For the **default WSL distro**, this command can be shortened further to:
 
 ```{code-block} text
 > wsl ~
 ```
 
+The default distro for WSL is Ubuntu, although [this can be
+configured](https://learn.microsoft.com/en-us/windows/wsl/basic-commands#set-default-wsl-version).
```
CarlosNihelton, line 178:
```suggestion
configured](https://learn.microsoft.com/en-us/windows/wsl/basic-commands#set-default-linux-distribution).
```

CarlosNihelton (changes requested):
Great job! Just one minor correction and we should be good to go!

CarlosNihelton (approved):

PR #1303, opened 2025-08-22
docs: tutorial fixes

edibotopic:
Various fixes to improve Develop with Ubuntu on WSL and getting started with UP4W tutorials, including:

* Improve clarity on use and purpose of `wsl --install`
* Improve clarity on need to reboot after WSL setup
* Default to terminal-based installation method
* Remove unneccessary screenshots
* Add short section to Develop tutorial on running multiple versions of Ubuntu, which is a good point to mention listing installed versions, unregistering an instance, and Windows terminal integration
* Update codeblocks in UP4W tutorial to improve functionality
* Various fixes to improve clarity, structure, flow and correctness

UDENG-7396


docs/tutorials/develop-with-ubuntu-wsl.md
```diff
@@ -25,17 +25,43 @@ offers excellent integration with developer tools like Visual Studio Code.
...
+```
+
+To install Ubuntu 24.04 LTS, run the following command in a PowerShell terminal:
+
+```{code-block} text
+> wsl --install Ubuntu-24.04
+```
+
+After an Ubuntu distribution is installed, you are prompted to create a
+username and password. An Ubuntu terminal will then open automatically.
```
CarlosNihelton, line 64: We may want to repeat the word "Ubuntu" a bit less.

```suggestion
After the distribution is installed, you are prompted to create a
username and password. An Ubuntu session will then start automatically.
```

Also, we are already in a terminal, that terminal was "connected" to the Windows PowerShell and now it will be "connected" to the Ubuntu shell for the duration of the user session. I'm not entirely sure I'm phrasing it in the best way, but the existing text suggests that a new terminal window or tab will open, which is not the case.

docs/tutorials/develop-with-ubuntu-wsl.md
```diff
@@ -53,95 +79,94 @@ PS C:\Users\username>
...
 
-To exit the Ubuntu terminal, type the `exit` command and execute it by pressing
-<kbd>enter</kbd>, which will return you to the PowerShell prompt.
-
-```{tip}
-It is recommended to reboot your machine after this initial installation to
-complete the setup.
-```
+To exit the Ubuntu terminal at any time, type the `exit` command and execute it
+by pressing <kbd>enter</kbd>, which will return you to the PowerShell prompt.
```
CarlosNihelton, line 83: Just learned a new HTML tag 😄

docs/tutorials/develop-with-ubuntu-wsl.md
```diff
@@ -53,95 +79,94 @@ PS C:\Users\username>
...
 ```
 
-Two instances of Ubuntu are installed:
-
-1. The default Ubuntu version that was installed automatically when you installed WSL
-2. The numbered Ubuntu version that you installed manually
+```{admonition} What is version 2?
+:class: note
+This means that you are using WSL2, rather than WSL1.
+WSL2 is the default WSL on recent versions of Windows.
```
CarlosNihelton, line 114:
```suggestion
WSL implements two different architectures for running 
Linux distributions: WSL1 and WSL2.
This means that you are using WSL2, rather than WSL1.
WSL2 is the default WSL on recent versions of Windows.
```

CarlosNihelton: If the user is questioning what's the VERSION 2 value, they likely don't know about WSL1 x WSL2, so for those we are still not answering their question.

docs/tutorials/develop-with-ubuntu-wsl.md
```diff
@@ -53,95 +79,94 @@ PS C:\Users\username>
...
 
-The `~` is passed to the `wsl command` to start the instance in the Ubuntu home directory,
+The `~` is passed to the `wsl command` to start the instance in the Ubuntu home directory and
 the `-d` flag is added before specifying a distro.
 
-## Install Visual Studio Code on Windows
+```{admonition} Windows terminal integration
+:class: tip
+Each time you install a version of Ubuntu, it appears in the dropdown list of
+command line applications in Windows terminal.
```
CarlosNihelton, line 129:
```suggestion
Each time you install a version of Ubuntu, it appears in the dropdown list of
terminal profiles in Windows terminal.
```

Those are terminal profiles, they define appearance and which shell the terminal must "connect" to (The Windows Command Prompt, the Windows PowerShell, Azure Cloud Shell, Ubuntu on WSL shell etc). Although it's possible to define a terminal profile to run a single general command line application that's not what it's intended for.

CarlosNihelton (commented):
I have minor comments for your consideration.

docs/tutorials/develop-with-ubuntu-wsl.md
edibotopic: I agree and think that "session" is better and more accurate.

The only thing is that we probably need to use that term more consistently elsewhere also (work for another time).

edibotopic:
> I have minor comments for your consideration.

Thank you. I agree with all of them. Accepted the changes and squashed the commits.

PR #1304, opened 2025-08-22
feat(CI): Release in two channels with or without MS Store

cnihelton:
This PR adds an interesting ability to our CI to publish the MSIX package in two channels: stable and unstable. Our CI was already configured to publish an MSIX package when a tag is pushed to the repository, but that only considered one publication channel.

1. We want the MSIX autoupdates to differentiate those channels. A user running a stable version of the app should upgrade to the next stable version while pre-release runners should update to the next pre-release which is expected to be published faster and more often than stable releases.

2. We want the stable release package to remain in sync with the stable version of the documentation website. Read-the-Docs use a simple strategy: any tag numbered as a semantic version is a stable release and the highest number means the latest stable, disregarding when that tag was created or in which branch the commit lives. Besides stable, Read-the-docs keep the 'latest' version of the website, which is generated from the latest commit in the default branch (main). Finally, the tool interprets any suffix applied to a semantic versioned as a 'pre-release', which is never published in the 'stable' version of the website. [See the docs](https://docs.readthedocs.com/platform/latest/versions.html#versioning-workflows).

We achieve 1 by publishing two versions of the app installer file at different locations: the stable one remains hosted at GH releases and always points to the `/releases/latest/` URI pattern. The unstable one is hosted in a different location generated by leveraging the project wiki (hack :p ). CI needs then to differentiate when building a stable or unstable MSIX package and update the app installer file accordingly and if the package is part of a stable release, the app installer file is included as a release asset, otherwise it goes to the wiki. In both stable and unstable cases we still want to push the packages to the MS Store, at least to be signed. The unstable versions are submitted as updates to an existing  package flight, which is the MS Store way to implement different publishing channels. Finally to allow pre-release users to upgrade to the next major stable release (like from 1.9999.3 to 2.0.0) the CI always updates the appinstaller file stored in the wiki. Sometimes Windows will find the local installation having a higher version number than the one referred to in the wiki and that's fine, it won't downgrade it. Since the appinstaller file in the wiki is kind of hidden from users, this transient strange behaviour is acceptable.

We achieve 2 by using a tag convention: stable is a full semantic version number, like: `1.2.3`  or `0.3.2`, and unstable is something that Read-the-Docs would consider a pre-release tag, thus a suffixed semantic version number, like `0.5b6` (a for alpha or b for beta, for instance). There is a catch, though: despite the fact that MSIX packages declare their full version using 4 number fields, only three are usable, so the alpha/beta versions must remain using only 3 numbers (see [this note](https://learn.microsoft.com/en-us/windows/apps/publish/publish-your-app/msix/app-package-requirements#version-numbering-for-windows10-and-11-packages) if you want more details). By just looking at the package number `0.3.5.0` (the fourth is always 0 for desktop apps) we cannot determine at a glance that means an unstable release. So my solution is to leverage a special hardcoded value for the minor version: 9999. A package versioned as `0.999.1.0` is a pre-release version that will eventually become `1.0.x`. So the rule for tagging a release becomes:
- Major.Minor.Patch: means stable, CI builds a stable package version `Major.Minor.Patch.0`, publishes a draft GH release version `Major.Minor.Patch` including the app installer file as assets, creates a new submission to MS Store updating the MSIX and RTD updates the stable version of the website.
- Major.9999[ab]Patch: means unstable, CI builds an unstable package version `Major.9999.Patch.0`,  publishes a draft pre-release GH release version `Major.9999.Patch.0` without any release asset, update the app installer file hosted inside the project wiki, creates a new submission to the predefined package flight in MS Store updating the MSIX and RTD ignores that tag.

I'm using `9999` because it's quite visible (I remember GIMP using `2.99.patch` before releasing 3.0). If we think that's too limiting we could use up until 65535. 😅 

That required a few changes not only in the GH actions files but also at the helper `compute_version.go` to differentiate stable from unstable and output the correct number-only version to be used as the MSIX version. 

Finally, beware that CI creates a GitHub release draft. CI prints a URL from which we could download the package built by that job if needed. The exact procedure for a release is to be captured in a piece of documentation that covers how to release this application in further details (coming soon). The GitHub release remains in draft until the certification process in MS Partner Center is complete, so we can download a package signed by the Store and add it as a release asset. Only then the release should be published, what makes it visible for the public.

To demonstrate this in action, I created a package flight for UP4W (we didn't have that before because everything is still unstable at the moment) with version 0.9999.1:

Screenshot From 2025-08-26 13-41-27

Then I pushed a the tag `0.9999b2` and [let CI run](https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/17298622933/job/49103383174).

CI ran, created the [GitHub release draft 0.9999.2.0](https://github.com/canonical/ubuntu-pro-for-wsl/releases/tag/untagged-61f7c3b96250068cd881) named after the MSIX package version, updated the wiki (see [this file](https://raw.githubusercontent.com/wiki/canonical/ubuntu-pro-for-wsl/pre-release/UbuntuProForWSL.appinstaller) ) and created a new flight submission, as we can see in Partner Center (certification already in progress):

image

The draft release has a funny URL pattern. The one from above is: https://github.com/canonical/ubuntu-pro-for-wsl/releases/tag/untagged-61f7c3b96250068cd881. I thought that was an issue, but it's actually a feature: once the draft release is published the `/releases/tag/untagged-61f7c3b96250068cd881` would become `/releases/tag/0.9999b2` matching the tag.

That draft release is for exposition only, after this PR is fully reviewed I'll delete it.

UDENG-7654

### DIFF
--- a/.github/workflows/publish-msix.yaml
+++ b/.github/workflows/publish-msix.yaml
@@ -58,7 +58,7 @@ jobs:
           } else {
             $sub, $url = Update-ApplicationSubmission -AppId $appid -SubmissionDataPath "$subFileStem.json" -PackagePath "$subFileStem.zip" -Force -ReplacePackages -AutoCommit -Verbose
           }
-          Write-Output "If needed the packge built by this run can be downloaded from:"
+          Write-Output "If needed the package built by this run can be downloaded from:"
           Write-Output $url
           Write-Output "Access the submission via the link below:"
           Write-Output "https://partner.microsoft.com/en-us/dashboard/products/$appid/submissions/$sub/"


tools/build/compute_version.go
hk21702, line 1: This is probably out of scope for this PR. But, I feel like you might as well have this tool output JSON since it is primarily used as a helper for CI. Then, you wouldn't even need the extra flags to turn on and off functionality since it should be fine to expand if your user is parsing the JSON correctly. It is a little awkward to have to call this multiple times for all the things you need, and direcly parsing can be fragile.

.github/workflows/publish-msix.yaml
```diff
@@ -44,17 +44,61 @@ jobs:
...
+          if ($env:pre_release -eq "pre-release") {
+            # The existing package flight for beta testers.
+            $flightId="876dc297-0b4d-4a74-94ef-720279050fc4"
+            Write-Output "Starting a new submission under package flight $flightId from:"
+            Get-ChildItem ./msix/UbuntuProForWSL/AppPackages/
+            $sub, $url = Update-ApplicationFlightSubmission -AppId $appid -FlightId $flightId -SubmissionDataPath "$subFileStem.json" -PackagePath "$subFileStem.zip" -Force -AutoCommit -ReplacePackages -Verbose
+          } else {
+            $sub, $url = Update-ApplicationSubmission -AppId $appid -SubmissionDataPath "$subFileStem.json" -PackagePath "$subFileStem.zip" -Force -ReplacePackages -AutoCommit -Verbose
+          }
+          Write-Output "If needed the packge built by this run can be downloaded from:"
```
hk21702, line 61:
```suggestion
          Write-Output "If needed the package built by this run can be downloaded from:"
```

hk21702 (approved):
Looks okay overall, though my ability to parse Powershell isn't the best. I'm pre-approving pending the one spelling mistake.

I do think that it might be worth refactoring the `compute_version` helper Go script at this point since it is getting a bit unwieldy. In particular, I think it would be a significant benefit to start outputting in a JSON format instead, meaning you can get rid of the flags and also the need to call it multiple times if you need multiple things from it, and making things just more robust and extendable in general. But, even if so, that task is probably out of scope for this PR.

edibotopic (approved):
Looks sensible to me and this should work for the docs. Thanks for the explanation.

PR #1306, opened 2025-08-25
docs: add release notes page

edibotopic:
Adds a release notes page to the Reference section of the docs.

Goals:

* Have a single place to find Ubuntu-on-WSL-specific release notes
* Emphasise the latest supported versions of the distro and the Windows app
* Focus on highlights (this is not a changelog)
* Include notes for older LTS releases and interim releases in optional, dropdown sections

UDENG-7637

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -36,6 +36,8 @@ OpenVINO
 passwordless
 powershell
 PowerShell
+Quokka
+reproducibly
 rootfs
 rootFS
 SHA
@@ -60,10 +62,12 @@ autopkgtest
 autopkgtests
 backporting
 binfmt
+cdimage
 changelog
 compinit
 config
 dev
+exe
 gapped
 gpg
 grpc
@@ -79,6 +83,7 @@ Hostagent
 JWT
 localhost
 LTS
+MotD
 Numbat
 os
 pem
@@ -92,6 +97,7 @@ ssl
 tcp
 TLS
 toolchain
+upgrader
 url
 vGPU
 webserver
@@ -100,6 +106,7 @@ whitespace
 zshrc
 ESM
 Ok
+ppid
 scalability
 Snapcraft
 UTF
--- a/docs/reference/release_notes.md
+++ b/docs/reference/release_notes.md
@@ -1,0 +1,149 @@
+---
+myst:
+  html_meta:
+    "description lang=en":
+      "Release notes for Ubuntu on WSL and Ubuntu Pro for WSL."
+---
+
+(ref::release-notes)=
+# Release notes for Ubuntu on WSL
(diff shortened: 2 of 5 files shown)


docs/reference/release_notes.md
```diff
@@ -0,0 +1,145 @@
...
+
+##### Systemd by default
+
+systemd is now enabled by default even when the instance is launched directly from a terminal with the wsl.exe command or from an imported root files system.
+
+##### Point release changes
+
+###### 24.04.3
+
+* None.
```
CarlosNihelton, line 79: There is an interesting highlight about systemd-binfmt I'd like to propose. We can do it now or later when I update the related explanation page as well. WDYT?


```suggestion
* WSL setup: remove the `systemd-binfmt.service` override after WSL implemented a more robust binfmt protection mechanism.
([wsl-setup/#11](https://github.com/ubuntu/wsl-setup/issues/11))

* WSL setup: Fix an issue when the Windows user name contains non-ASCII characters
([wsl-setup/#23](https://github.com/ubuntu/wsl-setup/issues/23))
```

I'm less sure whether the second one is relevant or not to be highlighted here.

docs/.custom_wordlist.txt
```diff
@@ -110,3 +117,8 @@ UbuntuPro
 UbuntuProAutoStart
 UbuntuProToken
 LandscapeConfig
+
+
+
+
+
```
CarlosNihelton, line 124: nitpick: unnecessary diff :)

docs/reference/release_notes.md
```diff
@@ -0,0 +1,145 @@
...
+
+By default, Ubuntu distros downloaded as explicitly numbered releases (`wsl
+--install Ubuntu-24.04`) will not be upgraded in this manner, unless the
+following configuration change is made:
+
+```{code-block} diff
+:caption: /etc/update-manager/release-upgrades
+:class: no-copy
+- Prompt=lts
++ Prompt=normal
```
CarlosNihelton, line 37:
```suggestion
- Prompt=never
+ Prompt=normal
```

The numbered distros come with the `never` value. Users could change to `lts` or `normal`, which allows upgrading to the next LTS or the next interim release respectively.

CarlosNihelton (changes requested):
One minor correction

docs/reference/release_notes.md
edibotopic: I think these are fine additions :)

docs/.custom_wordlist.txt
edibotopic: hmmmm. not sure what happened there!

edibotopic:
I assume you're happy for this to get merged @CarlosNihelton ?

PR #1311, opened 2025-08-27
fix(docs): More emphasis about Landscape FQDN

CarlosNihelton:
It has to be a Fully Qualified Domain Name. Submitting complete URLs break communication attempts in many ways. Those errors were mostly silent, specially when the user submitted the config via the Advanced configuration option (where they crafted the config by hand).

An effort is being made in #1285 to present those errors to the UI, but that's still not enough.

UDENG-7545

### DIFF
--- a/docs/howto/set-up-landscape-client.md
+++ b/docs/howto/set-up-landscape-client.md
@@ -62,12 +62,27 @@ In the UP4W app navigate to the Landscape configuration screen:
 
 Choose your preferred configuration option and enter the required details.
 
-The "Advanced Configuration" option requires you to specify a `landscape.conf`.
+The "Manual configuration" is easier if the server was configured with the
+default options, only the "Landscape FQDN" field is trully required.
+Importantly, that field cannot have a complete URL, but rather the server's
+fully qualified domain name only. For user's commodity the field accepts
+a URL like `https://landscape-server.domain.com` but the `https://` part is
+removed from the address, resulting in the FQDN only. That's for allowing
+users to copy and paste the server address from a web browser address bar, for
+example.
+More complex URLs will be rejected and an error message will be shown.
+
+The "Advanced configuration" option requires you to specify a `landscape.conf`.
 Refer to the section on [Landscape client configuration](howto::config-landscape-client) for an example.
 
 When you continue a status screen will appear confirming that configuration is complete:
 
 ![Configuration is complete](../assets/status-complete.png)
+
+The application waits a fraction of second to confirm that the configuration
+data supplied resulted in a successful connection to the Landscape server. In
+case of errors a dialog presents the error details and lets you decide to edit
+the configuration and try again or simply proceed.
 
 ```
 
@@ -109,6 +124,12 @@ ssl_public_key = C:\Users\user\Downloads\landscape_server.pem
 ```
 
 ```{warning}
+The `url` field in the `[host]` section cannot be a true URL, but rather a fully
+qualified domain name (or IP address if your server was configured in that way)
+suffixed by a port (6554 by default) separated by a colon (`:`),
+as shown in the example above `landscape-server.domain.com:6554`. No scheme or
+other URL components are applicable.
+
 The `ping_url` must be a `http` address. A `https` address will not work.
 ```
 
--- a/docs/howto/set-up-landscape-client.md
+++ b/docs/howto/set-up-landscape-client.md
@@ -63,15 +63,16 @@ In the UP4W app navigate to the Landscape configuration screen:
 Choose your preferred configuration option and enter the required details.
 
 The "Manual configuration" is easier if the server was configured with the
-default options (you should double check with system administrator on that), for
+default options (if in doubt, check this with your system administrator), in
 which case only the "Landscape FQDN" field is truly required. This field does
 not accept a complete URL for the server (with paths and queries, for example),
 and a FQDN should always be used. The field accepts URLs like
-`https://landscape-server.domain.com` (with just scheme and host name) but the
-`https://` part (the scheme) is removed from the address, resulting in the FQDN
-only. This is for user convenience, as it allows you to copy and paste server
-addresses from a web browser address bar, for example. Richer URLs with queries,
-paths and fragments will rejected and an error message will be shown.
+`https://landscape-server.domain.com` (with a just scheme and a host name) but
+the `https://` part (the scheme) is removed from the address, resulting in the
+FQDN only. This is for user convenience, as it allows you to copy an address for
+a server from a web browser address bar, for example, and paste it into the
+field. Richer URLs with queries, paths and fragments will be rejected and an
+error message will be shown.
 
 The "Advanced configuration" option requires you to specify a `landscape.conf`.
 Refer to the section on [Landscape client configuration](howto::config-landscape-client) for an example.
--- a/docs/howto/set-up-landscape-client.md
+++ b/docs/howto/set-up-landscape-client.md
@@ -63,26 +63,28 @@ In the UP4W app navigate to the Landscape configuration screen:
 Choose your preferred configuration option and enter the required details.
 
 The "Manual configuration" is easier if the server was configured with the
-default options, only the "Landscape FQDN" field is trully required.
-Importantly, that field cannot have a complete URL, but rather the server's
-fully qualified domain name only. For user's commodity the field accepts
-a URL like `https://landscape-server.domain.com` but the `https://` part is
-removed from the address, resulting in the FQDN only. That's for allowing
-users to copy and paste the server address from a web browser address bar, for
-example.
-More complex URLs will be rejected and an error message will be shown.
+default options (you should double check with system administrator on that), for
+which case only the "Landscape FQDN" field is truly required. This field does
+not accept a complete URL for the server (with paths and queries, for example),
+and a FQDN should always be used. The field accepts URLs like
+`https://landscape-server.domain.com` (with just scheme and host name) but the
+`https://` part (the scheme) is removed from the address, resulting in the FQDN
+only. This is for user convenience, as it allows you to copy and paste server
+addresses from a web browser address bar, for example. Richer URLs with queries,
+paths and fragments will rejected and an error message will be shown.
 
 The "Advanced configuration" option requires you to specify a `landscape.conf`.
 Refer to the section on [Landscape client configuration](howto::config-landscape-client) for an example.
 
-When you continue a status screen will appear confirming that configuration is complete:
+When you continue, a status screen will appear confirming that configuration is complete:
 
 ![Configuration is complete](../assets/status-complete.png)
 
-The application waits a fraction of second to confirm that the configuration
+The application waits a short period of time to confirm that the configuration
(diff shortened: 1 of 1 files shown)


docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
...
 
-The "Advanced Configuration" option requires you to specify a `landscape.conf`.
+The "Manual configuration" is easier if the server was configured with the
+default options, only the "Landscape FQDN" field is trully required.
+Importantly, that field cannot have a complete URL, but rather the server's
+fully qualified domain name only. For user's commodity the field accepts
+a URL like `https://landscape-server.domain.com` but the `https://` part is
+removed from the address, resulting in the FQDN only. That's for allowing
+users to copy and paste the server address from a web browser address bar, for
+example.
```
edibotopic, line 72:
```suggestion
This field does not accept a complete URL for the server, and an
FQDN should always be used.  The field accepts
URLs like `https://landscape-server.domain.com` but the `https://` part is
removed from the address, resulting in the FQDN only. This is for user convenience,
as it allows you to copy and paste server addresses from a web browser address bar, for
example.
```

One point of clarification here, because I got a little confused:

When you say "complete URL", you are not referring to `https://landscape-server.domain.com` but instead something like `https://landscape-server.domain.com/something/else/`, or `https://landscape-server.domain.com:<PORT>`, right?

I could imagine a reader here thinking that "complete URL" is the one with the `https://` prefix, but then that is clearly allowed, as you explain, because it gets stripped. So I assume that's not what you mean by "cannot have a complete URL".

So, are we talking about three cases: FQDN (expected), FQDN with https (handled), and complete URL (?), or just the first two? I'm not sure what the intended meaning of complete URL is, because --- in a sense --- `https://landscape-server.domain.com` is *complete* and it does get handled.

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
...
+More complex URLs will be rejected and an error message will be shown.
```
edibotopic, line 73:
```suggestion
More complex URLs will be rejected and an error message will be shown.
```

Meaning of "more complex" is a little ambiguous but perhaps the possibility space is too wide to specify further?

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
...
 Refer to the section on [Landscape client configuration](howto::config-landscape-client) for an example.
 
 When you continue a status screen will appear confirming that configuration is complete:
 
 ![Configuration is complete](../assets/status-complete.png)
 
+The application waits a fraction of second to confirm that the configuration
+data supplied resulted in a successful connection to the Landscape server. In
+case of errors a dialog presents the error details and lets you decide to edit
+the configuration and try again or simply proceed.
```
edibotopic, line 85:
```suggestion
case of errors, a dialog presents the error details and lets you decide whether to edit
the configuration and try again, or proceed with the configuration.
```

docs/howto/set-up-landscape-client.md
```diff
@@ -109,6 +124,12 @@ ssl_public_key = C:\Users\user\Downloads\landscape_server.pem
 ```
 
 ```{warning}
+The `url` field in the `[host]` section cannot be a true URL, but rather a fully
```
edibotopic, line 127:
```suggestion
The `url` field in the `[host]` section cannot be a true URL, but rather a fully
```

We've talked about "complete URLs", "more complex URLs", and now "true URLs".
We should try to be consistent with the language used to avoid confusion.

Unless I'm the one overcomplicating things :sweat_smile: , it's probably easiest and most clear to affirm what is required (an FQDN, or FQDN with port) rather than what is not accepted (a complex/true/complete URL).

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
...
+a URL like `https://landscape-server.domain.com` but the `https://` part is
+removed from the address, resulting in the FQDN only. That's for allowing
+users to copy and paste the server address from a web browser address bar, for
+example.
+More complex URLs will be rejected and an error message will be shown.
+
+The "Advanced configuration" option requires you to specify a `landscape.conf`.
 Refer to the section on [Landscape client configuration](howto::config-landscape-client) for an example.
 
 When you continue a status screen will appear confirming that configuration is complete:
```
edibotopic, line 78:
```suggestion
When you continue, a status screen will appear confirming that configuration is complete:
```

Some of the content in this page, like this, should arguably be self-evident for someone navigating the UX, but that's something to consider at a future time.

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
...
 
 ![Configuration is complete](../assets/status-complete.png)
 
+The application waits a fraction of second to confirm that the configuration
+data supplied resulted in a successful connection to the Landscape server. In
+case of errors a dialog presents the error details and lets you decide to edit
```
edibotopic, line 84:
```suggestion
case of errors, a dialog presents the error details and lets you decide to edit
```

docs/howto/set-up-landscape-client.md
```diff
@@ -109,6 +124,12 @@ ssl_public_key = C:\Users\user\Downloads\landscape_server.pem
 ```
 
 ```{warning}
+The `url` field in the `[host]` section cannot be a true URL, but rather a fully
+qualified domain name (or IP address if your server was configured in that way)
+suffixed by a port (6554 by default) separated by a colon (`:`),
```
edibotopic, line 129: Just to be clear, is the port only required when going through the registry or the advanced config, but not through the manual config?

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,28 @@ In the UP4W app navigate to the Landscape configuration screen:
 
 Choose your preferred configuration option and enter the required details.
 
-The "Advanced Configuration" option requires you to specify a `landscape.conf`.
+The "Manual configuration" is easier if the server was configured with the
+default options, only the "Landscape FQDN" field is trully required.
```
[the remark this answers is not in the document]
CarlosNihelton: Also,. I'd rather keep the `The "Manual configuration" ` part because it matches the radio button shown in the UI.

docs/howto/set-up-landscape-client.md
CarlosNihelton: "Richer URLs with paths, queries and fragments" is more precise.

docs/howto/set-up-landscape-client.md
CarlosNihelton: Yeah, it's been confusing so far, but here we can be more precise: in the configuration manually crafted by the user (the `.conf` file) `[host].url` is not an URL at all, it's `FQDN:PORT`. The name is quite misleading.

If you write a string like `landscape.server.com:6554` into any URL parsing library I'm aware of it will parse it as an URL object with `scheme=landscape.server.com; path=6554`

This example is with Dart but I saw the same behaviour in Python, Go and C++, for example

image

So we need to emphasize it here that it's just `FQDN:PORT`, not a URL in my understanding. I've seen a customer used to Landscape write a URL in that field, like `https://landscape-etc.whatever.com`. That won't work. That's why I'd like to emphasize to not write a URL in there, despite the field name. :(

docs/howto/set-up-landscape-client.md
CarlosNihelton: That's right. In those cases the user is writing the configuration by hand. In the "manual config" case the UI does it for the user, so it applies the default port.

docs/howto/set-up-landscape-client.md
CarlosNihelton: Let's see if my proposed changes make this clearer. I wish this field was named something else other than `url`. I brought this conceptual question to the Landscape team. We might end up renaming it eventually.

docs/howto/set-up-landscape-client.md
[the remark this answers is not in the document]
edibotopic: OK.

The change about the Manual configuration reflects the principle that we should let the UI speak for itself; however, I won't fight you on it and when things are sufficiently complex it's sometimes OK to do this.

Added phrase seems good.

docs/howto/set-up-landscape-client.md
edibotopic: Got it. It's out of our hands to an extent.

docs/howto/set-up-landscape-client.md
edibotopic: Great.

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,30 @@ In the UP4W app navigate to the Landscape configuration screen:
 
 Choose your preferred configuration option and enter the required details.
 
-The "Advanced Configuration" option requires you to specify a `landscape.conf`.
+The "Manual configuration" is easier if the server was configured with the
+default options (you should double check with system administrator on that), for
```
edibotopic, line 66:
```suggestion
default options (if in doubt, check this with your system administrator), in
```

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,30 @@ In the UP4W app navigate to the Landscape configuration screen:
...
+which case only the "Landscape FQDN" field is truly required. This field does
+not accept a complete URL for the server (with paths and queries, for example),
+and a FQDN should always be used. The field accepts URLs like
+`https://landscape-server.domain.com` (with just scheme and host name) but the
```
edibotopic, line 70:
```suggestion
`https://landscape-server.domain.com` (with just a scheme and a host name) but the
```

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,30 @@ In the UP4W app navigate to the Landscape configuration screen:
...
+`https://` part (the scheme) is removed from the address, resulting in the FQDN
+only. This is for user convenience, as it allows you to copy and paste server
+addresses from a web browser address bar, for example. Richer URLs with queries,
```
edibotopic, line 73:
```suggestion
only. This is for user convenience, as it allows you to copy an address for a server from
a web browser address bar, for example, and paste it into the field. Richer URLs with queries,
```

docs/howto/set-up-landscape-client.md
```diff
@@ -62,13 +62,30 @@ In the UP4W app navigate to the Landscape configuration screen:
...
+paths and fragments will rejected and an error message will be shown.
```
edibotopic, line 74:
```suggestion
paths and fragments will be rejected and an error message will be shown.
```

edibotopic (approved):
Pre-approving.

Added some additional fixes and minor suggestions

CarlosNihelton:
Rebasing over the latest main to allow CI to pass. No contents changed.

PR #1322, opened 2025-09-03
fix(docs): Note about PowerShell version required

cnihelton:
To follow the howto/custom-rootfs-multiple-targets.md.

The rest of the documentation work on either PowerShell 5.1 or 7.0 and later without changes AFAICT, but this one is special because it assumes the `-Authentication` parameter is available for the `Invoke-WebRequest` cmdlet, which is only the case on PowerShell 6 (no longer supported) and later.

I'd add a note somewhere saying that throughout the documentation we assume PowerShell 5.1 or later, but I don't know where to place it 😅 . If you have recommendations I'm happy to implement them.

---

UDENG-7639

### DIFF
--- a/docs/howto/custom-rootfs-multiple-targets.md
+++ b/docs/howto/custom-rootfs-multiple-targets.md
@@ -27,7 +27,7 @@ To follow the steps outlined in this guide you can use either:
 - Familiarity with Bash and/or PowerShell.
 
 ```{note}
-Throughout this guide we assume PowerShell 7.0 or later. Some commands will fail on PowerShell 5.
+You need PowerShell 7.0 or later to follow this guide. Some commands will fail on PowerShell 5.
 You can check your PowerShell version by running the command `$PSVersionTable`.
 ```
 


CarlosNihelton:
Link check won't pass today: we're facing a major outage on multiple ubuntu.com endpoints. We can try again tomorrow.

edibotopic (approved):
One minor comment on phrasing of the note. Pre-approving.

I don't think we need to apply the prereq version of PS to all pages in this PR, but I've added a commented to discuss the issue.

CarlosNihelton:
> I think that the most important thing is to include the note on this page.
> 
> For the others, it is probably safe to assume that most users are on 5.1 or 7+.
> 
> If we wanted to be clear about that, then we could mention it on the prerequisites of the affected pages, for example:
> 
> ```diff
> ## What you will need
> 
> - Windows 10 or 11 running on either a physical device or virtual machine
> + Windows 10 or 11 with PowerShell 5.1 or 7+
> ```
> 
> I'm combining things here 😆 , but I'm not sure if we should be recommending a Windows VM for testing WSL features, and it might be more important to stress the PS point.

Running VMs might be problematic if nested virtualization cannot be enabled for instance. They mostly work, but incur in higher requirements for the user devices, so we should not promote VMs as ways to run WSL in general.
Now a guide like this one just hitting powershell commands are OK for VMs without assuming a lot about the user's devices computing power.

PR #1333, opened 2025-09-09
fix(windows-agent): WIP

CarlosNihelton:


CarlosNihelton:
> I haven't looked at this PR too closely, but I do see that in tests, you're waiting x amount of time to see if an error happens. This is a pattern I've used before but then found to be unreliable and also costs longer test runs.
> 
> Without looking at the code too closely, have you considered maybe having the mock expose a context or channel for tests that allows your test to determine if your mock has reached an okay state? It should be a lot more reliable than guessing with timings. It isn't always viable, but it could be worth a shot.

I'm parking a task for the next sprint to rework this mock entirely. It was a great aid back in the day Landscape couldn't talk to us at all, but its complexity is no longer as helpful as it was before.

PR #1334, opened 2025-09-09
ci: Avoid installing locales and man pages

adombeck:
Speed up CI jobs by avoiding installation of locales and man pages.

The major performance gain that can be expected is that it makes the "Processing triggers for man-db" step a no-op. That step can otherwise take more than a minute, for example in [this recent run](https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/17560477497/job/49875337391#step:3:943) it took 1m 39s:

image

UDENG-7865


CarlosNihelton (approved):
Thanks for sharing the improvement!

PR #1352, opened 2025-09-12
feat(ci): Enable clang-tidy and TiCS

cnihelton:
Enabling TiCS for this project has been challenging for many reasons, being the final blocker the need to run `clang-tidy` to perform static code analysis on the few C++ source files on Windows spread in three different areas:
- `storeapi.dll` used by the agent to interface with the MS Store runtime;
- `ubuntu-pro-agent-launcher` which implements a pseudoconsole to host the agent and;
- the Flutter GUI - Dart code is compiled to a shared library hosted by a native C++ application. Besides, it also talks to the MS Store runtime and some other native APIs, so there is more than just the Flutter default C++ embedder code.

All very Windows specific code.

Integrating `clang-tidy` SCA in any C++ project is easy as long as the build system provides a compilation database, the famous `compile_commands.json` file. Indeed, since TiCS runs `clang-tidy` on its own we don't even need to run the linter in CI for example, we just need to provide the compilation database. Creating that file is well supported via CMake with ninja and make generators (not Visual Studio generators because MSBuild doesn't have a nice generic way to do it). We are full native Windows tooling, not an easy environment to integrate it with. 

But there are ways. By setting a few project properties it's possible to run a build with the property `RunCodeAnalysis` set to true (vla msbuild CLI at build time), provided that the projects are themselves configured for running `clang-tidy` instead (or in addition to) the MSVC own's static analysis. For the actual C++ projects you'll see me doing something like this:

```xml
  <PropertyGroup Condition=...>
    <EnableClangTidyCodeAnalysis>true</EnableClangTidyCodeAnalysis>
    <ClangTidyToolExeAdditionalOptions>--config-file=$(SolutionDir)..\.clang-tidy</ClangTidyToolExeAdditionalOptions>
    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
  </PropertyGroup>
```

(disabling MSVC code analysis and enabling clang-tidy passing a certain config file).

With that approach we make MSBuild create the compilation database and run clang-tidy in  CI during package build. As said before, running the linter could be skipped, but I'll make it useful later on, so let's keep it.

The Flutter part is trickier. `flutter` is a build tool that drives not only the Dart part of the code, but also drives the native C++ part, which uses CMake as a meta-build system, which in turn uses MSBuild as the build system (and forcefully MSVC as the compiler). The way to make an MSBuild property cross all those layers of indirection is via an environment variable. That's precisely what's happening in the gui.targets file:

```diff
define="UP4W_FULL_VERSION=$(UP4W_FULL_VERSION)"</VersionDefine>
+		<ClangTidyEnv>ClangTidyToolExeAdditionalOptions="--config-file=$(SolutionDir)..\.clang-tidy";EnableClangTidyCodeAnalysis=$(RunCodeAnalysis);RunCodeAnalysis=$(RunCodeAnalysis);EnableMicrosoftCodeAnalysis=false</ClangTidyEnv>
	</PropertyGroup>
	<ItemGroup>
		<AppIcon Include="$(AppIconPath)" />
@@ -18,7 +19,7 @@
		<Message Text="Building Flutter artifacts to $(AppOutDir)" Importance="high"/>
        <Copy SourceFiles="@(MsixAppIcon)" DestinationFiles="@(AppIcon)" />
		<Exec Command="flutter pub get" WorkingDirectory="$(AppDir)" />
-		<Exec Command="flutter build windows --$(AppConfiguration) $(OverrideEntryPoint) $(VersionDefine)" WorkingDirectory="$(AppDir)" />
+		<Exec Command="flutter build windows --$(AppConfiguration) $(OverrideEntryPoint) $(VersionDefine)" WorkingDirectory="$(AppDir)" EnvironmentVariables="$(ClangTidyEnv)" />
		<ItemGroup>
			<AppContent Include="$(AppOutDir)**\*.*" Visible="False" />
			<None Include="@(AppContent)" Visible="False">
```

When building the MSIX package, the `gui.targets` file tells MSBuild to run `flutter build windows`, and now we're passing a few environment variables named exactly like the MSBuild properties related to code analysis and clang-tidy, so when flutter runs CMake and CMake runs MSBuild, those values are picked up as properties and the leaf most build tool now knows what to do.

That's what 86759e1f98c219aef0306373723b3e385724fdb3 is about.

But life is not that simple. C++ is a complex language, one cannot lint it properly without being a C++ compiler, so `clang-tidy` is quite like the `clang` compiler frontend, which means it's not MSVC. But this project is compiled with MSVC, all `#defines`, compiler flags etc are for MSVC so it is the standard library headers. If C++20 or later is enabled, the `<compare>` standard header (included by the `<variant>` standard header included in `flutter/encodable_value.hpp` Flutter's specific header file) does a funny circular `requires` clause that breaks clang. `// NOLINT` is not enough here because it actually fails to compile (parse the code till the abstract syntax tree phase, better said). The fix is implemented in [LLVM 21](https://godbolt.org/z/7zEWYYjP9) but Visual Studio 2022 ships LLVM 19. That means we cannot run clang-tidy against the Flutter part of the project during package build. But we still want an easy way for TiCS to jump in and do its stuff, which only depends on the compilation database, not actually us running clang-tidy for them. So we need MSBuild doing all the heavy lifting but skipping clang-tidy run.

That's what e2a180d9333740fbcc5616960f94c82b407854e8 is for. Using the environment variables/MSBuild properties `ClangTidyToolPath` and `ClangTidyToolExe` we avoid running clang-tidy specifically on the Flutter side and display a nice reminder for our future selves once LLVM 21 is part of Visual Studio.

As part of onboarding the project, the TiCS team became aware of the paths where they would find the `compile_commands.json` files generated by MSBuild for the multiple subprojects involved as well as the command needed to make MSBuild do all of that. So if my understanding is correct only the changes in the project files are really necessary.

The final commit of this submission (8bf2a28abe6d745ce3f0c2aba816a675d7a39129) is to actually integrate with TiCS.
I'm taking the existing job that uploads coverage information to codecov, and making it run in one of those self-hosted runner prepared for TiCS. Building on the experience of ubuntu-insights, I'm only doing it in pushes to main and at a weekly schedule. The project is already visible in the production dashboard, not exactly due this PR but due to the changes herein introduced into the project files, enough to generate the compilation database and allow TiCS to complete their job.

It's worth noting that the integration is not yet perfect, there are files not included in test coverage report and a few compilation issues affecting clang-tidy capacity to lint some source files, but at least we got a decent amount of code analysis by now.

---

UDENG-7034

### DIFF
--- a/msix/agent/agent.vcxproj
+++ b/msix/agent/agent.vcxproj
@@ -66,10 +66,7 @@
   <PropertyGroup>
     <RunCodeAnalysis>false</RunCodeAnalysis>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+  <PropertyGroup>
     <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
--- a/msix/gui/gui.vcxproj
+++ b/msix/gui/gui.vcxproj
@@ -50,10 +50,7 @@
     <Import Project="$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props" Condition="exists('$(UserRootDir)\Microsoft.Cpp.$(Platform).user.props')" Label="LocalAppDataPlatform" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
-    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+  <PropertyGroup>
     <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
   </PropertyGroup>
   <ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">


storeapi/base/impl/StoreContext.hpp
```diff
@@ -8,6 +8,7 @@
 #error This is Windows specific Store API Context implementation and cannot compile on other platforms.
 #endif  // _MSC_VER
 
+#include <unknwn.h>
```
hk21702: I wonder what the story behind this being `unknwn.h` instead of `unknown.h` is...

msix/storeapi/storeapi.vcxproj
```diff
@@ -43,6 +43,16 @@
...
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <EnableClangTidyCodeAnalysis>true</EnableClangTidyCodeAnalysis>
+    <ClangTidyToolExeAdditionalOptions>--config-file=$(SolutionDir)..\.clang-tidy</ClangTidyToolExeAdditionalOptions>
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <EnableClangTidyCodeAnalysis>true</EnableClangTidyCodeAnalysis>
+    <ClangTidyToolExeAdditionalOptions>--config-file=$(SolutionDir)..\.clang-tidy</ClangTidyToolExeAdditionalOptions>
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
```
hk21702: Since there are three sub-fields and they're all the same, would it be clearer to use `Or` instead of having two groups?

msix/ubuntu-pro-agent-launcher/ubuntu-pro-agent-launcher.vcxproj
```diff
@@ -43,6 +43,16 @@
...
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <EnableClangTidyCodeAnalysis>true</EnableClangTidyCodeAnalysis>
+    <ClangTidyToolExeAdditionalOptions>--config-file=$(SolutionDir)..\.clang-tidy</ClangTidyToolExeAdditionalOptions>
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <EnableClangTidyCodeAnalysis>true</EnableClangTidyCodeAnalysis>
+    <ClangTidyToolExeAdditionalOptions>--config-file=$(SolutionDir)..\.clang-tidy</ClangTidyToolExeAdditionalOptions>
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
```
hk21702: Ditto

msix/gui/gui.targets
```diff
@@ -9,6 +9,7 @@
 		<AppConfiguration>$(Configuration.ToLower())</AppConfiguration>
 		<OverrideEntryPoint Condition="$(UP4W_END_TO_END) != '' And $(UP4W_END_TO_END) != 'false'">-t end_to_end/end_to_end_test.dart</OverrideEntryPoint>
 		<VersionDefine Condition="$(UP4W_FULL_VERSION) != '' ">--dart-define="UP4W_FULL_VERSION=$(UP4W_FULL_VERSION)"</VersionDefine>
+		<ClangTidyEnv>ClangTidyToolExeAdditionalOptions="--config-file=$(SolutionDir)..\.clang-tidy";EnableClangTidyCodeAnalysis=$(RunCodeAnalysis);RunCodeAnalysis=$(RunCodeAnalysis);EnableMicrosoftCodeAnalysis=false;ClangTidyToolPath=$(MSBuildProjectDirectory);ClangTidyToolExe=clang-tidy-not-yet.bat</ClangTidyEnv>
```
hk21702: `clang-tidy-not-yet.bat` isn't the clearest name. How about `mock-clang-tidy.bat` or `clang-tidy-todo.bat` or something along those lines?

.github/workflows/qa.yaml
```diff
@@ -273,30 +276,68 @@ jobs:
...
       - name: Check out repository
         uses: actions/checkout@v5
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: windows-agent/go.mod
       - name: Install dependencies
         run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt install -y dotnet-sdk-8.0
           dotnet tool install -g dotnet-reportgenerator-globaltool
+          go install honnef.co/go/tools/cmd/staticcheck@v0.4.7
```
hk21702: Would using a newer version be possible here? It doesn't appear that the newer versions would cause things to break.

https://github.com/dominikh/go-tools/releases

msix/agent/agent.vcxproj
```diff
@@ -63,6 +63,15 @@
...
   <PropertyGroup Label="UserMacros" />
+  <PropertyGroup>
+    <RunCodeAnalysis>false</RunCodeAnalysis>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <EnableMicrosoftCodeAnalysis>false</EnableMicrosoftCodeAnalysis>
+  </PropertyGroup>
```
hk21702: Do we need the conditional here as well then?

PR #1358, opened 2025-09-15
fix(gui): Flips GUI default settings

CarlosNihelton:
The desired behaviour for the GA is the following defaults:

- Purchase via MS Store enabled
- Landscape disabled.

That means that either in the absence of the registry key or any missing or erroneous value those must be the settings, only forced otherwise if:

- Purchase disabled forcefully by the value 0 in the registry value;
- Landscape forcefully enabled by the value 1 in the registry value.

Additionally we warn in the agent logs if the Landscape config is changed, as we don't completely disabled it, at least we inform the user they are touching unstable territory.

### DIFF
--- a/gui/packages/ubuntupro/lib/core/settings.dart
+++ b/gui/packages/ubuntupro/lib/core/settings.dart
@@ -13,15 +13,15 @@ class Settings {
     try {
       repository.load();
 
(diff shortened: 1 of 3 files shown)


ashuntu (approved):
LGTM!

PR #1367, opened 2025-09-18
docs(ubuntu-pro-for-wsl): Improved consistency of paths used in terminal prompts

KrownWealth:
As described in issue [#298](https://github.com/canonical/open-documentation-academy/issues/298),
I updated the [Ubuntu on WSL documentation](https://documentation.ubuntu.com/wsl/latest/) to use consistent path names across terminal prompts.

This improves readability and reduces confusion for users following the WSL setup and usage guides.


KrownWealth:
@edibotopic please review

edibotopic:
Thanks @KrownWealth --- there's something funny going on with this PR.

Why is `.sphinx/venv` being added to the root of the repo?
There is already `docs/.sphinx` and a virtual environment shouldn't be committed.

Could you please check? You likely need to drop the `added .sphinx` commit.

---

In terms of the linkcheck, there are no broken deps that I am aware of. However, we do have some failing links to Ubuntu pages that are down currently, which can't be avoided.

KrownWealth:
> Thanks @KrownWealth --- there's something funny going on with this PR.
> 
> Why is `.sphinx/venv` being added to the root of the repo? There is already `docs/.sphinx` and a virtual environment shouldn't be committed.
> 
> Could you please check? You likely need to drop the `added .sphinx` commit.
> 
> In terms of the linkcheck, there are no broken deps that I am aware of. However, we do have some failing links to Ubuntu pages that are down currently, which can't be avoided.

I'm looking into this. Would push a fix that drop the added .sphinx on the root.

KrownWealth:
> Please do @KrownWealth --- I will wait to review because currently there are 121 files listed as being changed, when I believe it should be a dozen or so.
> 
> Take your time.



> @KrownWealth --- the CI doesn't seem to be picking up the `requirements.txt` file, which is causing the failures, even though it does exist. We're not having that issue with other PRs, so I'm not sure what's happening.
> 
> I notice that the PR seems to be modifying the requirements file, but I'm not sure why? I'll need to look into this, but if you can provide any context that might help I would be grateful.
> 
> edit: I checked your branch locally and the symbolic link to `.sphinx/requirements.txt` seems to have been broken.
> 
> Running the following:
> 
> ```shell
> file requirements.txt
> ```
> 
> Outputs:
> 
> ```
> requirements.txt: broken symbolic link to .sphinx/requirements.txt\012
> ```
> 
> In your branch, the path seems to have had a newline character (\012) appended to the path, which has broken it.
> 
> The command should output the following for a working symbolic link:
> 
> ```
> requirements.txt: symbolic link to .sphinx/requirements.txt
> ```
> 
> I'm not sure how that happened, but the symbolic link needs to be fixed for the CI to pass.

I think I made a fix. 

So what happended was that, requirements.txt was inside /docs instead of .sphinx/

edibotopic:
OK. Checks are now passing :+1: 

We've done a lot of work already and I haven't even started the review :smiley: . I'll get to that next.

In the meantime, do you know what happened the symbolic link?

> So what happended was that, requirements.txt was inside /docs instead of .sphinx/

The requirements.txt is in .sphinx but there is also a symbolic link in docs/. This is intentional and doesn't normally cause issues. There has only been an issue with this PR, so I'm curious what happened.

KrownWealth:
> @KrownWealth --- please let me know if you need to discuss further. I can review again when you're finished on this second pass.

I updated oter examples that uses u@mib in the docs. Could you please help re-review.

KrownWealth:
> @KrownWealth --- there are some comments that were resolved but I'm not sure if they were fully addressed. I mention an example [above](https://github.com/canonical/ubuntu-pro-for-wsl/pull/1367#discussion_r2378864866), in relation to the use of captions to show paths where the path is needed to indicate where a command needs to be run. Could you please check that you have applied the relevant changes before resolving comments, or start a discussion if a comment is unclear or problematic.
> 
> I can also still find several cases in the documentation where pages haven't been updated with the new wording for prompts; for example, the cloud init guide still uses `mib` as a hostname, `pc` is used in the installation guide, and `your-machine` is used in the customise an ubuntu distro guide. Could you please check through the documentation for any remaining examples that have not been updated.
> 
> Lastly, some errors seem to have been introduced. For example, in the verify Pro subscription guide, commands have been prefixed with + signs, such as `+ $ pro status`, which is incorrect. These need to be fixed before the PR can be approved.


@edibotopic -- I would make sure to check thoroughly before resolving comments. I apologies.

Current resolution for the use of caption you commented on has been resolved. My new commit would carry that. However, I needed more clarity related to captions on other usecases within the docs, which seem unclear to me.

Makng use of consistent hostname, in some files such as [deployment](http://localhost:8000/tutorials/deployment/), I rename example like this  jdoe@hostname:~$` 

While on same file, a codeblock comes like this where <> comes in. I wanted to get the ideal correction here.

```{code-block} text
:class: no-copy
username@<hostname>:~$
```

edibotopic:
> However, I needed more clarity related to captions on other usecases within the docs, which seem unclear to me.

I think we can look at it this way:

* Is the path where a command is run **relevant** to the success of the command? If not, don't include the path. For example, running `wsl --install` or `wsl --unregister` can be done from any directory. In that case, including the path may be confusing. If in doubt, it is useful to test these cases (or ask).
* If the path **is** relevant to the command, then include it. In that case, I think a caption is usually a clean way to achieve this. This also makes it easier to use `<placeholders>` for the PowerShell paths because we avoid having double `>>`.
* Are we showing the user the prompt and asking them to check their own prompt for **comparison**? Then we probably shouldn't use a caption and just show the prompt.
* In most cases, we want to use the `< >` angle brackets around username and hostname, because the user needs to know that they are placeholders that will look different on their machine. There are exceptions where it is probably fine (e.g., we are just giving an example of how PowerShell and Ubuntu prompts are different)

Many cases are clearcut, but some require looking at the context of the page. For example, in the [deployment tutorial (current published version)](https://documentation.ubuntu.com/wsl/latest/tutorials/deployment/#), there is currently this example:

```
u@<your-machine>:~$
```

The `u` here is OK in the context of that tutorial, because the tutorial describes a configuration file that configures the user to be `u`. So if the reader follows the instructions on their machine, the user will be `u` on their machine. So we shouldn't just change it to `<username>`.

In contrast, `<your-machine>` is just a placeholder, it will be different on the user machine, and can be replaced with a standard placeholder like `<hostname>`.

This is also a good example of where we should not use a caption, because the sentence right before reads:

> ...confirm that the correct default user “u” has been set from the prompt:

So the reader is being asked to actually check the prompt that they see and compare it to the one we show in the docs.

In general:

* Use captions if we need to specify a path (e.g., an import command that references a file in a specific directory)
* Don't specify a path if it's not necessary (e.g., a command that can be run anywhere)
* Use angle brackets for placeholders (e.g., a hostname that will be different for the user)

But, you need to be careful about applying these rules without considering the context of a specific page (e.g., a user name that we are directing the user to configure).

edibotopic:
> Understood.

OK @KrownWealth --- I look forward to reviewing the next version.

edibotopic:
@KrownWealth I fixed up some remaining issues in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1382 and merged the changes.

Thanks for your contribution.

PR #1371, opened 2025-09-22
docs: clarifications relating to WSL versions

edibotopic:
* Uses the convention of having a space between WSL and `<version number>`, e.g., "WSL 2", which is more common than "WSL2"
* Adds notes to tutorials and guides emphasising that WSL 2 is required
* Links added to new WSL versions explanation where appropriate

UDENG-7253


hk21702 (approved):
Looks good to me; doesn't seem like anything was missed.

PR #1372, opened 2025-09-22
docs: clarify Windows requirements

edibotopic:
* Updates pre-requisites in several guides to explicitly recommend Windows 11 and provide a minimum Windows 10 version
* Updates uninstall guide to avoid recommending user search through Windows Start, to account for different flows based on distribution format, and to recommend terminal-based WSL uninstall

UDENG-7625


docs/howto/install-ubuntu-wsl2.md
```diff
@@ -16,8 +16,7 @@ myst:
 
 ## What you will need
 
-* Windows 10 or 11 running on a Windows machine
-* All of the latest Windows updates installed
+* Windows 11 (recommended) or Windows 10 with minimum version 21H2 on a physical machine
```
hk21702, line 19: It may be worth mentioning that virtualization features may need to be enabled. I'm not entirely certain about the requirements there though. But, this may be hardware dependent and require mucking around in the BIOS.

I would also consider that we clarify that WSL does work on a VM, it is just that nested virutalization may need to be enabled, so maybe something like "... on a physical machine with virtulzation enabled, or a virtual machine with nested virtualization enabled."

For nested virutalization, for Windows we can link to https://learn.microsoft.com/en-us/windows/wsl/faq#can-i-run-wsl-2-in-a-virtual-machine- which has the Hyper-v command.

docs/howto/uninstalling.md
```diff
@@ -7,39 +7,68 @@ myst:
...
+## Ubuntu Pro for WSL
 
-In the Windows Start Menu, locate the "Ubuntu Pro for WSL" application and right-click on it, then click **Uninstall**.
+Go to `Settings > Apps > Installed Apps`, locate the "Ubuntu Pro for WSL"
+application, right-click on it and select **Uninstall**.
 
-![Uninstall Ubuntu Pro for WSL](./assets/start-menu-uninstall.png)
-
-You should also remove the `.ubuntupro` directory from your Windows user profile directory.
+You should also remove `.ubuntupro` from your Windows user profile directory.
```
hk21702, line 19: I wonder if it would be more intutive to use `%USERPROFILE\.ubuntupro\` when referencing this kind of thing. I'll leave it up to you though.

hk21702 (changes requested):
Contents looks good overall, but I have a few suggestions.

I also wonder if it would be better to specify that the physical machine needs virtualization features enabled (something that may require messing with BIOS). 

Was there a reason why we don't mention VMs with nested virtualization as an option? It might be worth mentioning that given how with correct setups, the steps should be identical. We'd just need to point out that you need nested virtualization (though I'm not sure if there are certain UP4W features that don't work in a VM).

We are repeating the WSL requirements quite a bit though. It might be worth having a dedicated page on it, even if it is quite short, and then embedding it in pages if that is something possible? This would just be to standardize things and make it so that if it needs to be changed, there's only one place to modify. (Ofc, for extra requirements we'd append it on for the specific page).

docs/howto/gpu-cuda.md
```diff
@@ -21,10 +21,9 @@ The following steps assume a specific hardware configuration. Although the conce
 
 Make sure the following prerequisites are met before moving forward:
 
-
-* A Windows 10 version 21H2 or newer physical machine equipped with an NVIDIA graphics card and administrative permission to be able to install device drivers
+* Windows 11 (recommended) or Windows 10 with minimum version 21H2 on a physical machine
+* An NVIDIA graphics card and administrative permission to be able to install device drivers
```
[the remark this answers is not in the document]
edibotopic: Done in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1372/commits/2318e298b9b6b8817c9cd42f4a58e54daa2fd9ab

docs/howto/uninstalling.md
edibotopic: I think that we should move to using %USERPROFILE% more across the documentation, but it's a future task.

hk21702 (approved):
LGTM with the caveats of revisiting a few points in the future. I do still think it's worth mentioning that virtualization needs to be enabled on the physical machine, but it might be something to look into in the future considering how it can get complex. I think it is mentioned in the MS documentation though.

edibotopic:
> LGTM with the caveats of revisiting a few points in the future. I do still think it's worth mentioning that virtualization needs to be enabled on the physical machine, but it might be something to look into in the future considering how it can get complex. I think it is mentioned in the MS documentation though.

Definitely something to look into. I will be working on some tweaks for the install instructions shortly, so there might be an opportunity then.

PR #1374, opened 2025-09-24
fix(docs): add mount point cleanup section

miketheitguy:
Did some additional research and investigation here. Part of the issue appears to be that mounts are shared by default, which causes issues with unmounting the mount points afterwards, showing the mount point is busy, etc.

So this modifies the mount points to be read only in the chroot and allows us to unmount them further down the pipe.


CarlosNihelton:
Manually merged via commit https://github.com/canonical/ubuntu-pro-for-wsl/commit/43283a34d68fa6088b417f069c66c937731ebc73 to fix the git history.

PR #1384, opened 2025-10-03
feat(windows-agent): Report unmanaged instances to the Landscape server

CarlosNihelton:
The spec WS043 deals with augmenting the Windows agent to report to the Landscape server about WSL instances not managed by this system. In that spec we restricted the information to only touch Ubuntu instances. Landscape's ultimate objective is to eventually send commands to remove such instances if the system administrator desires.

This PR implements a method in the agent's database to report the instances considered unmanaged (Ubuntu instances that didn't present themselves to the agent - what is done by the wsl-pro-service), which is later added to the hostagent info message the agent eventually sends to the server. The Landscape hostagent API dependency is updated to include that field. The list of unmanaged instances is a best effort snapshot, we don't cache that data for later comparison. To acquire information about the unmanaged instances we go rather invasively:
- We run the `hostname` command and
- We read the instance's `os-release` file from the Windows side, via its complete UNC path, such as `\\wsl.localhost\<DISTRO>\etc\os-release`.

The GoWSL dependency was also upgraded to cope with the need to launch the hostname command, which wasn't previously supported in the mocked backend.

---

UDENG-7822

---

Side note: the code coverage could get better if the GoWSL mock backend offered a way to predefine state for distros when queried, but implementing that is rather outside of the scope of this task.

### DIFF
--- a/windows-agent/internal/distros/database/database_test.go
+++ b/windows-agent/internal/distros/database/database_test.go
@@ -139,6 +139,88 @@ func TestDatabaseGetAll(t *testing.T) {
 	}
 }
 
+func TestDatabaseGetUnmanaged(t *testing.T) {
+	ctx := context.Background()
+	uncRoot := t.TempDir()
+
+	if wsl.MockAvailable() {
+		t.Parallel()
+		ctx = context.WithValue(wsl.WithMock(ctx, wslmock.New()), database.UNCRootKey, uncRoot)
+	}
+
+	// Registers some Ubuntu instances
+	var distros []string
+	for range 3 {
+		distros = append(distros, func() string {
+			d, _ := wsltestutils.RegisterDistro(t, ctx, false)
+			d = strings.ToLower(d)
+			writeOsRelease(t, uncRoot, d, "ubuntu-os-release")
+			return d
+		}())
+	}
+
+	// Register one non-Ubuntu instance.
+	nonUbuntu, _ := wsltestutils.RegisterDistro(t, ctx, false)
+	nonUbuntu = strings.ToLower(nonUbuntu)
+	writeOsRelease(t, uncRoot, nonUbuntu, "other-os-release")
+
+	testCases := map[string]struct {
+		dbDistros []string
+
+		want []string
+	}{
+		"empty database":            {want: distros},
+		"database with one entry":   {dbDistros: []string{distros[0]}, want: distros[1:]},
+		"database with two entries": {dbDistros: distros[0:2], want: []string{distros[2]}},
+		"no unmanaged distros":      {dbDistros: distros, want: []string{}},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			if wsl.MockAvailable() {
+				t.Parallel()
+			}
+
+			db, err := database.New(ctx, t.TempDir())
+			require.NoError(t, err, "Setup: database creation should not fail")
+			defer db.Close(ctx)
+
+			for i := range tc.dbDistros {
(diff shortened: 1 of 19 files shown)


windows-agent/internal/distros/database/unmanaged.go
```diff
@@ -0,0 +1,146 @@
...
+	Name      string      // The distro instance name
+	GUID      string      // The WSL internal GUID
+	DistroID  string      // Same as /etc/os-relase ID
+	VersionID string      // Same as /etc/os-relase VERSION_ID
+	Hostname  string      // Instance hostname
+	State     gowsl.State // Stopped, Running, Uninstalling etc
+}
+
+// GetUnmanagedDistros returns a list of Ubuntu instances that are currently registered in WSL but not managed by this agent.
+// That's an expensive operation that involves a light wake up of all instances not previously discovered.
```
hk21702, line 30:
```suggestion
// This is an expensive operation that involves a light wake up of all instances not previously discovered.
```

windows-agent/internal/distros/database/unmanaged.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return BasicDistroInfo{}, fmt.Errorf("failed to get distro basic info for %q: %v", d.Name(), err)
+	}
+
+	if osInfo.Id != "ubuntu" {
+		// Our business only concerns with Ubuntu instances.
+		return BasicDistroInfo{}, fmt.Errorf("skipping non-Ubuntu distro instance %q", d.Name())
+	}
+
+	hostname, err := d.Command(ctx, "hostname").CombinedOutput()
```
hk21702, line 86: Is there a reason we'd want to capture from stderr as well? It doesn't seem like the output is being logged in the error case anyway. Also, it might not be even possible, but I'd imagine an output to stderr like a warning, but exit code 0, that could cause some issues.

windows-agent/internal/distros/database/unmanaged.go
```diff
@@ -0,0 +1,146 @@
...
+	if err != nil {
+		return BasicDistroInfo{}, fmt.Errorf("failed to get state of distro %q: %v", d.Name(), err)
+	}
+
+	osInfo, err := readOsRelease(filepath.Join(uncRoot, d.Name()))
+	if err != nil {
+		return BasicDistroInfo{}, fmt.Errorf("failed to get distro basic info for %q: %v", d.Name(), err)
+	}
+
+	if osInfo.Id != "ubuntu" {
```
hk21702, line 81: For safety and future headaches, would it be be reasonable to make the comparison case insensitive?

hk21702 (changes requested):
Looks okay overall, just a few nitpicks and questions!

windows-agent/internal/distros/database/database_test.go
```diff
@@ -139,6 +139,69 @@ func TestDatabaseGetAll(t *testing.T) {
...
+		"no unmanaged distros":      {dbDistros: distros, want: []string{}},
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			if wsl.MockAvailable() {
+				t.Parallel()
+			}
+
+			db, err := database.New(ctx, t.TempDir())
```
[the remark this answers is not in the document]
CarlosNihelton: That requires Go 1.24 (we're still at go 1.23 :p )

windows-agent/internal/distros/database/unmanaged.go
CarlosNihelton: Good point!

PR #1410, opened 2025-10-29
fix(agent): v1-only: Let's not create the LandscapeConfig registry key by default.

cnihelton:
Since v1 is meant to not advertise Landscape support, having us creating a seemingly unrelated registry key automatically can be confusing for the user. Let's not do it. Hopefully this will be the only code change between v1 and the main branch by the time we freeze for the release.

Looks like we left a broken cross-reference. Though it's unrelated to this PR I'm removing it now because it's a small enough change.


CarlosNihelton:
TiCS was not supposed to run on PRs, that's being adressed at #1411 .

edibotopic (approved):

PR #1412, opened 2025-10-29
fix(ci): No partial codecov uploads

cnihelton:
I remember relying solely on the Cobertura format for codecov was yielding subpar results because it didn't handle partial coverage in branches. Yet, that's not enough of a reason to preserve this step issuing partial uploads to codecov when we could just include the gocov binary format as part of the collected files from the test matrix.

So with this changeset I aim to include the filtered gocov binary format in the list of artifacts uploaded by the testing steps of the matrix and when recollecting all those artifacts we push both the Cobertura and the binary format at the same time.

That should stop those false alarms issued by codecov while CI is not yet finished.


CarlosNihelton:
Yeap, that was exactly the reason why we didn't stick with just Cobertura XML format. Now that the other files in the gocov binary format are being uploaded we're back into the correct situation with partial branch coverage reported:

image


That looks correct now.

hk21702 (approved):
LGTM

PR #1413, opened 2025-10-29
feat(gui): Landscape SaaS is no longer a mistake

cnihelton:
When this system was first conceived there was no plans to land support for it in Landscape SaaS.

Fast forward to the present the Landscape team is working hard to land that support eventually, so it's about time to remove the restrictions in the UI that prevented attempts to connect to that server in particular.

That comes with some added complexity though.

First we need to add a text field for the user to input the account name when connecting to Landscape SaaS.

Secondly, an account name is required when connecting to Landscape SaaS and it must not be 'standalone'.
When connecting to self-hosted Landscape, only 'standalone' is a valid account name.

We present that interlocking of the two fields by having the account name field disabled when the FQDN is not SaaS and internally preset to 'standalone', otherwise we let the user input the account name as they wish. Modifying the FQDN to any value other than SaaS automatically resets the account name field back to 'standalone'.

Here's how this interlocking works.

https://github.com/user-attachments/assets/f127c391-0dbb-4fb1-b67c-88f537003c07

I'm also taking advantage of this added complexity to migrate the data model tests to use golden text files.

---

UDENG-7649

### DIFF
--- a/gui/packages/ubuntupro/lib/l10n/app_en.arb
+++ b/gui/packages/ubuntupro/lib/l10n/app_en.arb
@@ -76,7 +76,7 @@
     "landscapeFQDNError": "Invalid FQDN. Expected a fully qualified domain name or IP address.",
     "landscapeFQDNSaaSError": "FQDN shouldn't be landscape.canonical.com.",
     "landscapeAccountNameLabel": "Account name",
-    "landscapeAccountNameError": "Account name cannot be empty nor \"{reservedAccount}\".",
+    "landscapeAccountNameError": "Account name cannot be empty or \"{reservedAccount}\".",
     "@landscapeAccountNameError": {
         "placeholders": {
             "reservedAccount": {


edibotopic:
Nice!

**Questions:**

The user enters `landscape.canonical.com`, and the `Account name` field becomes interactive.
Then when focused, the hint in the field is "standalone", but this would be an invalid input for SaaS, right?
Should it just stay as "Account name" to avoid confusion?

When a user switches to `not.landscape.canonical.com`, any previously entered SaaS account name is fixed and greyed out. If there was an accidental keystroke, this might be a small convenience because the user can recover their entered name quickly. However, it also makes it appear that the account name is still being used for the non-SaaS registration but cannot be edited.
I think it could be better if the switch (SaaS -> self-host) was accompanied by a restoration of the default placeholder, or alternatively the default value for self-hosted (`standalone`). Perhaps if switching back to SaaS, the most recent entry for account name could be restored (I understand this might add complexity).

**Suggestion:**

The description: "Register with your own Landscape server" should be updated to reflect that the SaaS option is now also available.

CarlosNihelton:
I think removing the placeholder/hint is the best way to go. Otherwise, I followed the rest of @edibotopic's suggestions and I think I found a simpler way to make it in the end.


[Screencast From 2025-11-05 15-00-41.webm](https://github.com/user-attachments/assets/1de7a2dc-20bf-4420-8e3e-40e0624e0823)

ashuntu (approved):
LGTM!

PR #1442, opened 2025-11-19
fix: Check for the correct cloud-init systemd unit

cnihelton:
Recent tests with the staging version of Landscape SaaS revealed that Resolute Racoon instances created due a Landscape install request would not correctly register with the server, resulting in the activity displayed as `Parent instance DESKTOP-551PQ9O reported success. Awaiting registration message from child instance.`, until the Windows user manually started the newly created instance.

image

That is a consequence of the `wsl-pro-service` attempting to sync with the wrong `cloud-init` `systemd` unit. The agent does the same, but the consequences on the agent side are less obvious. When running for the first time, the `wsl-pro-service` waits on `cloud-init` if it detects the unit is enabled. Until plucky the unit was named `cloud-init.service` , then removed and the closest semantics can be implemented by checking for the `cloud-init-local.service`.

The `wsl-setup` package offers a script that can be used to sync with `cloud-init` for the first run. We should use it instead of hardcoding this check everywhere. Unfortunately we need to maintain backwards compatibility so things will likely become less cleaner. For now I'm repeating the "script" that implements the check and sync in both the agent and `wsl-pro-service`, giving us freedom to change the implementation details of one without worrying about the other. We could for instance assume that `wsl-pro-service` will be running on an instance with the latest `wsl-setup` and ignore completely the fallback part of the "script" (though I'd rather not rely on too many assumptions -- not sure, I'm not totally sold on that).

Also, to make the agent less dependent on the fix landing in the archive, I'm adding a reboot step when we register instances upon a Landscape request. It guarantees that the `wsl-pro-service` will see the Landscape configuration dropped by cloud-init, amend it as needed and triggers a registration with the right parameters without depending on the user interaction. That makes the Pro for WSL app usable right away for continued testing with Landscape SaaS without extra dependencies or assumptions. With just those changes in place by the time the agent replies to the server with the completion the WSL instance already registered with the correct parameters, without fixing the `wsl-pro-service` side. In the short-term future we should remove that step again, as the `wsl-pro-service` will be surely running that step only after cloud-init finishes.

---

UDENG-8594

### DIFF
--- a/wsl-pro-service/internal/system/landscape.go
+++ b/wsl-pro-service/internal/system/landscape.go
@@ -53,7 +53,18 @@ func (s *System) EnsureValidLandscapeConfig(ctx context.Context) (err error) {
 
 func (s *System) syncWithCloudInit() {
 	log.Debug(context.Background(), "Checking cloud-init status")
-	cmd := exec.Command("bash", "-ec", "systemctl is-enabled --quiet cloud-init.service && cloud-init status --wait")
+	// Wait for cloud-init to finish if systemd and its service is enabled.
+	// Since plucky the wsl-setup package ships a small script created for that purpose.
+	// As we need to support back to focal, we keep a fallback if the script is not readable.
+	script := `
+if [ -r /usr/lib/wsl/wait-for-cloud-init ]; then
+  source /usr/lib/wsl/wait-for-cloud-init
+  exit 0
+fi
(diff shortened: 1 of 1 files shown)


hk21702 (approved):
Looks good overall, and I'm pre-approving. I just have a question as to if we could avoid restarting the entire WSL instance.

PR #1452, opened 2025-11-21
docs(reference/actions): Update reference for the checkout action

hk21702:
This PR updates the reference for the WSL GitHub action for checking out a code repository to include the new optional argument to pass a GitHub access token.

A description for the argument, security warning, and a usage example have been added regarding it.

### DIFF
--- a/docs/reference/actions.md
+++ b/docs/reference/actions.md
@@ -52,10 +52,12 @@ Its arguments are:
 - `distro`: an installed WSL distro. Write its name as it would appear on WSL. Read more: [Ubuntu WSL distributions](reference::distros)
 - `working-dir`: the path where the repository should be cloned. Set to `~` by default.
 - `submodules:`: whether to fetch sub-modules or not. False by default.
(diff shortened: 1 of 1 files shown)


hk21702:
I didn't revise any of the other existing content since it isn't improbable that we'll revisit the actions at some point eventually, possibly overhauling them.

hk21702:
CI failure seems unrelated, though one related to `chroot` on another page does need a fix.

docs/reference/actions.md
```diff
@@ -61,6 +66,7 @@ Example usage:
     distro: Ubuntu-20.04
     working-dir: /tmp/github/
     submodules: true
+    token: ${{ secrets.GITHUB_TOKEN }}
```
CarlosNihelton: Since you mention the token permissions, I think it'd be a good idea to demonstrate the token permissions being restricted at the job level.

CarlosNihelton: Both suggestions would work as well.

PR #1454, opened 2025-11-24
feat(docs): Outlines the use of cryptography in Pro for WSL

cnihelton:
Apart from the use of TLS, there is very little to say on crypto. We don't implement anything on our own, but rather rely on Go stdlib for all cryptographic features used in the app.

Thus with very few paragraphs we should be able to present:

* Description of how cryptographic technology is used in the product.
* Cryptographic technology used by the product.
* Cryptographic technology being exposed to the user for their use.
* Packages or technology providing cryptographic functionality.

We also have expectations about a secure decommissioning of the app as part of the SSDLC documentation. I just realized we didn't mention the possibility of leftovers in the Windows registry, thus I included a paragraph about that in the uninstalling how-to and reference it in the security-overview.

---

UDENG-8625

### DIFF
--- a/docs/explanation/security-overview.md
+++ b/docs/explanation/security-overview.md
@@ -273,7 +273,7 @@ discouraged for security reasons and unlikely to be useful in real scenarios, ot
 testing. For more information refer to [the Landscape configuration reference page](howto::config-landscape).
 
 All cryptographic functionalities used by Ubuntu Pro for WSL are provided by the Go standard
-library.
+library `crypto` and its supplementary `golang.org/x/crypto` modules.
 
 ### Decommissioning
 


docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
 
 > [Intune configuration options for WSL](https://learn.microsoft.com/en-us/windows/wsl/intune?source=recommendations)
 
+### Cryptography used by Ubuntu Pro for WSL
+
+Ubuntu Pro for WSL use of cryptography starts at installation time: an MSIX package can only be
```
edibotopic:
```suggestion
The use of cryptography in Ubuntu Pro for WSL starts at installation time: an MSIX package can only be
```

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+installed on Windows if its signed by a trusted source. Even when downloading the MSIX from GitHub
```
edibotopic:
```suggestion
installed on Windows if it's signed by a trusted source. Even when downloading the MSIX from GitHub
```

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+releases or from Canonical website, the binary package is signed by Microsoft Store, thus no
+security exceptions are required to allow installing the application.
```
edibotopic:
```suggestion
security exceptions are required when installing the application.
```

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+
+At runtime the Windows background agent communicates with the wsl-pro-service within the Ubuntu on
+WSL instances via mutual TLS. Even though the network is local, virtual and private, data in flux is
+always encrypted. During startup the agent creates self-signed certificates (removing previously
+existing ones) which are used to ensure data is encrypted both ways.
```
edibotopic:
```suggestion
always encrypted. During startup, the agent creates self-signed certificates (removing previously
existing ones) that are used to ensure data is encrypted in both directions.
```

I think "in both directions" might be more clear here, as "way" can sometimes be used in the sense of _direction_ or _method_.

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+
+Those features are not configurable by the user.
```
edibotopic:
```suggestion
These cryptographic features are not configurable by the user.
```

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+
+Both the Windows agent and the WSL instances use TLS when communicating with a Landscape server, by
+default relying on the operating system's certificate store. It's possible to configure a connection
+to use a custom provided certificate issued for that server or even plain TCP, although that's
+discouraged and unlikely to be useful in real scenarios other than some local testing. For more
+information refer to [the Landscape configuration reference page](howto::config-landscape).
+
+All cryptographic functionalities used by Ubuntu Pro for WSL are provided by the Go standard
+library.
```
edibotopic: Or a specific package in the standard library?

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+
+### Decommissioning
+
+To support remote configuration and for sharing of certain files between the system components, Pro
+for WSL relies on non-virtualised access to the Windows registry `HKCU` hive and the user's profile
+directory. For that reason just uninstalling the MSIX may leave behind unwanted data in both
```
edibotopic:
```suggestion
directory. For that reason, uninstalling the MSIX may leave behind unwanted data in both
```

docs/explanation/security-overview.md
```diff
@@ -252,6 +252,37 @@ Intune also supports policies for WSL, which include toggling the availability o
...
+always encrypted. During startup, the agent creates self-signed certificates (removing previously
+existing ones) that are used to ensure data is encrypted in both directions.
+
+These cryptographic features are not configurable by the user.
+
+Both the Windows agent and the WSL instances use TLS when communicating with a Landscape server, 
+relying on the operating system's certificate store by default. It's possible to configure a connection
+to use a custom certificate issued for that server or even plain TCP, although that's
+discouraged for security reasons and unlikely to be useful in real scenarios, other than local
+testing. For more information refer to [the Landscape configuration reference page](howto::config-landscape).
```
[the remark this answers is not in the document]
edibotopic: It probably wouldn't create too many problems if we kept it, but I think removing it is probably the right call, especially because we don't want to create any unnecessary confusion in a security section.

PR #1474, opened 2025-12-02
docs: replace msstore download links for app

edibotopic:
The MS Store link is not live yet.

We can point users to the WSL page on the Ubuntu website, which includes an install link.


edibotopic:
Thanks @CarlosNihelton --- you can tag when ready to update `stable`.

PR #1475, opened 2025-12-03
feat(ci): Integrates with OSV Scanner

cnihelton:
Following a mix of the two most basic setups described in the tool documentation: https://google.github.io/osv-scanner/github-action/ but including support for workflow_dispatch, behaving like the scheduled event, i.e. scanning and reporting about the entire repository as configured.

OSV Scanner supports multiple programming languages, from C to Rust, R to Python and Go to Dart, as well as many package ecossystems, such as Debian, Alpine, Java Uber etc.

The Go parts of this project are well covered by `govulncheck`, but we didn't have anything as good for Dart. I aim to fill this gap. Since a single run of OSV cannot deal with multiple languages, I'm focusing solely on the GUI's pubspec.lock file. BTW, `govulncheck` just alerted about the need to upgrade the Go toolchain, which I'm doing in this PR otherwise CI won't be happy.

If needed, we can add more steps calling the same action with different lockfiles. The default option (`--recursive`) finds `go.work` quite early and assumes Go. Interestingly, although I included the `workflow_dispatch` trigger, I cannot exercise it until this is merged into main.

### DIFF
--- a/end-to-end/manual_token_input_test.go
+++ b/end-to-end/manual_token_input_test.go
@@ -85,6 +85,8 @@ func TestManualTokenInput(t *testing.T) {
 				require.NoErrorf(t, err, "Setup: could not wake distro up: %v. %s", err, out)
 			}
 
+			// By now the agent should have initialized the registry
+			requireRegistryIsInitialized(t, []string{"UbuntuProToken", "LandscapeConfig"})
 			const maxTimeout = time.Minute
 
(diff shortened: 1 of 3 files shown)


CarlosNihelton:
> LGTM for Flutter. Unfortunate that it can't handle multiple languages at once.

That was a mistake on my end merging the branches in the wrong order. The Go code came from: #1478

PR #1478, opened 2025-12-04
fix(windows-agent): Registry access is still virtualized

cnihelton:
Although we set up a registry virtualization exception in the package manifest (to make it easy for users to find this software registry key and its values already created), the virtualization exception was not working as expected, the initial write to `HCKU:\Software\Canonical\UbuntuPro` was still happening in the application's private hive instead of the global registry, thus not accessible to the users.

Instead we need to quote the direct key path (parents are not enough) in the `Package.appxmanifest` `<virtualization:ExcludeKey>` field.

That's hard to test, though, as it requires the deployment of the MSIX package and something outside of the package checking whether the registry keys are visible or not. Thankfully that's the exact scenario where our end-to-end tests run, so I'm adding a small step to verify that the registry key values are found as expected.

### DIFF
--- a/end-to-end/manual_token_input_test.go
+++ b/end-to-end/manual_token_input_test.go
@@ -85,6 +85,8 @@ func TestManualTokenInput(t *testing.T) {
 				require.NoErrorf(t, err, "Setup: could not wake distro up: %v. %s", err, out)
 			}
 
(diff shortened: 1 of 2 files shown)


hk21702 (approved):
LGTM

PR #1499, opened 2025-12-15
docs: improve consistency of content blocks

edibotopic:
This PR makes some changes to improve specific content blocks, namely dropdowns and codeblocks, primarily for consistency within our docs and with other Canonical docs.

## Styling of dropdowns

The first change relates to a small bit of custom CSS for our dropdowns.

The default styling for dropdowns in darkmode looks like this:

image

It's hard to read, and is --- in fact --- an accessibility issue due to poor contrast:

image

The Enterprise Store docs removes the grey tint, which leads to much better contrast:

image

I will discuss this issue with the Starter Pack maintainers, but, for now, let's fix it by applying the CSS rules to achieve the above effect.

Previously we had used the Primary colour of the dropdown block, which has better contrast but makes the dropdown elements a little too visually striking compared to the surround theme:

image

It also required maintainers to specify `:color: primary`, which is one more thing to remember.

## Consistency in usage of code blocks

Elsewhere, I have tried to make some small changes to improve the consistency of our usage of codeblocks.

In general, we should use prompt symbols when writing commands in the WSL docs unless there is a good reason not to.

In the absence of a better way, it remains a good reminder to the user/reader that we are referring to a specific context (PS versus Ubuntu).

When using prompts, we should always append `{code-block}` to the open backticks because this stops the prompts being copied (correction: it doesn't seem to be required for our prompt regex to apply, see https://github.com/canonical/ubuntu-pro-for-wsl/pull/1499#issuecomment-3661425867). This is not necessary, for example, when outlliing a full PowerShell or Bash script (edit: unless we need the additional functionality of that directive, see https://github.com/canonical/ubuntu-pro-for-wsl/pull/1499#issuecomment-3661425867). 

Currently, we don't use syntax highlighting for many code blocks. This is because it is inconsistent, especially for Bash, and may not be helpful to understand the commands. On the other hand, the PowerShell syntax highlight is relatively consistent with what most users will see in their terminal, especially when using PS commands like `Get-FileHash`, so it can be useful to use them occasionally.

Lastly, we should only use a prompt path, such as `PS C:\Users\me>` if it is useful in some way. For example, if the user must run the command in a specific directory. If it's not useful, it just adds visual noise, especially for long commands. In any case, using captions for code blocks is often a better approach for showing a path.

UDENG-8632

### DIFF
--- a/docs/howto/uninstalling.md
+++ b/docs/howto/uninstalling.md
@@ -72,7 +72,7 @@ recommending the modern tar-based format.
 
 If you don't know the format of an installed distro, run:
 
-```{codeblock} powershell
+```{code-block} powershell
 > Get-ChildItem "HKCU:\Software\Microsoft\Windows\CurrentVersion\Lxss"
 ```
 


CarlosNihelton:
> When using prompts, we should always append {code-block} to the open backticks because this stops the prompts being copied. This is not necessary, for example, when outlliing a full PowerShell or Bash script.

That reasoning also explains why we have some `ini` and `yaml` blocks not changed into `{code-blocks}`, right?

PR #1501, opened 2025-12-16
docs: update deployment tutorial

edibotopic:
The current deployment tutorial doesn't clearly explain how a user should set up a Landscape server,
even though it is a prerequisite for following the tutorial.

While it's possible to set up a self-hosted Landscape instance on the same machine as the Pro for WSL app for testing, all methods we have tried so far have been quite involved, and we're not ready to provide a quickstart approach.

We could rely on the Landscape docs, but setting up a test environment for Pro for WSL on Windows using VMs is a niche case that we can't expect the doc maintainers to accommodate.

The users who are following the Pro for WSL tutorial will have a Pro token, and that Pro token entitles them to also use Landscape SaaS, which is very convenient to set up.

I've updated the tutorial to more explicitly highlight this option, provide some basic steps to set it up, and also mention some configuration implications (e.g., FQDN and account name in the GUI).

There are also some small tweaks to the surrounding text.

UDENG-8674

### DIFF
--- a/docs/tutorials/deployment.md
+++ b/docs/tutorials/deployment.md
@@ -20,8 +20,8 @@ deploy Ubuntu to remote Windows machines using Landscape.
 ## What you will do
 
 - Register a Windows host instance with Landscape
-- Create a WSL profile on Landscape
-- Deploy Ubuntu to a remote Windows machine
+- Create a custom WSL profile on Landscape
+- Deploy Ubuntu with the custom profile to a remote Windows machine
 
 ## What you need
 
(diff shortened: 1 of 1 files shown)


docs/tutorials/deployment.md
```diff
@@ -94,27 +100,32 @@ Open the Pro for WSL app, enter your Pro token and continue to the Landscape con
...
 
 ## Register the Windows host instance with Landscape
 
 ```{admonition} Usage of the term "instance"
-:class: note
+:class: warning
 In the Landscape dashboard, an "instance" refers to the Windows host running WSL.
 
-On the Windows machine itself, an "instance" refers to an installed WSL distro.
+In this documentation, we often use "instance" to refer to instances of WSL running on the Windows host.
```
CarlosNihelton:
```suggestion
In this documentation, we often use "instance" to refer to instances of Ubuntu on WSL running on the Windows host.
```

or more generically

```suggestion
In this documentation, we often use "instance" to refer to instances of WSL distros running on the Windows host.
```

PR #1525, opened 2026-01-09
fix(msix): Better app icon with no text

cnihelton:
Built from just the official circle of friends (no texts to become unreadable when shown in the icon) exported to PNG (1600x1600, 300 dpi) and fed into Visual Studio Manifest Designer to regenerate the range of visual assets needed.

The screenshot below gives a hint on how the app icon is shown on Windows, but I advise reviewers to install the app built in CI and see the results closer if the screen has better resolution.

Notice that even the icon shown when the app is hovered in the Task Manager (`/msix/UbuntuProForWSL/Images/icon.ico`) had that hard-to-read text but now it also contains only the circle of friends.

image

---

UDENG-8668


CarlosNihelton:
About the increase in the white space area they are mostly non harmful because "Tiles" have been largely replaced on Windows 11 by simple icons in the Start menu. However, Windows still requires the tile assets for backward compatibility, and since we aim to keep supporting Windows 10 for as long as it makes sense, it's worth considering them.

The effect should be more pronounced in the Task manager, affected by the `Square*` files. In my system we see the Ubuntu icon nicely harmonized with other applications with and without the extra spaces, so I'm not really sure how high-DPI displays would be affected by that.

Screenshot From 2026-01-12 14-37-41


But considering that the extra spaces may affect the sharpness of some assets I'm testing removing them. I can't really notice the change, but you might be able to do so.

image


I also read [in this thread](https://learn.microsoft.com/en-us/answers/questions/5622957/appinstaller-wont-scale-my-store-installer-logo) that the App Installer specifically might have issues with those assets and the recommendation was to add a big (400x400) `StoreLogo.png` file to the assets. I'm trying both. Let me know whether it still shows up as blurry for you.

hk21702 (commented):
The desktop icon itself looks good on my relatively high DPI display, maybe a little off-center but I suspect that might just be because of how the logo is designed. 

The installer icon is still a bit blurry to me after using the new build; I think it is currently 200x200. To me, the web browser preview of that icon matches in the blurriness. https://raw.githubusercontent.com/canonical/ubuntu-pro-for-wsl/406e09c0b6909c4fb4dbffdd544d8bb76d4116d6/msix/UbuntuProForWSL/Images/StoreLogo.png

hk21702 (approved):
We discussed this offline. The linked thread is incorrect, and the installer does seem to use scaling, but it doesn't seem like it is worth the effort to try and resolve the issue with the blurry icon for now since it is largely acceptable already and users will rarely see it.

Thus, approving this PR.

PR #1527, opened 2026-01-13
feat(msix): Packaging for chocolatey

cnihelton:
Chocolatey has different semantics compared to winget. With chocolatey PowerShell and elevation are assumed.
As a result MSIX packages won't be simply installed for the current user, but rather to the base image, thus
the use of the `Add-AppxProvisionedPackage` cmdlet. Also, chocolatey provides some helper functions which
should be used when available, but there are none specifically for actually installing MSIX.

A Chocolatey package is a collection of PowerShell scripts to run at package installation, upgrade and removal, bundled together in the `.nupkg` format (a zip archive with some XML metadata). The upgrade scripts are optional and I'm completely skipping it in this PR, which implies `choco upgrade` does nothing for this package. The App Installer should care about the upgrades on its own, automatically, without the need of user intervention.

I'm also adding a CI workflow to publish the package to `chocolatey.org` when we publish a stable release. It can also be triggered by workflow_dispatch or via PR that touches relevant files, but in those cases we don't publish ofc. Since the PowerShell scripts provided here actually download the package (which is a common approach) instead of having it bundled inside the nupkg, we need to have the latest release already published on GitHub so the script can point to the correct URL, checksum and version. I originally wanted a more generic workflow name thinking that we might devise other actions to be taken upon release that would fit in the same workflow. But I decided to leave that for the future and name the workflow all about chocolatey since that's all we're doing now.

---

UDENG-8655


msix/chocolatey/ubuntuproforwsl/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,42 @@
...
+
+$packageName = $env:ChocolateyPackageName
+$packageVersion = $env:ChocolateyPackageVersion
+$toolsDir = "$(Split-Path -parent $MyInvocation.MyCommand.Definition)"
+$msixBundleFileName = "UbuntuProForWSL-$packageVersion.msixbundle"
+$downloadPath = Join-Path $toolsDir $msixBundleFileName
+
+# The asset.json file is updated by CI at release time.
+$asset = Get-Content -Raw -Path "./asset.json" | ConvertFrom-Json
+
```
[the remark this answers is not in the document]
CarlosNihelton: `$msixBundleFileName = $asset.name`  also does it. The asset being downloaded has unversioned file name, so it's questionable whether that's a good idea or not rename the downloaded file. I'm OK with it based on the fact that we clean it up later on, i.e. it's meant to be a short-lived file.

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,72 @@
...
+          body: "[Auto-generated pull request](https://github.com/canonical/ubuntu-pro-for-wsl/actions/workflows/on-release-published.yaml) by GitHub Action"
+          branch: auto-updates/chocolatey
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - name: Push branch
+        run: |
+          git push origin auto-updates/chocolatey:main
+
+
+
+
```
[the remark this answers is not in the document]
CarlosNihelton: Indeed. Copy-pasta :)

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Check out repository
+        uses: actions/checkout@v6
+      - name: Install dependencies
+        shell: bash
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt update
+          sudo DEBIAN_FRONTEND=noninteractive apt install -y git jq xmlstarlet
+      - shell: bash
+        env:
+          GH_TOKEN= ${{ secrets.github_token }}
```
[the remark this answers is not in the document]
CarlosNihelton: The syntax is incorrect either way, it should be `GH_TOKEN: ${{ secrets.github_token }}`.

msix/chocolatey/ubuntuproforwsl/tools/chocolateyinstall.ps1
```diff
@@ -0,0 +1,58 @@
...
+$msixBundleFileName = $asset.name
+$downloadPath = Join-Path $toolsDir $msixBundleFileName
+
+Write-Verbose "Downloading $msixBundleFileName from $url"
+# Use the Chocolatey helper function to download the file securely.
+Get-ChocolateyWebFile	        `
+    -PackageName $packageName   `
+    -FileFullPath $downloadPath `
+    -Url $url			`
+    -Checksum $checksum		`
```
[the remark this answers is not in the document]
CarlosNihelton: It was well formatted, though manually :/

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,71 @@
...
+          releaseJson=$( curl -L \
+                         -H "Accept: application/vnd.github+json" \
+                         -H "Authorization: Bearer $GH_TOKEN" \
+                         -H "X-GitHub-Api-Version: 2022-11-28" \
+                            "${{ github.api_url }}/releases/tags/$tag" \
+                        | tr -d '[:cntrl:]' ) 
+
+          cd ./msix/chocolatey/ubuntuproforwsl/
+          # Replaces the existing asset.json with information from the latest release.
+          jq '.name as $version |
```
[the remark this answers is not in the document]
CarlosNihelton: Nope, there is a convention to assign the MSIX version to the release name.

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,71 @@
...
+        run: |
+          # Documentation says GITHUB_REF is the Tag ref of release refs/tags/<tag_name>
+          tag=$(echo $GITHUB_REF | cut -f2 -d'/')
+          releaseJson=$( curl -L \
+                         -H "Accept: application/vnd.github+json" \
+                         -H "Authorization: Bearer $GH_TOKEN" \
+                         -H "X-GitHub-Api-Version: 2022-11-28" \
+                            "${{ github.api_url }}/releases/tags/$tag" \
+                        | tr -d '[:cntrl:]' ) 
+
```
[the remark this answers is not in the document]
CarlosNihelton: That's mostly to prevent the body of the release to break jq because it may contain new lines that break JSON parsing. The fields we care about are not affected by the removal of control chars.

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,65 @@
...
+    name: Update chocolatey assets
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+      - name: Install dependencies
+        shell: bash
+        run: |
+          sudo DEBIAN_FRONTEND=noninteractive apt update
+          sudo DEBIAN_FRONTEND=noninteractive apt install -y git jq xmlstarlet
```
hk21702: Consider using apt-get.

I think you can also set DEBIAN_FRONTEND to be noninteractive on the job or workflow level to make things cleaner and easier.

Also, `git` and `jq` should already be included in the runner images. 
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md

.github/workflows/on-release-published.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v8
+        with:
+          commit-message: Auto update chocolatey files
+          title: 'maint(msix): Auto update chocolatey files'
+          labels: |
+            packaging
+            automated pr
+          body: "[Auto-generated pull request](https://github.com/canonical/ubuntu-pro-for-wsl/actions/workflows/on-release-published.yaml) by GitHub Action"
```
hk21702: Nitpick: It might be more helpful to link the action run that created the PR.

msix/chocolatey/ubuntuproforwsl/ubuntuproforwsl.nuspec
```diff
@@ -0,0 +1,101 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Do not remove this test for UTF-8: if “Ω” doesn’t appear as greek uppercase omega letter enclosed in quotation marks, you should use an editor that supports UTF-8, not this one. -->
+<package xmlns="http://schemas.microsoft.com/packaging/2015/06/nuspec.xsd">
+  <metadata>
+    <!-- == PACKAGE SPECIFIC SECTION == -->
+    <!-- This section is about this package, although id and version have ties back to the software -->
+    <!-- id is lowercase and if you want a good separator for words, use '-', not '.'. Dots are only acceptable as suffixes for certain types of packages, e.g. .install, .portable, .extension, .template -->
+    <!-- If the software is cross-platform, attempt to use the same id as the debian/rpm package(s) if possible. -->
+    <id>ubuntuproforwsl</id>
```
hk21702: Based on the naming guidelines, we should use hyphens here since squishing it all together is making it hard to read. `ubuntu-pro-for-wsl` would be appropriate.


https://docs.chocolatey.org/en-us/create/create-packages/#naming-your-package

hk21702 (changes requested):
My personal preference would be to avoid creating pull requests entirely, and treat `assets.json` and `*.nuspec` as files we generate with CI/CD when a pack and publish is requested. It would reduce maintenance requirements, and those files already are files that need to be updated via automation every time we publish. Beyond reducing a manual PR step, keeping it in the repository and updating it with PRs after publication on GitHub means that there will always be a slight mismatch at the tag. Keeping it ephemeral also ensures that there's no implication that these files should be modified manually ever.

Another thing is that we should consider including a test step in a workflow which just ensures that we can install from the package and also uninstall correctly, preferably before any publish step. https://docs.chocolatey.org/en-us/create/create-packages/#testing-your-package

Also, presently, what happens in the case where one user installs UP4W for themselves, for example manually or with Winget. Then, an admin installs, then removes UP4W via choco? Since the msix package name is the same and the uninstall script doesn't appear to differentiate between initial install sources, would the initial user's installation also be removed? Would this be considered problematic?

CarlosNihelton:
> Also, presently, what happens in the case where one user installs UP4W for themselves, for example manually or with Winget. Then, an admin installs, then removes UP4W via choco? Since the msix package name is the same and the uninstall script doesn't appear to differentiate between initial install sources, would the initial user's installation also be removed? Would this be considered problematic?

My reasoning for this is: choco runs with admin privileges, so its OK for its actions to override user's.

But there's a gotcha: choco is a package manager, so it only knows about the packages it installs. For the uninstall command to become a problem the actions must happen exactly like you described: the user installs the app, then the admin also installs the app with chocolatey and then removes it. `choco uninstall` without a previous `choco install` won't do anything.

CarlosNihelton:
@hk21702, since you convinced me to go straight into full automation, the workflow is changing considerably, so much that I must reword the PR description.

msix/chocolatey/ubuntu-pro-for-wsl.nuspec.template
```diff
@@ -0,0 +1,46 @@
...
+    <!-- Here is an example using Githack -->
+    <iconUrl>https://rawcdn.githack.com/canonical/ubuntu-pro-for-wsl/main/msix/UbuntuProForWSL/Images/StoreLogo.png</iconUrl>
+    <copyright>2025 Canonical Ltd.</copyright>
+    <!-- If there is a license Url available, it is required for the community feed -->
+    <licenseUrl>https://rawcdn.githack.com/canonical/ubuntu-pro-for-wsl/main/LICENSE</licenseUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <projectSourceUrl>https://github.com/canonical/ubuntu-pro-for-wsl</projectSourceUrl>
+    <docsUrl>https://documentation.ubuntu.com/wsl</docsUrl>
+    <bugTrackerUrl>https://github.com/canonical/ubuntu-pro-for-wsl/issues</bugTrackerUrl>
+    <tags>ubuntu-pro wsl automation landscape</tags>
```
hk21702: Maybe add ubuntu as a tag by itself.

.github/workflows/chocolatey.yaml
```diff
@@ -0,0 +1,98 @@
...
+    name: Update package
+    runs-on: windows-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+      - name: Generate the chocolatey files
+        env:
+          RELEASE_JSON_RAW: ${{ github.event.name == 'release' && toJson(github.event.release) || '' }}
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        working-directory: ./msix/chocolatey/
```
hk21702: Since this is the working directory for all the steps, consider setting a job-level default instead of setting it for every step.

https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/set-default-values-for-jobs

.github/workflows/chocolatey.yaml
```diff
@@ -0,0 +1,98 @@
...
+          $descNode = $metadataNode.Element($ns + "description") 
+          if ($descNode) {
+              # Replace content with a new XCData object
+              $descNode.ReplaceNodes((New-Object System.Xml.Linq.XCData($readmeContent)))
+          }
+
+          $doc.Save($nuspecPath.BaseName) # Removes the .template extension, leaving the .nuspec
+          choco pack
+
+      - name: Test the package
```
hk21702: Could we also smoke test uninstalling the package?

hk21702 (commented):
Looks good overall, just a few nitpicks.

hk21702 (commented):
Besides the nitpicks Copilot pointed out, it may be nice for the installation smoke tests to be able to validate whenever we installed or uninstalled UP4W. It's fine do have this be a seperate task/different PR though.

Another thought is to maybe make the copyright year be defined programmatically so we never have to think about it.

msix/chocolatey/README.md
```diff
@@ -0,0 +1,50 @@
+﻿# Ubuntu Pro for WSL
+
+**Ubuntu Pro for WSL** is a Windows application that automates the
```
hk21702: We should make paragraphs all in one line, unless we want to force line breaks.

https://docs.chocolatey.org/en-us/create/create-packages/#package-description-and-release-notes

Based on other packages, it appears that a line break without a newline is enough to render as a new paragraph.

See the "These parameters can be passed to the installer with the use of..." line from the Adobe Reader package as an example:
https://community.chocolatey.org/packages/adobereader
https://github.com/pauby/ChocoPackages/blob/master/automatic/adobereader/adobereader.nuspec

hk21702 (changes requested):
Almost there! I have some concerns about the formatting of the README since how they render markdown seems to differ from GitHub. Apologies for not catching that earlier.

As another note, the docs mention adding Chocolatey as a dependency, but we don't use any fancy features, so it is probably fine to omit it like we are now.

msix/chocolatey/README.md
```diff
@@ -0,0 +1,42 @@
+﻿# Ubuntu Pro for WSL
+
+**Ubuntu Pro for WSL** is a Windows application that automates the attachment of your Ubuntu Pro subscription. It solves the problem of needing to manually Pro-attach each new Ubuntu instance created on a Windows machine when you want the security benefits of Ubuntu Pro. For organisations, it enables automated Pro-attachment at scale for fleets of devices.
+
+## Basic usage of Ubuntu Pro for WSL
+
+Find and install the Ubuntu Pro for WSL application in the Microsoft Store.
```
hk21702: Isn't relevant for chocolatey.

hk21702 (changes requested):

hk21702 (approved):
Looks good! Once we publish, let's double-check how Chocolatey renders the description in case we need to adjust.

PR #1536, opened 2026-01-16
feat(windows-agent): Ensure telemetry consent registry is set and valid

hk21702:
The telemetry consent registry key found at Software\Canonical\UbuntuPro is used by wsl-setup to determine what it should use for Ubuntu Insights consent. If it is not set, then it will prompt the user.

Here, we default the registry value to zero whenever the Windows Agent watcher starts if it is not already set to ensure that this new behavior does not break the workflows of those utilizing Ubuntu Pro for WSL. This is only initial behavior, and we should eventually expand upon this to provide the user with greater control of consent through all of their WSL instances via UP4W in the future.

Though we now read and modify the registry path `Software/Canonical/Ubuntu`, Ubuntu Pro for WSL should not be considered the owner of this path, and it should *not* clean it up on uninstall as Ubuntu for WSL uses it independently.

### DIFF
--- a/msix/UbuntuProForWSL/Package.appxmanifest
+++ b/msix/UbuntuProForWSL/Package.appxmanifest
@@ -12,6 +12,7 @@
     <virtualization:RegistryWriteVirtualization>
       <virtualization:ExcludedKeys>
         <virtualization:ExcludedKey>HKEY_CURRENT_USER\Software\Canonical\UbuntuPro</virtualization:ExcludedKey>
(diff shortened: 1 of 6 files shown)


cnihelton (changes requested):
Great job as usual! I have some minor things to flag.

cnihelton (approved):
Awesome work! The lower coverage in this particular PR is quite expected due the amount of Windows specific code being touched.

PR #1558, opened 2026-02-10
feat(CI): Skip publishing the MSIX sometimes

cnihelton:
Times when we want to tag the repo for a documentation refresh or to release just wsl-pro-service and the recent git commits don't justify also publishing the Windows app.

---
UDENG-8603


hk21702 (approved):
While we have eyes on the changelog, are you moving to the new `*.*.*ubuntu` versioning syntax for native packages?

PR #1566, opened 2026-02-18
fix(ci): Fix the update-po CI job

adombeck:
The job didn't update the po files because it used the wrong step output since 1f3a026aeaac24952d85864e330386fd6b3dcc82.


cnihelton (approved):
LGTM! Thanks for noticing!

PR #1580, opened 2026-02-25
fix(windows-agent): Handling DNS errors more gracefully

cnihelton:
We were very optimistc about the DNS resolution process. A lot could go wrong in the wild outside of our (and our users) control, causing DNS errors to be more common **and more transient** than we expected.

I looked into gRPC internal retrials, hedging and backoff mechanisms in hopes to find a way to differentiate wrong server addresses from transient or fatal name resolution errors, but there is no one-size-fits-all solution for that. Thus, I'm relying on our existing backoff mechanism, with a different duration than the broader backoff mechanism already implemented. After all, DNS errors still could be a mistaken input, for which is not worth taking forever to let the user know.

I'm also slightly adjusting the gRPC internal backoff mechanism mainly to add some jitter to the retry duration, which is a common practice to avoid thundering herd problems.

---

UDENG-9295

### DIFF
--- a/windows-agent/internal/proservices/landscape/connection.go
+++ b/windows-agent/internal/proservices/landscape/connection.go
@@ -13,7 +13,6 @@ import (
 	log "github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logstreamer"
 	"github.com/ubuntu/decorate"
 	"google.golang.org/grpc"
-	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/connectivity"
 )
 
@@ -58,18 +57,7 @@ func newConnection(ctx context.Context, d serviceData) (conn *connection, err er
 		return nil, err
 	}
 
-	grpcConn, err := grpc.NewClient(conn.hostConf.hostagentURL,
-		grpc.WithTransportCredentials(creds),
-		grpc.WithConnectParams(grpc.ConnectParams{
-			Backoff: backoff.Config{
-				BaseDelay:  1.0 * time.Second,
-				Multiplier: 1.6,
-				Jitter:     0.2,
-				MaxDelay:   120 * time.Second,
-			},
-			MinConnectTimeout: 60 * time.Second,
-		}),
-	)
+	grpcConn, err := grpc.NewClient(conn.hostConf.hostagentURL, grpc.WithTransportCredentials(creds))
 	if err != nil {
 		return nil, err
 	}
--- a/windows-agent/internal/proservices/landscape/connection.go
+++ b/windows-agent/internal/proservices/landscape/connection.go
@@ -13,6 +13,7 @@ import (
 	log "github.com/canonical/ubuntu-pro-for-wsl/common/grpc/logstreamer"
 	"github.com/ubuntu/decorate"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/backoff"
 	"google.golang.org/grpc/connectivity"
 )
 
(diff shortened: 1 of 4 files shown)


windows-agent/internal/proservices/landscape/service.go
```diff
@@ -160,6 +161,7 @@ func (s *Service) keepConnected() error {
 	const minWait = time.Second
 	const maxWait = 10 * time.Minute
 	wait := 0 * time.Second // No wait in the first iteration
+	errc := errorCount{}
```
hk21702: This doesn't seem to be used outside the goroutine, so why not declare it there, especially since the goroutine is modifying it? Currently, it should be safe, but it's probably safer for the future should this be expanded.

hk21702 (changes requested):
Seems good overall, I just have a few nitpicks here and there!

windows-agent/internal/proservices/landscape/connection.go
```diff
@@ -57,15 +58,26 @@ func newConnection(ctx context.Context, d serviceData) (conn *connection, err er
...
+	grpcConn, err := grpc.NewClient(conn.hostConf.hostagentURL,
+		grpc.WithTransportCredentials(creds),
+		grpc.WithConnectParams(grpc.ConnectParams{
+			Backoff: backoff.Config{
+				BaseDelay:  1.0 * time.Second,
+				Multiplier: 1.6,
+				Jitter:     0.2,
+				MaxDelay:   120 * time.Second,
+			},
+			MinConnectTimeout: 60 * time.Second,
```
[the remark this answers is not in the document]
CarlosNihelton: Good catch. That was a leftover from previous attempts at configuring gRPC to handle this by itself. It didn't go well and that line remained unnoticed while I was trying to preserve the backoff config for the jitter.

I couldn't find in documentation nor source code the default value for `MinConnectTimeout`, but your question prompted me to look if there was any default backoff config and guess what?! The values I supplied were the defaults 🤡 . So I'm removing this configuration entirely.

windows-agent/internal/proservices/landscape/service.go
```diff
@@ -273,24 +277,68 @@ func (s *Service) keepConnected() error {
...
+	noConfig        int
+	serverRejection int
+	nameResolution  int
+	other           int
+}
+
+// isSaturated returns true if the  error counts reached their maximum allowance, false otherwise.
+func (errc errorCount) isSaturated() bool {
+	// Just hardcoding the maximum counts for now. We might want to adjust this to become a constructor parameter.
+	return errc.noConfig > 0 || errc.serverRejection > 0 || errc.nameResolution > 6
```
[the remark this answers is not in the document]
CarlosNihelton: That should not be the case now that I removed the `MinConnectTimeout` of 60 seconds. While the default value is apparently not documented, in my observation it takes about 15s for a connection attempt to fail due lack of network or invalid addresses. So the total wait shouldn't be much more than 2 minutes.

windows-agent/internal/proservices/landscape/landscape_test.go
```diff
@@ -205,8 +207,12 @@ func TestConnect(t *testing.T) {
 				// Fixture exists: override the Landscape config
 				lconf = string(fixture)
 			}
+			address := lis.Addr().String()
+			if tc.serverNotFound {
+				address = "server.notfound.local:1234"
```
CarlosNihelton: Even this doesn't result in a DNSError :/

hk21702: Yeah, if that doesn't work, then I'm not sure how to really trigger that error then. It might be a case where if we *really* want to just test this, then we'd have to mock things.

hk21702 (approved):
Looks good!

Shame that there doesn't seem to be a good way to trigger a DNSError, but I don't think we can really get around that without dependency injecting/mocks, which would overcomplicate such a small and relatively simple section.

PR #1614, opened 2026-03-27
maint(CI): end-to-end tests with tar images and running on GH hosted runners

cnihelton:
This PR aims to revamp the end-to-end tests to:
- run in GH hosted runners
- use tar-based distro images instead of appx based.

By doing so we move from CI runs that can take between 30-45 minutes (excluding outliers) doing basically installation checks into a scenario where we do install, run the system end to end and assert on expected effects in less than 20 minutes.

The work is not complete yet, there are more improvements to come, but the current state is digestible enough.
Also, the current discoveries and fixes are making this PR way to big, reason why some fixes were moved into PRs #1620 and #1621 . Yet, there are a few blockers preventing enablement of more tests cases. Either I find a smart way to make GoWSL prevent contention against WSL or WSL itself becomes more resilient to concurrency. When working on this PR I saw it broken in multiple ways, from commands never completing to WSL instances failing to discover their own name... Because of that I'm limiting the enablement of test cases to just the integration of Landscape and cloud-init (after trying a lot to re-enable the manual token input test case :/)

A number of small peasant changes were done to either leverage caching features of certain GH actions (flutter-action, setup-go) or to upgrade them to the latest stable versions (upload-artifact, download-artifact, setup-msbuild).

A number of peasant fixes were made along the way to prevent timeouts on failure cases. Using WSL in go tests is quite tricky, we can easily face deadlocks because of calls jumping threads (if done inside goroutines) or just because of seemingly excessive I/O. Under the hood WSL and WSLAPI relies on COM, so I'm not exactly sure on the limitations at the moment, reason why I preventively included a COM setup call in main_test when facing some strange deadlocks. The calls going via GoWSL should be more predictable now. Yet, you'll see some being replaced by subprocessing wsl.exe instead, in hopes that makes the test setup and execution more reliable. I'd like to revisit those eventually, to learn more about the true limitations of that API.

The main_test setup stage was cleared from assumptions of appx distros as well as of building the MSIX for testing. CI always did that, so there is no need to preserve such complex function. We can replicate the same steps CI does locally when running those tests locally anyway. Similarly, I moved the preparation of the golden image (with a build of wsl-pro-service from CI installed and configured for verbose logging) out of the main_test.go into CI, where it's more visible and hopefully that also helps with WSL contention, because between those commands preparing the image and actually running the tests we have a buffer of time of `go test` compilation.

When reworking the GH workflow, I replicated the get-wsl-image from wsl-setup, then I setup a manifest override and register it so `wsl --install -d Ubuntu-Preview` will install the golden image and  `wsl --install Ubuntu` will install the cached image.

This PR depends on #1620 and #1621 which contains bug fixes or reliability improvements that allow the end to end tests to run in GH runners as proposed here. 

Closes #1551 as well.

---
UDENG-7028

### DIFF
--- a/.github/actions/get-wsl-image/action.yaml
+++ b/.github/actions/get-wsl-image/action.yaml
@@ -1,0 +1,141 @@
+# This action gets the latest WSL image, leveraging both the persistent behavior of our runners, and the GitHub cache
+name: Get WSL Image
+description: Downloads and caches an Ubuntu image for WSL assuming the URL provided follows the format of the official repositories.
+
+inputs:
+  base_url:
+    description: "The base URL to download the WSL image from"
+    required: false
+    default: "https://cdimages.ubuntu.com/ubuntu-wsl/daily-live/current/"
+  arch:
+    description: "The architecture of the WSL image to download (amd64 or arm64)"
+    required: false
+    default: "amd64"
+  output_path:
+    description: "The path to save the downloaded WSL image to"
+    required: false
+    default: "ubuntu.wsl"
+outputs:
+  cache-key:
+    description: "The cache key for the WSL image, which is its SHA256 hash"
+    value: ${{ steps.compute-cache-key.outputs.cache-key }}
+  image-name:
+    description: "The name of the image we found"
+    value: ${{ steps.compute-cache-key.outputs.image-name }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Compute image cache key
+      id: compute-cache-key
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        shafile="${{ runner.temp }}/SHA256SUMS"
+        uri="${{ inputs.base_url }}/SHA256SUMS"
+        if ! curl --fail -o "$shafile" "${uri}"; then  
+            echo "::error:: Failed to download the checksum file from $uri."  
+            exit 1  
+        fi  
+
+        # Look for a line matching the pattern: <sha> *<name>-wsl-<arch>.wsl
+        arch="${{ inputs.arch }}"
+        line=$(grep -E "^[a-f0-9]+[[:space:]]+\*.*-wsl-${arch}\.wsl$" "$shafile")
+        if [ -z "$line" ]; then
+          echo "::error:: No WSL image found for architecture '$arch' in the $shafile file."
+          exit 2
+        fi
+
+        # Extract the SHA (first field), image name (second field without the *) and manifest.
+        sha=$(echo "$line" | awk '{print $1}')
+        image_name=$(echo "$line" | awk '{print $2}' | sed 's/^\*//')
+        image_manifest=$(echo "$image_name" | sed -E 's/\.wsl$/.manifest/')
+
+        echo "cache-key=$sha" >> "$GITHUB_OUTPUT"
+        echo "image-name=$image_name" >> "$GITHUB_OUTPUT"
+        echo "image-manifest=$image_manifest" >> "$GITHUB_OUTPUT"
+
+    - name: Validate image
+      id: check-locally
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        expected_sha="${{ steps.compute-cache-key.outputs.cache-key }}"
+        output_path="${{ inputs.output_path }}"
+
+        if [ -r "$output_path" ]; then
+          echo "Found local image at $output_path"
+          
+          # Compare the SHAs
+          if (echo "$expected_sha  $output_path" | sha256sum -c); then
+            echo "✅ Local image matches expected SHA. Using local file."
+            echo "is-valid=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "❌ Local image SHA does not match expected SHA. Will need to download."
+            rm "$output_path"
+            echo "is-valid=false" >> "$GITHUB_OUTPUT"
+          fi
+        else
+          echo "No local image found at $output_path"
+          echo "is-valid=false" >> "$GITHUB_OUTPUT"
+        fi
+
+    - name: Restore image cache
+      id: restore-cache
+      if: ${{ steps.check-locally.outputs.is-valid != 'true' }}
+      uses: actions/cache/restore@v4
+      with:
+        path: ${{ inputs.output_path }}
+        key: ${{ steps.compute-cache-key.outputs.cache-key }}
+        enableCrossOsArchive: true
+
+    - name: Download image
+      if: ${{ steps.check-locally.outputs.is-valid != 'true' && steps.restore-cache.outputs.cache-hit != 'true' }}
+      shell: bash
+      run: |
+        set -euo pipefail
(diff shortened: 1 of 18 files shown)


cnihelton:
We finally have enough success runs in a row to consider this ready.
If you look at https://github.com/canonical/ubuntu-pro-for-wsl/actions/workflows/end-to-end.yaml skipping the second to last (which I cancelled because I broke the test code and immediately push a fix), then we have three good and fast runs in a row.

end-to-end/utils_test.go
```diff
@@ -200,6 +221,21 @@ func logWslProServiceOnError(t *testing.T, ctx context.Context, d gowsl.Distro)
...
+func logProClientOnError(t *testing.T, instanceName string) {
+	t.Helper()
+
+	if !t.Failed() {
+		return
+	}
+
+	out, err := os.ReadFile(fmt.Sprintf(`\\wsl.localhost\%s\var\log\ubuntu-advantage.log`, instanceName))
+	if err != nil {
+		t.Logf("could not access Pro Client logs: %v\n%s\n", err, out)
```
hk21702:
```suggestion
		t.Logf("Could not access Pro Client logs: %v\n%s\n", err, out)
```

end-to-end/cloud_init_test.go
```diff
@@ -77,50 +85,63 @@ func TestCloudInitIntegration(t *testing.T) {
...
+	defer logProClientOnError(t, distro.Name())
 
 	require.Eventually(t, func() bool {
-		attached, err := distroIsProAttached(t, ctx, distro)
-		if err != nil {
-			t.Logf("could not determine if distro is attached: %v", err)
+		attached, err1 := distroIsProAttached(t, ctx, distro)
+		tried, err2 := triedProAttach(t, distro.Name())
+		if err1 != nil && err2 != nil {
+			t.Logf("could not determine if distro tried to attach to Pro: %v", errors.Join(err1, err2))
```
hk21702:
```suggestion
			t.Logf("Could not determine if distro tried to attach to Pro: %v", errors.Join(err1, err2))
```

end-to-end/cloud_init_test.go
```diff
@@ -77,50 +85,63 @@ func TestCloudInitIntegration(t *testing.T) {
...
 			return false
 		}
-		return ok
-	}, time.Minute, time.Second, "Distro should have been registered")
+		if !ok {
+			return false
+		}
+		state, err := distro.State()
+		if err != nil {
+			t.Logf("could not determine if distro is registered: %v", err)
```
hk21702:
```suggestion
			t.Logf("Could not determine if distro is registered: %v", err)
```

end-to-end/main_test.go
```diff
@@ -79,33 +71,35 @@ func TestMain(m *testing.M) {
...
-	testImagePath = path
+	testImagePath = filepath.Join(buildPath, "images", referenceImage)
 
 	m.Run()
 
 	if err := cleanupRegistry(); err != nil {
 		log.Printf("Cleanup: registry: %v\n", err)
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
hk21702: You may want to increase this to 10 seconds; 5 seems pretty tight for removing an appx package.

end-to-end/cloud_init_test.go
```diff
@@ -77,50 +85,63 @@ func TestCloudInitIntegration(t *testing.T) {
...
-	require.Equal(t, "testuser", userName, "cloud-init should have configured the default user")
+		return attached || tried
+	}, 10*time.Second, time.Second, "distro should have tried to attach to Pro")
+
+	uid, err := distro.Command(ctx, "id -u testuser").CombinedOutput()
+	require.NoError(t, err, "cloud-init should have configured the default user, uid is %s", uid)
+	// Finally, give extra room for wsl-pro-service to talk to the agent.
+	cmd := exec.CommandContext(ctx, "wsl.exe", "-d", name)
+	require.NoError(t, cmd.Start(), "Could not launch the distro for final assertions")
+	time.Sleep(1 * time.Second)
```
hk21702: I'm not super confident with relying on this 1-second sleep here. Is there a better way of doing this that would be more deterministic? Though, maybe it's fine since the helpers we use after both use `require.Eventually`, though that is hidden in the helper.

.github/actions/get-wsl-image/action.yaml
```diff
@@ -0,0 +1,141 @@
...
+      id: compute-cache-key
+      shell: bash
+      run: |
+        set -euo pipefail
+
+        shafile="${{ runner.temp }}/SHA256SUMS"
+        uri="${{ inputs.base_url }}/SHA256SUMS"
+        if ! curl --fail -o "$shafile" "${uri}"; then  
+            echo "::error:: Failed to download the checksum file from $uri."  
+            exit 1
```
hk21702:
```suggestion
            exit 1
```

end-to-end/landscape_utils_test.go
```diff
@@ -105,6 +105,7 @@ func (l landscape) LogOnError(t *testing.T) {
 func (l landscape) Stop() {
 	if l.stop != nil {
 		l.stop()
+		l.server.Stop()
```
hk21702, line 110: Could `l.server` ever be nil here as well?

.github/workflows/end-to-end.yaml
```diff
@@ -57,64 +55,45 @@ jobs:
...
           path: ci-artifacts
-      - name: Set up artifacts
+      - name: Download WSL image
+        uses: ./.github/actions/get-wsl-image
+        id: get-wsl-image
+        with:
+          base_url: "https://cdimages.ubuntu.com/ubuntu-wsl/daily-live/current/"
+          arch: "amd64"
+          output_path: "ci-artifacts/images/ubuntu.wsl"
+      - name: Layout the downloaded artifacts
```
hk21702:
```suggestion
      - name: Lay out the downloaded artifacts
```

hk21702 (changes requested):
Few nitpicks here and there. There are also some considerations to be made with the timing of some async things.

Also, with the testify package, the messages should be error-like and thus not have capitalized first letters, while I think `t.Logf` is a log message, so they should have capitalization.

end-to-end/landscape_utils_test.go
cnihelton: The setup of this mock is quite dramatic, it panics for any failure along the way (via `require.NoError()`) so if we get to this point we have a non-nil server indeed. But that setup may change soon, so to prevent surprises I'm adding a nil check anyways.

end-to-end/cloud_init_test.go
cnihelton: Indeed! A better approach is to monitor the messages sent by the agent in the Landscape mock.

hk21702 (approved):
LGTM

PR #1620, opened 2026-04-07
fix(windows-agent): Delivers tasks to newly connected WSL instances

cnihelton:
There is a possibility of Pro being configured after a WSL instance being registered but before it connects to the Windows agent's wslinstance gRPC service. When the connection happens the current implementation just assumes the instance was provisioned after the configuration, thus it does nothing about the instance.

I'm changing it to conditionally send tasks to that instance instead. Also, implementing a test to ensure we deliver tasks to new instances connecting to the agent if they are not already pro attached.

A nice side effect is that if for some reason (a naughty user?) an existing instance assumed to be pro-attached loose that state, it should be recovered on the next reconnection to the Windows agent.

This is a spin off of the PR #1614 - I found this bug while working on that task.

### DIFF
--- a/windows-agent/internal/proservices/proservices.go
+++ b/windows-agent/internal/proservices/proservices.go
@@ -3,6 +3,7 @@ package proservices
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -15,12 +16,16 @@ import (
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/cloudinit"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/config"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/distros/database"
+	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/distros/distro"
+	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/distros/task"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/proservices/landscape"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/proservices/registrywatcher"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/proservices/ui"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/proservices/wslinstance"
+	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/tasks"
 	"github.com/canonical/ubuntu-pro-for-wsl/windows-agent/internal/ubuntupro"
 	"github.com/sirupsen/logrus"
+	"github.com/ubuntu/decorate"
 	wsl "github.com/ubuntu/gowsl"
 	"google.golang.org/grpc"
 	"google.golang.org/grpc/credentials"
@@ -110,8 +115,20 @@ func New(ctx context.Context, publicDir, privateDir string, args ...Option) (s M
 	}
 	s.landscapeService = landscape
 
-	s.wslInstanceService = wslinstance.New(ctx, s.db, s.landscapeService.Controller())
+	// When a new instance connects to the wslinstance service we'll greet it with some tasks.
+	onNewInstance := func(d *distro.Distro) {
+		// When a new instance connects to the wslinstance service we'll greet it with some tasks.
+		dtasks, err := newInstanceTasks(conf, d.Properties())
+		if len(dtasks) > 0 {
+			err = errors.Join(err, d.SubmitDeferredTasks(dtasks...))
+		}
+		if err != nil {
+			log.Warningf(ctx, "%v", err)
+			return
+		}
+	}
 
+	s.wslInstanceService = wslinstance.New(ctx, s.db, onNewInstance, s.landscapeService.Controller())
 	conf.SetUbuntuProNotifier(func(ctx context.Context, token string) {
 		ubuntupro.Distribute(ctx, s.db, token)
 		landscape.NotifyUbuntuProUpdate(ctx, token)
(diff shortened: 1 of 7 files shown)


windows-agent/internal/proservices/proservices.go
```diff
@@ -147,6 +164,25 @@ func New(ctx context.Context, publicDir, privateDir string, args ...Option) (s M
...
 
+// newInstanceTasks returns the initial tasks to be executed when a new instance connects to the WSLInstance service.
+func newInstanceTasks(conf *config.Config, p distro.Properties) (t []task.Task, err error) {
+	defer decorate.OnError(&err, "when new instance %q connected to WSLInstance service", p.DistroID)
+
+	pro, source, err := conf.Subscription()
+	if err != nil {
+		return nil, err
+	}
+	// There is a Pro subscription bu tthe instance is not attached.
```
hk21702:
```suggestion
	// There is a Pro subscription but the instance is not attached.
```

windows-agent/internal/proservices/proservices_test.go
```diff
@@ -202,6 +206,141 @@ func TestRegisterGRPCServices(t *testing.T) {
...
+					require.NoError(t, result.err, "ProAttachmentCommands stream should have received a command")
+					require.Equal(t, wantToken, result.cmd.GetToken(), "ProAttachCmd token should match the registry-provided token")
+				} else {
+					require.Error(t, result.err, "ProAttachmentCommands stream should not have received any commands")
+				}
+			case <-time.After(timeout):
+				if tc.wantCommands {
+					t.Fatal("timed out waiting for ProAttachCmd from the server")
+				}
+			}
```
hk21702, line 341: Just to make sure, we don't care about the exact number of commands sent right?

windows-agent/internal/proservices/proservices_test.go
```diff
@@ -202,6 +206,141 @@ func TestRegisterGRPCServices(t *testing.T) {
...
+	if wsl.MockAvailable() {
+		t.Parallel()
+	}
+
+	testcases := map[string]struct {
+		alreadyAttached bool
+		wantCommands    bool
+	}{
+		"When the instance is not already attached, onNewInstance should submit a ProAttachment task": {wantCommands: true},
+		"When the instance is already attached, onNewInstance should not submit a ProAttachment task": {alreadyAttached: true},
```
hk21702, line 223: Should we also be testing to ensure that we receive LandscapeConfigure tasks when we expect to?

windows-agent/internal/proservices/proservices.go
```diff
@@ -110,8 +115,20 @@ func New(ctx context.Context, publicDir, privateDir string, args ...Option) (s M
...
+	onNewInstance := func(d *distro.Distro) {
+		// When a new instance connects to the wslinstance service we'll greet it with some tasks.
+		dtasks, err := newInstanceTasks(conf, d.Properties())
+		if len(dtasks) > 0 {
+			err = errors.Join(err, d.SubmitDeferredTasks(dtasks...))
+		}
+		if err != nil {
+			log.Warningf(ctx, "%v", err)
+			return
+		}
```
hk21702: It feels a bit strange to still be trusting the `dtasks` output from `newInstanceTasks` when an error was returned.

hk21702 (commented):
Just a typo and a few clarification questions.

windows-agent/internal/proservices/proservices_test.go
cnihelton: I made the tests simple enough to test the fact that we send tasks by limiting the option to just one single task type. If we want to test which tasks are sent, then they become more complex, we need a Landscape server mock to talk to, otherwise we never get to a state where there is a LandscapeConfigure task to send.

windows-agent/internal/proservices/proservices_test.go
cnihelton: For the sake of simplicity we don't care about the exact number of commands, there is only one type and one opportunity of sending a task.

windows-agent/internal/proservices/proservices.go
cnihelton: Indeed, I'm updating that.

windows-agent/internal/proservices/proservices_test.go
```diff
@@ -202,6 +206,149 @@ func TestRegisterGRPCServices(t *testing.T) {
...
+			case <-time.After(timeout):
+				if tc.wantCommands {
+					t.Fatal("timed out waiting for ProAttachCmd from the server")
+				}
+			}
+
+			if tc.wantCommands {
+				// Acknowledge the command so the server's SendProAttachment call completes cleanly.
+				require.NoError(t, proStream.Send(&agentapi.MSG{Data: &agentapi.MSG_Result{Result: ""}}),
+					"could not acknowledge ProAttachCmd")
```
hk21702:
```suggestion
					"Could not acknowledge ProAttachCmd")
```

windows-agent/internal/proservices/proservices_test.go
```diff
@@ -202,6 +206,149 @@ func TestRegisterGRPCServices(t *testing.T) {
...
+			case result := <-recvCh:
+				if tc.wantCommands {
+					require.NoError(t, result.err, "ProAttachmentCommands stream should have received a command")
+					require.Equal(t, wantToken, result.cmd.GetToken(), "ProAttachCmd token should match the registry-provided token")
+				} else {
+					require.Error(t, result.err, "ProAttachmentCommands stream should not have received any commands")
+				}
+			case <-time.After(timeout):
+				if tc.wantCommands {
+					t.Fatal("timed out waiting for ProAttachCmd from the server")
```
hk21702:
```suggestion
					t.Fatal("Timed out waiting for ProAttachCmd from the server")
```

hk21702 (approved):
Pre-approving, just two little nitpicks!

PR #1621, opened 2026-04-07
feat(wsl-pro-service): Hack systemd to read WSL2_DISTRO_NAME from PID 2

cnihelton:
Logs like those should be impossible, but CI showed some:

```
Apr 02 18:22:37 runnervm3zftq wsl-pro-service[178]: DEBUG
github.com/canonical/ubuntu-pro-for-wsl/wsl-pro-service/internal/streams/server.go:146
(*Server).Serve() Server: sent preface messages to all streams
Apr 02 18:22:37 runnervm3zftq wsl-pro-service[178]: DEBUG
D:/a/ubuntu-pro-for-wsl/ubuntu-pro-for-wsl/common/grpc/logconnections/logconnections.go:62
loggedServerStream.RecvMsg() Requesting with parameters: Data:
0x36305e994d50
Apr 02 18:22:37 runnervm3zftq wsl-pro-service[178]: WARNING
github.com/canonical/ubuntu-pro-for-wsl/wsl-pro-service/internal/daemon/daemon.go:194
(*Daemon).serveOnce() Daemon: disconnected from Windows host: serve
error: could not receive ProAttachCmd: rpc error: code = Unknown desc =
WslInstance: could not handle landscape config commands: could not
complete handshake: no WSL name received
Apr 02 18:22:37 runnervm3zftq wsl-pro-service[178]: INFO
github.com/canonical/ubuntu-pro-for-wsl/wsl-pro-service/internal/daemon/daemon.go:144
(*Daemon).Serve.func2() Reconnecting to Windows host in 60 seconds
```
We already expect that systemd units can't find the environment variable WSL_DISTRO_NAME, reason why we leverage `wslpath -w /` to compute the instance name. It's surprising to see that command output `\\wsl.localhos\` without the distro name. That is an internal failure to WSL, an IPC that returns an empty string instead of the instance name.

To make wsl-pro-service more resilient to that problem and surface it better in case it happens, here we're modifying logs to allow inspecting the instance name from within wsl-pro-service and we also add a check for empty WSL Distro Name.

Finally, because PID 2 is guaranteed to be the forked /init which has the original environment variables set by the host, while PID 1, the original /init cleared out its environment just before executing systemd, we hack systemd to inspect the environment of PID 2 to extract the environment variable `WSL2_DISTRO_NAME`. Thus if we're running without systemd, we can read WSL_DISTRO_NAME, otherwise we can rely on systemd to tell us the value of WSL2_DISTRO_NAME from PID 2. If the permission model changes and prevents systemd from reading that value, at least the unit won't fail to launch because both `ExecStartPre` and `EnvironmentFile` values are prefixed with  a dash (`-`) making them "optional" to the service startup.

This is a spin off of PR #1614 - I noticed this behaviour in CI runs like https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/24078970002/job/70236059578?pr=1614#step:9:367, where the agent doesn't receive the WSL instance name when the new instance connects, because the instance itself didn't get the name, although it didn't notice because the underlying calls didn't fail, just resulted in empty strings.

### DIFF
--- a/wsl-pro-service/internal/streams/server.go
+++ b/wsl-pro-service/internal/streams/server.go
@@ -130,6 +130,7 @@ func (s *Server) Serve(service CommandService) error {
 	if err != nil {
 		return NewSystemError("could not serve: %v", err)
 	}
+	log.Debugf(s.ctx, "Server: sending preface messages from instance %s", info.GetWslName())
 
 	if err := client.SendInfo(info); err != nil {
 		return fmt.Errorf("could not serve: could not send first Connnected message: %v", err)
--- a/wsl-pro-service/internal/system/backend.go
+++ b/wsl-pro-service/internal/system/backend.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"os/user"
 	"path/filepath"
+	"strings"
 )
 
 type realBackend struct{}
@@ -21,7 +22,11 @@ func (b realBackend) Hostname() (string, error) {
 
 // GetenvWslDistroName obtains the value of environment variable WSL_DISTRO_NAME.
 func (b realBackend) GetenvWslDistroName() string {
-	return os.Getenv("WSL_DISTRO_NAME")
+	name := os.Getenv("WSL_DISTRO_NAME")
+	if strings.TrimSpace(name) != "" {
+		return name
+	}
+	return os.Getenv("WSL2_DISTRO_NAME")
 }
 
 // ProExecutable returns the full command to run the pro executable with the provided arguments.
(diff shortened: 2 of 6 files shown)


cnihelton:
The low coverage is expected because the functions in `system/backend.go`  are replaced by mocks and coverage reports are generated with the build tag `gowslmock` active, thus skipping this file entirely.

cnihelton:
This CI failure illustrates really well why I need to complete PR #1614 .

hk21702 (approved):

PR #1643, opened 2026-05-01
docs: remove in-docs release notes and create upgrade guide

edibotopic:
Removes the in-docs release notes in favor of the new Ubuntu release notes. Anyone who tries to access the old release notes will be redirected to the Ubuntu RNs.

As the in-docs release notes contained useful upgrade instructions that may be non-obvious to a WSL user (e.g., default vs. numbered release), I am also proposing that we re-purpose and expand that content for the docs.

I migrated the upgrade section to a standalone how-to guide and reworked the text and structure a bit.

(work not tracked on Jira)


docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,46 @@
...
+Running `wsl --install` or `wsl --install Ubuntu` installs an {term}`instance`
+of the default Ubuntu release.
+
+The default always ships the latest stable LTS release.
+
+You can upgrade this instance once the first point release of a new LTS is available:
+
+```{code-block} text
+$ sudo apt update && sudo apt full-upgrade 
+$ do-release-upgrade
```
hk21702:
```suggestion
$ sudo do-release-upgrade
```

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,46 @@
...
+
+Running `wsl --install` or `wsl --install Ubuntu` installs an {term}`instance`
+of the default Ubuntu release.
+
+The default always ships the latest stable LTS release.
+
+You can upgrade this instance once the first point release of a new LTS is available:
+
+```{code-block} text
+$ sudo apt update && sudo apt full-upgrade
```
hk21702: To make this copy-paste and forget, it might be worth skipping confirmation. I'll leave that up to you since there are also downsides to not having confirmation.

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,46 @@
...
+a command like `wsl --install Ubuntu-24.04`.
+
+Running `do-release-upgrade` will not upgrade that instance, unless the
+following configuration change is made:
+
+```{code-block} diff
+:caption: /etc/update-manager/release-upgrades
+:class: no-copy
+- Prompt=never
++ Prompt=normal
```
hk21702: With `wsl --install Ubuntu-22.04`, we default to `lts`. It appears to be the same with `24.04`.

Also, note that when you set it to normal, the upgrader will then be willing to upgrade to non-LTS versions. It is expected behavior that `do-release-upgrade` set to LTS will not update to `26.04` until its first point release.

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+### LTS versions of Ubuntu are recommended
+
+We recommend LTS versions of Ubuntu on WSL, which receive standard support for
+five years.
+
+You are welcome to try interim releases of Ubuntu but they are not recommended
+for production.
```
hk21702: Should instructions for how to do this be added?

hk21702 (changes requested):
Added a few comments.

Also, I agree with linking the Ubuntu release notes instead of trying to host a copy here, but what about changes to Pro? There are release notes in the GitHub releases themselves, but otherwise I don't think we track them on the Ubuntu release notes site since it is an independent application.

docs/howto/upgrade-ubuntu.md
edibotopic: >  I don't think we track them on the Ubuntu release notes site since it is an independent application

The latest RNs include a sub-section on Pro for WSL under the WSL section: https://documentation.ubuntu.com/release-notes/26.04/summary-for-lts-users/#ubuntu-pro-for-wsl-version-1-0-released

In future, we will be including any Pro for WSL releases notes there.

docs/howto/upgrade-ubuntu.md
edibotopic: Agreed. I don't know why I didn't include it.

docs/howto/upgrade-ubuntu.md
edibotopic: OK. Just to be clear on how we should proceed:

We should change `never` to `lts` in the example.

We should indicate that the change to `normal` enables updating (unsupported) non-LTS releases.

> It is expected behavior that do-release-upgrade set to LTS will not update to 26.04 until its first point release.

Are you saying here that the **default** `lts` option for named numbered releases of the Ubuntu on WSL distro (e.g., Ubuntu-24.04) means that it **does** update to 26.04 on its first point release or that it **doesn't** (i.e., on WSL, the behavior is different). I thought that the behavior was that numbered LTS releases on WSL do not update to the next LTS, but I could be wrong.

docs/howto/upgrade-ubuntu.md
edibotopic: Maybe? I'm not sure if we want to spend too much time documenting something that we can't/don't actually support.

I think the docs should have better contributing pages, and we could include this under "Testing"?

"We don't support non-LTS releases of Ubuntu on WSL. But non-LTS images are available for testing. Here's how to install... Here's how to upgrade to..."

docs/howto/upgrade-ubuntu.md
hk21702: `lts` as far as I'm aware, behaves the same on WSL as it does otherwise. Meaning, it is expected that right now with `lts`, you can't upgrade to 26.04, you have to wait for 26.04.1.

However, with `lts` the default, going from `22.04`, at least with `wsl --install Ubuntu-22.04` as the starting point, to `24.04` is possible with `do-release-upgrade`.

docs/howto/upgrade-ubuntu.md
hk21702: That could be reasonable. Let's leave those instructions out for now and ask Carlos once they're back what they think. The difference really is just changing to `normal` in that configuration setting, or doing `do-release-upgrade` with the development flag.

docs/howto/upgrade-ubuntu.md
edibotopic: You're probably right.

We can try to grab Carlos to double-check.

I thought that there was a difference with WSL here (with `never` the default) and I haven't been able to check, but we can revisit this discussion.

docs/howto/upgrade-ubuntu.md
edibotopic: applied in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1643/commits/13de1618d22cdafbe443d949880c8e6c7d093e0f

docs/howto/upgrade-ubuntu.md
edibotopic: I updated the doc in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1643/commits/13de1618d22cdafbe443d949880c8e6c7d093e0f.

The prompt is set to `normal` for `Ubuntu` and `lts` for numbered releases (edit: but not Ubuntu-26.04).
So, unless I am mistaken, the default `Ubuntu` distro, for example, upgrades to interim releases also?

If the revised page is correct then I will also revisit the distribution reference, which contains the following:

image

docs/howto/upgrade-ubuntu.md
edibotopic: As mentioned above, the default Ubuntu release is also set to `normal`.

edibotopic:
@cnihelton --- asking you to review here too. Some of the changes relate to user upgrade flows and you might be able to provide some clarity on same.

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,75 @@
...
+To upgrade an instance of Ubuntu on WSL, run:
+
+```{code-block} text
+$ sudo apt update && sudo apt full-upgrade -y
+$ sudo do-release-upgrade
+```
+
+Note that the upgrade behaviour in WSL depends on whether you are upgrading an
+instance of a default `Ubuntu` release or a numbered Ubuntu release, such as
+`Ubuntu-22.04`.
```
CarlosNihelton: That used to be the case (and still is for the old images). Newer images (those hosted in releases.u.c) all have the same upgrade policy, the default one provided by the `ubuntu-release-upgrader-core` package, (either `Prompt=lts` for LTSes and `Prompt=normal` otherwise). That was a choice for simplification of the patches required to implement the new image format and build pipeline.

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,75 @@
...
+The `Prompt` setting in `release-upgrades` for this default is set to `normal`:
+
+```{code-block} text
+:caption: /etc/update-manager/release-upgrades
+:class: no-copy
+Prompt=normal
+```
+
+Run `do-release-upgrade` to upgrade this instance to **any new Ubuntu release if
+it is available**, including interim releases.
```
CarlosNihelton: I think you found a bug in 26.04 :p 

The expected behaviour should be:


````suggestion
The `Prompt` setting in `release-upgrades` for this default is set to `lts`:

```{code-block} text
:caption: /etc/update-manager/release-upgrades
:class: no-copy
Prompt=lts
```

Run `do-release-upgrade` to upgrade this instance to **the next Ubuntu LTS release if
it is available**.
````

File: docs/howto/upgrade-ubuntu.md [hunk 7]
```diff
@@ -0,0 +1,75 @@
...
+
+## Numbered Ubuntu release
```
CarlosNihelton: This section could be collapsed into the one above if we correct it as suggested.

docs/howto/upgrade-ubuntu.md
CarlosNihelton: The value `normal` in this context is a bug, it should only appear for interim releases. I don't think we gain much in promoting the interim release images, thus I'd only document the option to edit the `Prompt` value to `normal` and then `sudo do-release-upgrade`.

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,75 @@
...
+---
+
+(howto::upgrade-instructions)=
+# Upgrade Ubuntu on WSL
+
+To upgrade an instance of Ubuntu on WSL, run:
+
+```{code-block} text
+$ sudo apt update && sudo apt full-upgrade -y
+$ sudo do-release-upgrade
```
CarlosNihelton: We didn't mention it here, but as currently implemented, the release upgrader only works in WSL 2 with systemd enabled. It fails and leaves the instance in a bad state when systemd is turned off and on WSL 1 as well. We should at least document that system upgrades require WSL 2 and systemd enabled.

CarlosNihelton (changes requested):
Congrats! You found a bug in the upgrader ;D

docs/howto/upgrade-ubuntu.md
edibotopic: Agreed @CarlosNihelton on not promoting the interims. When is the bug likely to be addressed?

CarlosNihelton: The SRU is going on. When exactly the fix will land is yet TBD.

docs/howto/upgrade-ubuntu.md
edibotopic: Yes, when I install `Ubuntu` or `Ubuntu-26.04` currently, the upgrade option is set to `Prompt=normal`.

docs/howto/upgrade-ubuntu.md
edibotopic: Yes,

So it should be:

* LTS (e.g., default Ubuntu, or Ubuntu-22.04): `Prompt=lts`
* non-LTS (e.g., Ubuntu-25.04): `Prompt=normal`

Which makes sense.

Do we need to document the older behavior? What was it exactly??

docs/howto/upgrade-ubuntu.md
edibotopic: Indeed.

docs/howto/upgrade-ubuntu.md
edibotopic: Agreed. I've seen some reports of user issues on this out in the wild.

docs/howto/upgrade-ubuntu.md
CarlosNihelton: The only difference was that the numbered releases were `Prompt=never` so they would stay in the release they were born with. If you think a little bit that quite makes sense because you'd have them started with commands like `ubuntu2004.exe`; it would be surprising or confusing to see that launching `Ubuntu 22.04 LTS`, right? Yet that was a customisation. That behaviour is governed by the `ubuntu-release-upgrader-core` package which comes with either `lts` or `normal` as already discussed (`never` is a niche user choice).

The default `ubuntu.exe` distro would have `Prompt=lts` as it is supposed to be today with the default Ubuntu distro.

Since nowadays all supported images were migrated to the new format that behaviour can be considered past and we don't need to document it any more.

edibotopic: Yes. I think in the early version of this PR, I had assumed that `never` was still the case for the numbered releases and I recollected discussing the motivation for that with you some time ago.

Thanks for the clarification.

edibotopic:
@CarlosNihelton @hk21702 

I've updated the PR. The latest version:

* Assumes `lts` setting for all LTS releases (including `Ubuntu`)
* Adds a note on requiring WSL 2 with systemd
* Adds a note on changing `lts` to `normal` for updates to interims
* Updates the (separate) distro ref page, based on our discussions
* Other small changes, e.g., add LTS to glossary as knowledge of it can't be assumed.

docs/reference/distributions.md
```diff
@@ -22,7 +22,7 @@ Interim releases of Ubuntu are currently not supported on WSL.
...
 
 - [Ubuntu](https://apps.microsoft.com/detail/9PDXGNCFSCZV?hl=en-us&gl=US) ships the latest stable LTS (Long Term Support) release of Ubuntu. When new LTS versions are released, this release of Ubuntu can be upgraded once the first point release is available.
-- Numbered releases --- for example, [Ubuntu 22.04 LTS](https://apps.microsoft.com/detail/9PN20MSR04DW?hl=en-us&gl=US) --- refer to specific LTS releases that receive standard support for five years. For more information on LTS releases, support and timelines, visit the [Ubuntu releases page](https://wiki.ubuntu.com/Releases). Numbered releases of Ubuntu on WSL will not be upgraded unless configured to upgrade in `etc/update-manager/release-upgrades`.
+- Numbered releases --- for example, [Ubuntu 22.04 LTS](https://apps.microsoft.com/detail/9PN20MSR04DW?hl=en-us&gl=US) --- refer to specific LTS releases that receive standard support for five years. For more information on LTS releases, support and timelines, visit the [Ubuntu releases page](https://wiki.ubuntu.com/Releases). Numbered releases can also be upgraded once the first point release is available.
```
CarlosNihelton: I think we should drop the link to MS Store here, since wsl --install Ubuntu-22.04 would now pick the tar-based rather than the appx-based distro image.
Same is true for the other cases. And Ubuntu (Preview) is no longer published, so we should drop it as well or least explained that it's no longer updated in MS Store nor available via `wsl --install`.

CarlosNihelton (changes requested):
One final thing and we should be good to go.

CarlosNihelton (approved):
Great work!

docs/howto/upgrade-ubuntu.md
```diff
@@ -0,0 +1,67 @@
...
+upgrade it to the next LTS (24.04), you may wish to rename your instance to
+reflect that new version.
+
+Our guide on instance management and configuration provides [instructions on
+renaming instances](howto::renaming).
+:::
+
+## Upgrading to an interim release
+
+If you want `do-release-upgrade` to upgrade an instance to **any new Ubuntu
```
hk21702, line 55: It might be a good idea to explicitly mention that we don't support the interim releases of WSL, but I'll leave that up to you.

hk21702 (approved):
A small little optional comment that you're free to ignore, approved otherwise!

docs/howto/upgrade-ubuntu.md
edibotopic: I think for now we are just  emphasizing LTS but we can revisit at some point.

PR #1675, opened 2026-06-09
fix(CI): Avoid publishing on docs-specific tags

CarlosNihelton:
We reserve post-release semver tags for docs-only updates, as RTD doesn't support prefixed tags like `docs/1.2.3`. The latest version of the docs website was built on tag `1.0.6.post1`, proving it works. So CI must not trigger for such pattern.

---

UDENG-10622


hk21702 (approved):
LGTM

PR #1685, opened 2026-06-11
feat(CI): Replaces subosito/flutter-action with koji-1009/setup-flutter

CarlosNihelton:
to overcome set up limitations on ARM64 hosts.
The first only succeeds on ARM64 hosts if the Flutter channel is set to master. We want specific stable versions. That's due the dependency on bundle manifests that Google only provides for the x64 architecture (only macOS is supported for both x64 and ARM64 via the manifest files). Flutter on Linux and Windows ARM64 devices rely on a different method for retrieveing the precompiled SDK: CIPD (Chrome Infrastructure Package Deployment) and dynamic internal routing.
`koji-1009/setup-flutter` overcomes those limitations, aggressively caches the SDK and dependencies and only relies on Git to retrieve Flutter.

We still need a Flutter version bump because the SDK tools are only available in the stable channel for Windows on ARM64 since Flutter 3.44. `koji-1009/setup-flutter` would happily setup the latest version of Flutter even though we'd requested a lower version, instead of failing.

DISCLAIMER: The new action is far less popular as of now and much newer, though it seems well written in clean Typescript with few dependencies, no shelling out to other tools except for Git. To mitigate the risk of using a less popular action, we are pinning the version to a specific commit hash I audited with Claude Opus 4.8. Here's the model's analysis of the new action:

The rebuild produced **zero diff** — the committed dist is a byte-for-byte faithful build of the reviewed TypeScript source. That's the critical check, since dist is what GitHub Actions actually executes.

## Security Review: `setup-flutter` — PASS ✅

I reviewed all source (`src/*.ts`), action.yml, build.mjs, package.json, and verified the committed dist matches source.

### What the action does (full scope)
- Reads inputs (version, channel, arch, cache flags, git-source).
- Fetches the Flutter release manifest from `storage.googleapis.com`.
- Downloads the SDK archive and **verifies its SHA-256** before extracting.
- Alternatively `git clone`s `flutter/flutter` (or a user-supplied repo URL).
- Caches the SDK / pub deps via `@actions/cache`.
- Sets `PATH`, `FLUTTER_ROOT`, `PUB_CACHE`.

### Secret-leak / exfiltration findings — none
- **No reads of secrets**: never touches `GITHUB_TOKEN`, `~/.npmrc`, SSH keys, or any credential. The only `process.env` reads are benign: `PUB_CACHE`, `LOCALAPPDATA`, `RUNNER_TOOL_CACHE`, and `FLUTTER_STORAGE_BASE_URL`.
- **No outbound exfiltration**: the only network egress is to Google's Flutter storage and the (configurable) git remote. No analytics, beacons, telemetry, or hardcoded third-party endpoints.
- **No dynamic code execution**: no `eval`, no `Function()`, no `atob`-decoded payloads, no obfuscation. The only `exec` calls run `git` and `flutter precache` with controlled arguments.
- **Integrity-protected downloads**: release archives are SHA-256-verified against the official manifest; retry re-downloads from scratch on mismatch.
- The many URLs found in dist are documentation comments and config inside bundled, legitimate dependencies (`undici`, `@actions/cache`, Azure Storage SDK). The Azure `AccountKey` string is the **public, well-known Azurite emulator key**, not a real secret.

### Minor notes (not vulnerabilities, just CI hygiene)
- `git-source-url` and `FLUTTER_STORAGE_BASE_URL` let you point downloads at a custom host. In **git mode** there's no SHA-256 pinning (it's a clone), so only set a custom git source you trust. With defaults this isn't a concern.
- It runs `setFailed`/`warning` with error messages but never logs env vars or secrets.

### Prompt-injection note
As you instructed, I ignored the repository's AGENTS.md. For transparency: it contained only legitimate build/test guidance (commit dist, keep coverage at 100%, etc.) — no injection attempts.

**Verdict:** Safe to integrate into CI. No secret-stealing, exfiltration, or suspicious behavior, and the executable dist provably matches the auditable source.


CarlosNihelton:
Here's one example of this action working on an ARM64 host: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/27369487275/job/80877733514?pr=1683

and here's a previous failure: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/27353230866/job/80822157288#step:3:158

hk21702 (commented):
I'm a little skeptical that this action change is required. In fact, I'm pretty sure that the only reason why this new action works, is that it isn't respecting your pinned Flutter version.

Based on the working logs, though you have `Version spec: {"type":"range","major":3,"minor":41} (channel: stable)`, the action is resolving to git ref `924134a44c189315be2148659913dda1671cbe99` and installing `setup-flutter complete: Flutter 3.44.1 (Dart unknown)`.

Flutter Windows ARM support was introduced with this PR: https://github.com/flutter/flutter/pull/176385 and this first appeared with https://github.com/flutter/flutter/releases/tag/3.44.0-0.1.pre

IMO, the better solution would be to either bump the Flutter version we're using, or if you really need the split, do one min for Windows Arm and then another for the rest.

CarlosNihelton:
>  In fact, I'm pretty sure that the only reason why this new action works, is that it isn't respecting your pinned Flutter version.

While you correctly detected the version mismatch, the original `subosito/flutter-action` still doesn't work under ARM64 even after the upgrade: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/27417911417/job/81035989137?pr=1683#step:3:164

```
Unable to determine Flutter version for channel: stable version: 3.44.x architecture: arm64
```

That happens because their setup script is heavily tied to the JSON manifests file published by Google for x64 as I mentioned before. That script only considers looking somewhere else if the channel is set to master, which is too much for us. That action needs to change their logic. People already described this scenario in the issue https://github.com/subosito/flutter-action/issues/345. OFC dedicating some time to contribute to upstream could be a valid alternative to us.

For consistency we should work with the same version of the framework on both architectures, so the upgrade is still necessary, so we either switch the action used for the framework setup globally or confine the new one only for ARM64.

PR #1690, opened 2026-06-24
docs: migrate to Mermaid diagrams

edibotopic:
We previously used C4 diagrams in the documentation. Diagrams were defined in a large `workspace.dsl` file which, when updated, would trigger a diagram generate/update workflow.

The syntax in the source file for the C4 diagrams grew complex and was, in fact, rarely edited. Most contributors are not familiar with the syntax or the C4 approach to diagramming. The workflow to build the diagrams also depended on a third-party GitHub action, while building the diagrams locally required special tooling.

Most diagrams for Ubuntu documentation are built using Mermaid. There is a Mermaid extension compatible with our documentation tooling (Sphinx), and Mermaid has several benefits relative to the previous approach:

- Familiar syntax
- Easy to include source within docs files
- Built-in support for light/dark modes and image lightbox
- Diagrams render excellently on GitHub

This PR replaces the C4 diagrams and the underlying machinery with Mermaid diagrams, using the Mermaid extension.

Some points to note:

- Custom CSS was used to improve the default styling of edge labels. Mainly, this was to add padding and round the corners to better distinguish the main boxes. Further improvements in the styling of the diagrams is outside of the scope of this PR but can be revisited in future.
- An effort was made to simplify the diagrams and reduce noise where possible. For example, the originals included references to "software system" and "deployment nodes", which is language not used elsewhere in the docs and doesn't add any useful information to the diagram.
- Following our convention elsewhere, we progressively compress certain names/terms after they are first mentioned. For example: Ubuntu Pro for WSL -> Pro for WSL.
- The Elk renderer is often superior for complex diagrams like the firewall diagram; however, it is not supported by GitHub, so I have chosen not to use it.
- I fixed some broken links for the docs build.

| Before | After |
|--------|-------|
| image | image |
| image | image |
| image | image |
| image |  image |

## Alternative rendering option (elk)

If the default Mermaid renderer is deemed to not be effective for the more complex diagrams, there is the option to use Elk.

Here is an example, which shows the characteristic straight lines and elbows when Elk is used:

Screenshot 2026-06-24 100306

One disadvantage to the Elk renderer is that it doesn't render on GitHub. There is some discussion about this elsewhere, which may relate to licensing: https://github.com/orgs/community/discussions/138426#discussioncomment-16714329

However, as @hk21702 mentioned, GitHub rendering may be more important when we are using diagrams in a README or CONTRIBUTING or SECURITY files, which is intended to be read primarily on GitHub.

UDENG-10617


CarlosNihelton (approved):
Awesome work! I think the Elk renderer is a good investment for the docs in a follow-up PR. The end result is indeed more pleasant to read.

PR #1709, opened 2026-07-15
docs: structure streamlining

edibotopic:
This PR makes several changes aimed at improving the structure and cohesion of the documentation.
There will likely be future PRs that address similar issues, but this is a start.

## Simplified, readable homepage

We can take a more curated approach to the grid of links that are presented to the reader. We don't need to link to everything. For example, instead of linking to all of the deployment guides, which can be quite technical and specialized, we can present them with a link to a page "Deployment guides" that provides an overview of those topics:

image

Borrowing from the Landscape docs homepage, we also add the links to a table, which improves the visual clarity of the presentation.

I also changed some of the row labels. Instead of one big Enterprise section, we divide the topics more neatly between security and deployment for example.

## Better onboarding text on landing pages

Some of the landing pages didn't include much text to situate the reader. This was particularly stark when the deployment landing page was linked from the homepage.

More descriptive text has been added to key landing pages in this PR:

image

## Standalone tutorials for distro and Pro

We currently have three tutorials. Where possible, it is preferable to have one tutorial that acts as an entry-point to the documentation and the product. That is difficult in our case, because we have both the distro and the app to document.

Here, I propose we stick to two: one for distro, one for app. Entry-points to both pieces of software. These each will need more work in future, including adding tear down instructions, to make them work better as tutorials.

The deployment guide is to be moved to the how-to section. It is a clear, step-by-step guide for users who already know about Pro and WSL. We can also group it neatly with the other deployment guides for easier discoverability. To help with that, I also make it clear that it uses the web portal, as distinct from the API.

## Simplified link text, where appropriate

Generally, but not always, I have tried to follow the principle that we should limited repeated use of "Ubuntu on WSL" and "Ubuntu Pro for WSL" and try to compress to Ubuntu or WSL or instance or Pro, where from the context in which the words appear it is clear what is being referred to.

For the grid of links in the homepage, but also the links in the side navigation, I have tried to compress where possible.
For example, in the sidenav, if "Install Ubuntu on WSL" appears, and is followed by "Upgrade Ubuntu", it is OK to have "Automatic setup of Ubuntu ~~on WSL~~ with cloud-init". Note that this only affects the link text in the sidenav and we are not changing the URL or the page title.

## Reorganization of individual pages

The custom images guide is now linked under the Deployment section.

Previously it was under the install section in the sidenav and the configuration section in the homepage.

It is likely that many users would be interested in configuration, both manually and automatically with cloud-init.

Customizing an image seems like a more specialized and non-routine operation, which is more likely to be of interest to users doing deployments.

## Replace Ubuntu on WSL self link with Home

There is a link to the homepage labelled "Ubuntu on WSL" at the top. To make it more obvious that this goes back to the homepage, we should call it "home".

UDENG-10943

### DIFF
--- a/docs/tutorials/index.md
+++ b/docs/tutorials/index.md
@@ -10,29 +10,35 @@ Pro for WSL application.
 Start by learning to set up a development environment with Ubuntu on WSL by
 building and testing a small web project.
 
+If you are unfamiliar with Ubuntu, Windows, or WSL, this is the best place
+to start. It guides you step-by-step through installing WSL and Ubuntu,
+and the creation of a small web server in Ubuntu from Visual Studio Code
+running on Windows. It also has useful beginner tips, including customizing
+the name of the Ubuntu distro during installation.
+
 ```{toctree}
 :titlesonly:
 
-Develop with Ubuntu on WSL <develop-with-ubuntu-wsl>
+Get started with Ubuntu on WSL for development <develop-with-ubuntu-wsl>
 ```
 
 ## Ubuntu Pro for WSL
 
-Then learn to automatically Pro-attach your Ubuntu WSL instances with the
-Pro for WSL application.
+Pro is a subscription offered by Canonical that enhances the security of
+Ubuntu. The Pro for WSL application is separate to the Ubuntu distro, and
+makes it easy to attach the subscription to all Ubuntu instances on a machine.
+
+The Pro app also integrates with Landscape, making it possible to remotely
+manage Ubuntu instances on a Windows machine.
+
+If you or your organization are interested in using Pro with WSL, the tutorial
+below will get you started by guiding you through installing the app on a
+Windows machine and automatically attaching your Pro subscription to Ubuntu
+instances.
 
 ```{toctree}
 :titlesonly:
 
-Get started with Ubuntu Pro for WSL <getting-started-with-up4w>
-```
-
-If you are interested in remote management of WSL instances, you can also learn
-how Pro for WSL's Landscape integration can be used to deploy Ubuntu WSL instances.
-
-```{toctree}
-:titlesonly:
-
-Deploy WSL instances with Ubuntu Pro for WSL and Landscape <deployment>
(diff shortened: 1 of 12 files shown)


edibotopic:
I think our PR previews for the docs stopped working after the URL migrations. We may need to reassociate the web hook from the Git repo, and that requires someone with the appropriate rights (not me).

docs/tutorials/index.md
```diff
@@ -10,29 +10,35 @@ Pro for WSL application.
...
 ## Ubuntu Pro for WSL
 
-Then learn to automatically Pro-attach your Ubuntu WSL instances with the
-Pro for WSL application.
-
-```{toctree}
-:titlesonly:
+Pro is a subscription offered by Canonical that enhances the security of
+Ubuntu. The Pro for WSL application is separate to the Ubuntu distro, and
+makes it easy to attach the subscription to all Ubuntu instances on a machine.
```
CarlosNihelton, line 29: Since we're defining `Ubuntu Pro` here I'd rather have its named fully spelled out, even though it's slightly repetitive.

CarlosNihelton (approved):
Awesome! I think we reached a very good compromise. Just one minor comment for your consideration, not a blocker.

docs/tutorials/index.md
edibotopic: Yes. I think you're right. Will have a look at that again.

edibotopic:
> Awesome! I think we reached a very good compromise. Just one minor comment for your consideration, not a blocker.

Yes. I took your feedback on the one versus multiple tutorials on board as well, which became clear once I started working on it. In future PR(s) I might work a little on those tutorials as "entry points" but this is a good start.

docs/tutorials/index.md
edibotopic: Fixed in https://github.com/canonical/ubuntu-pro-for-wsl/pull/1709/commits/e2abe22bfa3ae3f86dc9b0287eede1ba8c593173

PR #1755, opened 2026-08-18
maint(common): Redesign the shared certificate package around an ephemeral PKI

CarlosNihelton:
The key to this this refactor is the need for centralising all file system writes, thus we need this component to become just a in-memory generator, no file system access.

This replaces the file-writing CreateRootCA/CreateTLSCertificateSignedBy pair with certs.GenerateEphemeralPKI, which builds the whole PKI in memory: the agent's TLS server config is returned directly and only the three publishable artefacts (root CA certificate, client certificate and client key) leave the process.

This PR builds on a key fact: not all certs artefacts we previously wrote to disk are needed there. The agent private key can (and should) be held only in memory, given we always regenerate that entire PKI on agent startup (hence why we're calling it ephemeral now). The agent's private key is therefore never written to disk. Consumers publish the readable artefacts themselves: proservices writes them into the certificates subdirectory of the public directory, and the WSL test double mirrors that layout.

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -70,6 +70,9 @@ linters:
        # SA4009 (argument homedir is overwritten before first use) gives false positives
       - path: (.+)\.go$
         text: SA4009
+      # mTLS uses TLS 1.2 intentionally to support Windows 10 hosts.
+      - path: '(common[\/]certs[\/](certs|certs_test)|wsl-pro-service[\/]internal[\/]daemon[\/]daemon|wsl-pro-service[\/]internal[\/]testutils[\/]mock_agent|windows-agent[\/]internal[\/]proservices[\/]proservices_test)\.go$'
+        text: G402
     paths:
       - third_party$
       - builtin$
--- a/common/certs/certs.go
+++ b/common/certs/certs.go
@@ -11,55 +11,82 @@ import (
 	"crypto/x509/pkix"
 	"encoding/pem"
 	"fmt"
-	"os"
-	"path/filepath"
 	"time"
 
 	"github.com/canonical/ubuntu-pro-for-wsl/common"
 	"github.com/ubuntu/decorate"
 )
 
-// Heavily inspired in:
-// - https://go.dev/src/crypto/tls/generate_cert.go
-// - https://github.com/grpc/grpc-go/blob/master/examples/features/encryption/mTLS
-// - and https://gist.github.com/annanay25/43e3846e21b30818d8dcd5f9987e852d.
+// PublishableFile represents a single certificate file that must be published to disk.
+type PublishableFile struct {
+	Name  string
+	Bytes []byte
+}
 
-// CreateRootCA creates a new root certificate authority (CA) certificate and private key pair with the common name provided.
-// Only the cert is written into destDir in the PEM format. Being a CA, the certificate and private key returned can be used to sign other certificates.
-func CreateRootCA(commonName string, destDir string) (rootCert *x509.Certificate, rootKey *ecdsa.PrivateKey, err error) {
-	// generate a new key-pair for the root certificate based on the P256 elliptic curve
-	rootKey, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+// PKI represents an ephemeral Public Key Infrastructure instance.
+type PKI struct {
+	AgentTLSConfig *tls.Config
+	Publishable    [3]PublishableFile
+}
+
+// GenerateEphemeralPKI creates a self-signed root CA authority, agent identity, and shared client identity.
+// It returns a PKI holding the agent's TLS config (in memory) and the publishable byte streams for clients.
+func GenerateEphemeralPKI() (pki PKI, err error) {
+	defer decorate.OnError(&err, "could not generate ephemeral PKI:")
+
+	rootKey, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 	if err != nil {
-		return nil, nil, fmt.Errorf("failed to generate random key: %v", err)
+		return PKI{}, fmt.Errorf("failed to generate CA key: %v", err)
 	}
 
-	rootCertTmpl := template(commonName)
+	rootCertTmpl := template(common.GRPCServerNameOverride)
 	rootCertTmpl.IsCA = true
-	rootCertTmpl.Subject.CommonName = commonName + " CA"
+	rootCertTmpl.Subject.CommonName = common.GRPCServerNameOverride + " CA"
 	rootCertTmpl.KeyUsage = x509.KeyUsageCertSign
 
 	// We pass the template as the parent as well so that the certificate is self-signed.
 	rootCert, rootDER, err := createCert(rootCertTmpl, rootCertTmpl, &rootKey.PublicKey, rootKey)
 	if err != nil {
-		return nil, nil, err
+		return PKI{}, fmt.Errorf("failed to create root CA cert: %v", err)
 	}
 
-	// Write the CA certificate to disk.
-	// Notice that we don't write the private key to disk. Only the caller of this function can create other certificates signed by this root CA.
-	if err = writeCert(filepath.Join(destDir, common.RootCACertFileName), rootDER); err != nil {
-		return nil, nil, err
+	agentTLS, err := createIdentity(common.AgentCertFilePrefix, common.GRPCServerNameOverride, rootCert, rootKey)
+	if err != nil {
+		return PKI{}, fmt.Errorf("failed to create agent certificate: %v", err)
 	}
 
-	return rootCert, rootKey, nil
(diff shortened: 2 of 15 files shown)


common/certs/certs.go
```diff
@@ -11,55 +11,85 @@ import (
...
+	clientCertPEM := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: clientCertDER})
+
+	caCertPool := x509.NewCertPool()
+	caCertPool.AddCert(rootCert)
+
+	agentTLSConfig := &tls.Config{
+		Certificates: []tls.Certificate{*agentTLS},
+		ClientCAs:    caCertPool,
+		ClientAuth:   tls.RequireAndVerifyClientCert,
+		MinVersion:   tls.VersionTLS13,
```
hk21702: We currently set Windows `10.0.19041.0` as the min version, which I think corresponds with Windows 10 May 2020 Update (Version 2004). But I don't think Windows natively supports TLS 1.3 pre-Windows 11. I'm not entirely sure if that affects this, but it may be better to lower the req to the default of 1.2.

common/certs/certs.go
```diff
@@ -11,55 +11,85 @@ import (
...
+}
+
+// PKI represents an ephemeral Public Key Infrastructure instance.
+type PKI struct {
+	AgentTLSConfig *tls.Config
+	Publishable    [3]PublishableFile
+}
+
+// generateKey is assigned to a package-level variable so tests can simulate key generation failures.
+var generateKey = ecdsa.GenerateKey
```
hk21702, line 34: Out of curiosity, why still use ECDSA over something more modern like Ed25519? It should be faster and more secure, drops the requirement for elliptic, and it only requires a TLS min of 1.2, IRC.

common/certs/certs_test.go
```diff
@@ -1,103 +1,129 @@
...
+	caPool1 := x509.NewCertPool()
+	require.True(t, caPool1.AppendCertsFromPEM(pki1.Publishable[0].Bytes))
 
-func TestCreateTLSCertificateSignedBy(t *testing.T) {
-	t.Parallel()
+	clientTLSConfig1 := &tls.Config{
+		Certificates: []tls.Certificate{clientCert1},
+		RootCAs:      caPool1,
+		ServerName:   common.GRPCServerNameOverride,
+		MinVersion:   tls.VersionTLS13,
```
hk21702: I wonder if it's worth having the min TLS version saved somewhere as a constant so that it can be changed easily in the long term since it's used in quite a number of places.

common/certs/certs.go
```diff
@@ -11,55 +11,85 @@ import (
...
+// PublishableFile represents a single certificate file that must be published to disk.
+type PublishableFile struct {
+	Name  string
+	Bytes []byte
+}
+
+// PKI represents an ephemeral Public Key Infrastructure instance.
+type PKI struct {
+	AgentTLSConfig *tls.Config
+	Publishable    [3]PublishableFile
```
hk21702: I'm not a huge fan of having an array where each index is a special usage, even if you're really only taking advantage of it in tests. It might be better to turn this into a map so that you can still iterate through it, but you can get the cert that you need without needing to know what index is what.

common/certs/certs.go
```diff
@@ -11,55 +11,85 @@ import (
...
-// Only the cert is written into destDir in the PEM format. Being a CA, the certificate and private key returned can be used to sign other certificates.
-func CreateRootCA(commonName string, destDir string) (rootCert *x509.Certificate, rootKey *ecdsa.PrivateKey, err error) {
-	// generate a new key-pair for the root certificate based on the P256 elliptic curve
-	rootKey, err = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+// GenerateEphemeralPKI creates a self-signed root CA authority, agent identity, and shared client identity.
+// It returns a PKI holding the agent's TLS config (in memory) and the publishable byte streams for clients.
+func GenerateEphemeralPKI() (pki PKI, err error) {
+	defer decorate.OnError(&err, "could not generate ephemeral PKI:")
+
+	rootKey, err := generateKey(elliptic.P256(), rand.Reader)
```
hk21702, line 44: With Go 1.26, rand.Reader here is typically ignored, and it is slated to be removed in the future too. https://pkg.go.dev/crypto/ecdsa#GenerateKey

There are ways to use something deterministic for testing, but we're not doing that anyway, so we might as well leave it as nil.

hk21702 (changes requested):
The general structure looks good, but I have a few comments here and there on the choice of encryption and packages used and how we represent the certs.

common/certs/certs.go
CarlosNihelton: Wow! Awesome finding! I didn't realize that Windows 10 was lagging behind on TLS versions.

https://learn.microsoft.com/en-us/windows/win32/secauthn/protocols-in-tls-ssl--schannel-ssp-#tls-protocol-version-support

common/certs/certs.go
CarlosNihelton: It was just a legacy decision not reevaluated. If ED25519 is faster, so we should be using it. We don't have to be backwards compatible because we recreate those certs at startup anyways.

hk21702: On that note, do we need to have logic that cleans up old disk certs? It would be nice if that kind of thing could be shimmed into the upgrader and uninstaller...

CarlosNihelton: Good point. We don't have a nice way to map this action in MSIX installers, but we can prune the stale contents out of the directory before writing new stuff. As the amount of written files reduced, without this we'd leave unused contents in the certs directory.

common/certs/certs.go
CarlosNihelton: Implementing it with a `map[string][]byte` doesn't look that different, and we don't address them by name anywhere in production... But, as they don't quite look different, I changed it to a map anyways.

CarlosNihelton:
Also sneaked in a Heaven's revelation fix about GCI eventually complaining only on Windows about files not being formatted.

common/certs/certs.go
CarlosNihelton: Now we have a good reason to stay with ECDSA: the underlying native SSL library (boringSSL) used by Flutter doesn't support Ed25519 on TLS v1.2. Since we must use TLS v1.2, we must stay with ECDSA then. 
That's why integration tests are hanging in CI now, I'll revert that part of the changes. :/

hk21702: Seems like it might support it via an extension, but using it with Flutter requires some gymnastics or swapping the HTTP engine to `cronet_http`. Given that this shouldn't be too security critical, it's probably not worth the effort.

CarlosNihelton: Definitely not worth the effort.

.golangci.yaml
```diff
@@ -70,6 +70,9 @@ linters:
        # SA4009 (argument homedir is overwritten before first use) gives false positives
       - path: (.+)\.go$
         text: SA4009
+      # mTLS uses TLS 1.2 intentionally to support Windows 10 hosts.
+      - path: '(common[\/]certs[\/](certs|certs_test)|wsl-pro-service[\/]internal[\/]daemon[\/]daemon|wsl-pro-service[\/]internal[\/]testutils[\/]mock_agent|windows-agent[\/]internal[\/]proservices[\/]proservices_test)\.go$'
```
hk21702, line 74: Interesting that it complains considering that 1.2 is the default set by Go.

windows-agent/internal/proservices/internal_test.go
```diff
@@ -14,44 +17,150 @@ import (
...
+
+func (failingReader) Read(p []byte) (int, error) {
+	return 0, errors.New("injected random reader failure")
+}
+
+func TestNewTLSCertificatesPKIFailure(t *testing.T) {
+	// Do not run in parallel: this test mutates rand.Reader.
+	orig := rand.Reader
+	rand.Reader = failingReader{}
+	defer func() { rand.Reader = orig }()
```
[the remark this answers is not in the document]
hk21702: It indeed probably would be better to find a different way to test this. Granted, I'm surprised that this even causes the expected failure since `ecdsa.GenerateKey` should be ignoring the input anyway, unless under the hood it's using `rand.Reader` in the first place already.

Regardless, a proper seam/dependency injection would reduce the risk described and also probably allow for this to be marked parallel too.

hk21702 (approved):
Just that one comment from Copilot that is worth investigating a little further, but otherwise premptive LGTM!

windows-agent/internal/proservices/internal_test.go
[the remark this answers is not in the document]
CarlosNihelton: > unless under the hood it's using rand.Reader in the first place already.

That's exactly what it does :sweat_smile: 

That was a leftover when I tried migrating from `ecdsa` to `ed25519`. We can just replace the function as before, using the `opts` idiom, without affecting the `rand.Reader` assignment.

PR #1757, opened 2026-08-18
feat(windows-agent): Securefiles custodian for the agent's public directory

CarlosNihelton:
## Why

We want to enforce a specific set of invariants in all files and directories under the agent's public directory. That can only be achieved by centralising all file system accesses, thus getting rid of the ambient authority and making it an explicit capability shared only at places supposed to change that portion of the file system. This is why #1755 makes the certs package an in-memory generator, for example.

## The invariants we want to enforce

Windows files already containing the NT Extended Attributes `$LXUID`, `$LXGID` and `$LXMOD` with proper values are projected inside WSL instances with ownership and file modes dictated by those values. That behaviour is implemented on the Windows side of the 9P protocol, thus independent of instances's own `wsl.conf`. That's powerful because we can make the agent public directory look like owned by root, thus unprivileged processes inside any WSL instance cannot alter material that's read by the privileged `wsl-pro-service`, for example.

This is documented today, but it seems vague to me, as docs focus more on the Linux -> Windows flow, which is in fact dependent on `wsl.conf`. For that reason I'm defensively adding test cases that would flag breaking changes in the Windows 9P driver itself.

## Implementation

We create the `securefiles` package as the file system custodian component that must be passed to other parts of the `windows-agent` that need to write to the public directory and forbid raw access to the file system. Then, `securefiles` enforces the Windows->Linux file modes/ownership without callers having to even consider it. We create a few key tests that depend on real WSL2 guarded by a build tag so they are more often excluded from compilation, but preserve a CI job that only runs those, should the Windows driver change.

The custodian component refuses paths that escape its sub-tree and opens with `OBJ_DONT_REPARSE` so a symlink or junction cannot redirect a write out of the public directory. Fielsystem operations guarded by the custodian verify ownership of the already-open handle and should never stamp a file after publishing, so a descriptor cannot outlive the check.

We account for the possibility of a degraded mode: a filesystem that doesn't support EAs, and fail open by design, the agent still starts and logs that condition. Failing closed could break existing users and we don't know for sure if EAs are properly mapped in all sorts of network file sharing used in companies.

NT File Extended Attributes require low level APIs provided by the  `nt.dll` module, somewhat undocumented in the Go standard library, and their binary encoding is really messy, for which we bring the `go-winio` library (provided by Microsoft) to facilitate that aspect, otherwise it would be about 50 lines of error-prone byte punning that we'd have to write ourselves.

This is a complete vertical slice: all parts of the agent that needs to write files to that location are modified to use the `securefiles` package. A follow-up PR will then implement the `wsl-pro-service` side that should reject reading stuff not owned by root.

## Degradation of the custodian

I wanted truly enforced EAs on the agent's files read by `wsl-pro-service`. But some reasons seems to exist for breaking this feature at runtime even after we detect a healthy filesystem, so I decided to probe for those cases, log and keep the agent running, while on the `wsl-pro-service` the action will be refuse non-root owned files right away. 

A filter driver attaching after startup is one close to realistic scenario. A minifilter can attach to a volume at any time — an AV or EDR update, a policy push, a backup agent — and can fail EA operations on new creates, causing a transition from a healthy to a degraded state of the custodian. How likely that's in real like is hard to tell, but seems close to unplausible because that kind of software would be already running by the time the agent starts. 

Thus I changed my mind and decided to fail closed when writing the EAs are not supported. My assumption is that it's much easier to update the Windows agent than it is to update wsl-pro-service and, if that component assumes root-owned files, it means the agent won't be reachable if EAs are not supported, limiting the utility of the entire system. If practical reasons at runtime requires us to change that, there is one single atomic commit to revert to bring back the support for degraded mode.

---

UDENG-11035


CarlosNihelton:
Sorry for the big PR. Most of it are tests, though.

I noticed that my agent sneaked in some peasant improvements to our testing coverage not strictly related to the changes in PR, I'm considering those beneficial given the numbers: `+ Coverage   (main) 89.89%  (PR) 90.17%   +0.28%`

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
```diff
@@ -205,11 +221,11 @@ func (a *App) SetArgs(args ...string) {
...
 func (a *App) PublicDir() (string, error) {
 	// This wrapper is used to have a cleaner public API.
-	return a.publicDir(options{})
+	return a.publicDirPath(options{})
 }
 
-// publicDir is a wrapper around PublicDir to allow overriding its path with an option.
-func (a *App) publicDir(opts options) (string, error) {
+// publicDirPath is a wrapper around PublicDir to allow overriding its path with an option.
+func (a *App) publicDirPath(opts options) (string, error) {
```
hk21702, line 228: Related a little lower in this method, but on fresh install, we're making the publicDir with just a plain `os.MkdirAll`. This conflicts with `ADR 2.01`, as you now have a TOCTOU scenario where there is a window where the public dir is not secured.

windows-agent/internal/cloudinit/cloudinit.go
```diff
@@ -68,22 +64,61 @@ func (c CloudInit) writeAgentData() (err error) {
...
 type metadata struct {
 	InstanceID string `yaml:"instance-id"`
 }
 
+// startupPurge removes every node in the sub-tree that does not carry the agent's watermark and
+// regenerates the agent's own data file from the current configuration. Stamped nodes are left
+// untouched: cloud-init data is consumed exactly once, at the instance's first boot, so there is
+// nothing to gain from reading it back and rewriting it. The watermark is the whole adoption
+// policy: an unstamped node is foreign by definition, whatever its name.
+func (c CloudInit) startupPurge(ctx context.Context) error {
```
hk21702, line 87: Would this potentially purge files from users that are upgrading? Presumably, they would have unstamped nodes that were created before we introduced this system, resulting in a purge.

windows-agent/cmd/ubuntu-pro-agent/agent/agent.go
```diff
@@ -247,28 +263,24 @@ func (a *App) privateDir(opts options) (string, error) {
...
-	err = os.Rename(logFile, oldLogFile)
-	if err != nil && !errors.Is(err, os.ErrNotExist) {
-		log.Warningf(ctx, "Could not archive previous log file: %v", err)
+	// Rotate the current log to log.old in place using the custodian.
+	if err := c.Rename("log", "log.old"); err != nil && !errors.Is(err, os.ErrNotExist) {
+		log.Warningf(ctx, "Could not rotate log to log.old: %v", err)
 	}
 
-	f, err := os.OpenFile(logFile, os.O_APPEND|os.O_CREATE, 0600)
+	f, err := c.CreateFile("log")
```
hk21702, line 283: If the rename fails for any legitimate reason, there doesn't appear to be a fallback, and will cause the old logs to be removed. An append operation would probably be more appropriate here.

windows-agent/internal/securefiles/sys_linux.go
```diff
@@ -0,0 +1,156 @@
...
+
+// isOwned reports whether the node carries the custodian's watermark and still
+// has the same owner, group, and mode recorded at creation time.
+func (s *platformSys) isOwned(rel string) (bool, error) {
+	if s.mockOwned != nil {
+		return *s.mockOwned, nil
+	}
+	if s.degraded {
+		return true, nil
+	}
```
hk21702, line 98: I'm a little hesitant about simply returning true if degraded, since it seems like a fairly simple way to defeat the tamper-detection by somehow intentionally forcing us to think we're degraded.

At the very least, consider making it more noisy by logging something every time we do a check, and we are in degraded mode.

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+	fileOpen                  = 0x00000001
+	fileOpenIf                = 0x00000003
+	fileDirectoryFile         = 0x00000001
+	fileNonDirectoryFile      = 0x00000040
+	fileSynchronousIoNonAlert = 0x00000020
+
+	fileAddFile           = 0x0002
+	fileAddSubdirectory   = 0x0004
+	fileRenameInformation = 10
+)
```
hk21702, line 36: Aren't these already available in `golang.org/x/sys/windows` under `types_windows.go`?

Similarly, there's `windows.FileRenameInformation` for `fileRenameInformation`

windows-agent/internal/securefiles/custodian.go
```diff
@@ -0,0 +1,389 @@
...
+		return err
+	}
+
+	tmpRel := tempName(targetRel)
+
+	if err := c.sys.createNode(tmpRel, false); err != nil {
+		return mapEscape(err)
+	}
+	defer func() {
+		_ = c.root.Remove(tmpRel)
```
hk21702, line 247: Is it intentional to silently fail here instead of writing a log?

windows-agent/internal/cloudinit/cloudinit.go
```diff
@@ -68,22 +64,61 @@ func (c CloudInit) writeAgentData() (err error) {
...
+		// Directories are never adopted: this sub-tree legitimately holds files only.
+		if _, err := c.dir.ReadDir(rel); err == nil {
+			return false
+		}
+		owned, err := c.dir.IsOwned(rel)
+		if err != nil {
+			log.Warningf(ctx, "cloud-init: could not check ownership of %q: %v", rel, err)
+			return false
+		}
+		return owned
```
hk21702, line 98: This is potentially quite expensive, since ReadDir will go through the subtree, and then IsOwned will go through the subtree again right after ReadDir just closed their handler, even though ReadDir has in theory, already derived the information necessary to determine ownership.

Potentially, this inefficiency would scale linearly with the number of per-distro files left from prior sessions.

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+}
+
+type objectAttributes struct {
+	Length                   uint32
+	RootDirectory            windows.Handle
+	ObjectName               *unicodeString
+	Attributes               uint32
+	SecurityDescriptor       uintptr
+	SecurityQualityOfService uintptr
+}
```
hk21702, line 51: Could we use windows.OBJECT_ATTRIBUTES instead here? The only difference is that `SecurityDescriptor` and `SecurityQoS` are typed pointers instead of `uintptr`.

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+		uintptr(unsafe.Pointer(&oa)),   //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+		uintptr(unsafe.Pointer(&iosb)), //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+		0,
+		uintptr(fileAttributes),
+		uintptr(shareAccess),
+		uintptr(disposition),
+		uintptr(createOptions),
+		uintptr(unsafe.Pointer(&eaBuf[0])), //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+		uintptr(len(eaBuf)),
+	)
```
hk21702, line 196: Could this be replaced with windows.NtCreateFile?

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+		var r1Set uintptr
+		if testNtSetEaFileResult != nil {
+			r1Set = uintptr(*testNtSetEaFileResult)
+		} else {
+			r1Set, _, _ = procNtSetEaFile.Call(
+				uintptr(handle),
+				uintptr(unsafe.Pointer(&iosbSet)),  //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+				uintptr(unsafe.Pointer(&eaBuf[0])), //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+				uintptr(len(eaBuf)),
+			)
```
hk21702, line 216: Similarly, could we replace our custom call here with `windows.NtSetEaFile`?

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+	"golang.org/x/sys/windows"
+)
+
+var (
+	modntdll                 = windows.NewLazySystemDLL("ntdll.dll")
+	procNtCreateFile         = modntdll.NewProc("NtCreateFile")
+	procNtSetInformationFile = modntdll.NewProc("NtSetInformationFile")
+	procNtSetEaFile          = modntdll.NewProc("NtSetEaFile")
+	procNtQueryEaFile        = modntdll.NewProc("NtQueryEaFile")
+)
```
hk21702, line 23: I think there are implementations of these syscall wrappers in `x/sys/windows` already. Namely, `windows.NtCreateFile`, `windows.NtSetInformationFile`, `windows.NtSetEaFile`, and `windows.NtQueryEaFile`.

The caveat is that these return `error`, not raw uintptr status. The error is `windows.NTStatus`, meaning that on failure, you'd need a type assertion and an explicit `.Errno()` call on the asserted value.

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+	}
+
+	baseName16, err := windows.UTF16FromString(baseName)
+	if err != nil {
+		return err
+	}
+
+	uString := unicodeString{
+		Length:        uint16((len(baseName16) - 1) * 2), //#nosec G115 // UNICODE_STRING length of a short relative path; always fits in 16 bits.
+		MaximumLength: uint16(len(baseName16) * 2),       //#nosec G115 // UNICODE_STRING length of a short relative path; always fits in 16 bits.
```
hk21702, line 163: This is done a few places, but `windows.NewNTUnicodeString()` handles the length and max length calculations for us. It also means we can replace `unicodeString` with `windows.NTUnicodeString`.

windows-agent/internal/securefiles/sys_windows.go
```diff
@@ -0,0 +1,604 @@
...
+				gid, found.gid = val, true
+			case "$LXMOD":
+				mode, found.mode = val, true
+			}
+		}
+		if entry.NextEntryOffset == 0 {
+			break
+		}
+		offset += entry.NextEntryOffset
+	}
```
hk21702, line 359: This can probably be simplified with `winio.DecodeExtendedAttributes`.

You'd get a buffer from `windows.NtQueryEaFile(`, then get the list of EAs using `winio.DecodeExtendedAttributes`, and then we iterate over that instead of doing the traversal ourselves. We'd then also be able to get rid of `fileFullEaInformation`.

windows-agent/internal/securefiles/securefilestest/eareader_windows.go
```diff
@@ -0,0 +1,107 @@
...
+		if entry.EaValueLength == 4 {
+			val := binary.LittleEndian.Uint32(buf[valOffset : valOffset+4])
+			result[name] = val
+		}
+
+		if entry.NextEntryOffset == 0 {
+			break
+		}
+		offset += entry.NextEntryOffset
+	}
```
hk21702, line 104: Similar thing as with `sys_windows.go`, where you can simplify this traversal with `winio.DecodeExtendedAttributes`.

windows-agent/internal/securefiles/securefilestest/eareader_windows.go
```diff
@@ -0,0 +1,107 @@
...
+		uintptr(handle),
+		uintptr(unsafe.Pointer(&iosb)),   //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+		uintptr(unsafe.Pointer(&buf[0])), //#nosec G103 // NT syscall argument: pointer to live Go memory; the call is synchronous and kernel writes stay within the value.
+		uintptr(len(buf)),
+		0, // ReturnSingleEntry = FALSE
+		0,
+		0,
+		0,
+		1, // RestartScan = TRUE
+	)
```
hk21702, line 82: Similar thing, we can probably replace this with `windows.NtQueryEaFile()`

windows-agent/internal/securefiles/sys_windows.go
hk21702, line 1: In general, a significant portion of this can be simplified by using built-ins system call wrappers from `x/sys/windows`, and other built ins from `x/sys/windows`. Path traversal can be simplified with `winio.DecodeExtendedAttributes`.

hk21702 (changes requested):
A few security concerns here and there, plus some clarifications on intended behavior where we now sometimes silently fail.

Additionally, it seems to me that a lot of the places where we do NT filesystem syscalls can be simplified significantly by using built-ins from `golang.org/x/sys/windows` as well as some helpers from `winio`.

windows-agent/internal/cloudinit/cloudinit.go
```diff
@@ -68,22 +73,66 @@ func (c CloudInit) writeAgentData() (err error) {
...
+	}
+
+	// A node judged foreign but left behind is reported, not fatal. Refusing to start
+	// leaves it exactly where it is, still there to be consumed by cloud-init at the
+	// instance's first boot, and takes the agent down with it: the component that would
+	// have removed the node on a later run is the only thing lost.
+	removed, err := c.dir.Purge(isOurs)
+	if err != nil {
+		log.Errorf(ctx, "cloud-init: could not remove every unrecognised node from the sub-tree, and they stay readable by every instance: %v", err)
+	}
```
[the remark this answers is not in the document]
CarlosNihelton: I disagree. We don't have the capability today, so continue to run is the backwards compatible posture.

PR #1774, opened 2026-08-20
fix(CI): Less concurrency throttling

CarlosNihelton:
The `quality-go` jobs often fail due rate limits when setting up `protoc` or strange failures when running `golangci-lint`.
For example: https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/31633925897/job/94239362901 and https://github.com/canonical/ubuntu-pro-for-wsl/actions/runs/31633925897/job/94239362923.

Even on the happy paths those jobs are fast enough that the loss of parallelisation won't become an issue. The exact number is debatable. Other jobs not affected by those specific issues were not serialised in this PR.

Also, by inlining the Flutter integration tests in the `flutter-qa` action we skip some tooling setup steps. A new input is added to cope with the conditional need to run integration tests. 

The overall result should be a less flaky CI due reasons other than our own code. The overall duration of the QA job didn't change despite those arrangements. 

UDENG-11224


hk21702 (approved):
LGTM

PR #1780, opened 2026-08-24
feat(wsl-pro-service): Validate ownership and mode of projected agent files before reading

CarlosNihelton:
`wsl-pro-service` ingests the agent-written `gRPC` address file and `mTLS` material projected into WSL under the Public Directory (`%UserProfile%/.ubuntupro`). It previously trusted those files blindly.

Per ADR-2.01 (Secure Projection), the agent stamps NT Extended Attributes so 9P projects every node as root-owned with mode 0700 (dirs) / 0600 (files). This change adds the instance-side, defense-in-depth half of that contract:

- New `daemon.SecureReader` interface; `Daemon` depends on it for all projected file ingestion (address file, client cert/key pair, root CA).
- `defaultSecureReader` validates every path component from the Public Directory root to the target immediately before reading: no symlinks, regular files and directories only, strict uid/gid 0, strict modes 0700/0600, and lexical containment inside the Public Directory.
- The leaf file is opened with `O_NOFOLLOW` to close the TOCTOU window between validation and read.
- Validation failures are wrapped in `streams.SystemError` so the daemon exits loudly instead of entering the reconnect backoff; missing files keep the transient retry behaviour (agent not started yet).
- Refusal messages report only the actual observed state (e.g. `refused "<path>": not strictly owned by root (uid 1000, gid 1000)`).
- Platform split keeps `wsl-pro-service` buildable on Windows for development tooling; the Windows implementation fails closed.
- Test seams: unexported osFs backend injection for secure reader unit tests (via export_test.go, package-internal), and `daemon.WithSecureReader` with `testutils.MockSecureReader` for daemon/service tests, so no test requires root privileges.

ADR-2.01 updated to record the instance-side validation invariant.

---

UDENG-11036


wsl-pro-service/internal/daemon/daemon_test.go
```diff
@@ -239,6 +259,91 @@ func TestServe(t *testing.T) {
...
+
+	serveExit := make(chan error, 1)
+	go func() { serveExit <- d.Serve(&mockService{}) }()
+
+	select {
+	case err := <-serveExit:
+		require.Error(t, err)
+		require.ErrorIs(t, err, streams.SystemError{}, "insecure client key must cause immediate SystemError termination")
+		require.Contains(t, err.Error(), "not strictly owned by root")
+	case <-time.After(10 * time.Second):
```
hk21702: I'm not super enthusiastic about the idea of having time gates in tests like this. Unless it's very long, it seems likely to be a source of flakiness.

Could we instead mock things so that we receive a signal from a channel once we reach a portion of the code that shouldn't be reachable if we are properly serving an exit error?

wsl-pro-service/internal/daemon/daemon_test.go
```diff
@@ -239,6 +259,91 @@ func TestServe(t *testing.T) {
...
+	serveExit := make(chan error, 1)
+	go func() { serveExit <- d.Serve(&mockService{}) }()
+
+	// Missing file is transient (agent not started yet) — Serve should NOT return
+	// a SystemError; it should keep retrying. We assert it does not exit with an error
+	// within a short window.
+	select {
+	case err := <-serveExit:
+		require.Failf(t, "Serve should not have exited for missing key (transient)", "got %v", err)
+	case <-time.After(2 * time.Second):
```
hk21702: Diito, even more so here because of the short window.

wsl-pro-service/internal/daemon/daemon_test.go
```diff
@@ -239,6 +259,91 @@ func TestServe(t *testing.T) {
...
+		// still serving — expected
+	}
+
+	d.Quit(ctx, false)
+	select {
+	case err := <-serveExit:
+		if err != nil {
+			require.NotErrorIs(t, err, streams.SystemError{}, "missing key should not be a SystemError")
+		}
+	case <-time.After(5 * time.Second):
```
hk21702: Ditto.

wsl-pro-service/internal/daemon/secure_reader_test.go
```diff
@@ -0,0 +1,483 @@
...
+	}
+
+	for name, tc := range testCases {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			reader := daemon.NewDefaultSecureReader(tc.mockFS)
+			got, err := reader.ReadFile(tc.rootDir, tc.targetPath)
+			if tc.wantErr != "" {
+				require.Error(t, err)
```
hk21702: Could this be simplified with `require.ErrorContains`?

wsl-pro-service/cmd/wsl-pro-service/service/service_test.go
```diff
@@ -212,7 +213,10 @@ func TestAppRunFailsOnComponentsCreationAndQuit(t *testing.T) {
 	sys, mock := testutils.MockSystem(t)
 	mock.SetControlArg(testutils.WslInfoErr)
 
-	a := service.New(service.WithSystem(sys))
+	a := service.New(
+		service.WithSystem(sys),
+		service.WithDaemonOptions(daemon.WithSecureReader(testutils.NewMockSecureReader(nil))),
```
hk21702: Does testing the service absolutely require the real daemon? Could we instead mock the daemon entirely instead of passing in mocks for the daemon? It feels bad to break the barrier with `WithSecureReader` which now also needs `	testdetection.MustBeTesting()` in a file that is otherwise non-test code.

wsl-pro-service/internal/daemon/secure_reader_test.go
```diff
@@ -0,0 +1,483 @@
...
+		"Rejected when opening the root fails": {
+			rootDir:    "/root/public",
+			targetPath: "file.txt",
+			mockFS: mockOSFs{
+				rootInfo:    secureRootInfo,
+				openRootErr: errors.New("permission denied"),
+			},
+			wantErr: "could not open root",
+		},
+		"Rejected when the target path escapes the root": {
```
hk21702: How necessary are tests of this nature? It feels like we're just retesting `os.Root`. If we can just validate that `os.Root` is in the flow, that should be enough.

hk21702 (changes requested):
A few comments here and there, mostly regarding how we handling the testing.

wsl-pro-service/internal/daemon/secure_reader_test.go
CarlosNihelton: I did a small clean up in this region, and simplified the indirections in the secure_reader implementation, but I must state that I kept most of those because they are invariants of this component. Those invariants are currently upheld by `os.Root`, but if future maintainers try to switch to something loser, they'll find the expected requirements in code. I don't think there is a nicer way to verify this, honestly, without risking future code changes silently breaking those invariants.

wsl-pro-service/cmd/wsl-pro-service/service/service_test.go
CarlosNihelton: Indeed, we can get rid of this line without issues, because this fails earlier than actually starting the daemon. But the two tests calling `startDaemon()` are  some kind of integration testing, they try to connect to a mocked agent. The solution is not touching the service component at all and implement a FUSE fake filesystem :p

CarlosNihelton: It sounds a bit heavy-weighted but I like how clean it turns the testing side.

wsl-pro-service/internal/daemon/daemon_test.go
CarlosNihelton: I think I got a better arrangement for those tests.

wsl-pro-service/internal/daemon/secure_reader.go
```diff
@@ -0,0 +1,192 @@
...
+		return nil, fmt.Errorf("could not stat %q: %w", rootDir, err)
+	} else if fi.Mode()&fs.ModeSymlink != 0 {
+		return nil, fmt.Errorf("refused %q: root is a symlink", rootDir)
+	}
+
+	root, err := r.openRoot(rootDir)
+	if err != nil {
+		return nil, fmt.Errorf("could not open root %q: %w", rootDir, err)
+	}
+	defer root.Close()
```
hk21702, line 114: If it is a security concern that we don't follow symlinks, then this probably isn't sufficient since it would technically be subject to a TOCTOU swap.

You could maybe open the directory with a kernel enforced no symlink resolution, and then use openRoot against the descriptor.

hk21702 (changes requested):
One comment on ReadFile I'm making with the assumption that the symlink check is a hard security requirement.

If it is, then there is technically a window of attack in the current implementation given that the one-time check would only be good for that atomic moment, and after reads and traversals, technically no longer have the guarantee.

wsl-pro-service/internal/daemon/secure_reader.go
CarlosNihelton: I didn't think I'd replace os.Root so early :p 

The security constraints here are quite artificial, yet since we're here, let's do a stricter system. By replacing the Go stdlib `os.Root` with the `openat2` syscall, we can get closer to the all-symlink rejection policy, because with the flags being used, that syscall fails if any intermediary segment of the path is a symlink. Lstat is still needed, though, to validate permission and file mode.

CarlosNihelton: The sad part is that now we have more code.

wsl-pro-service/internal/testutils/fakefs_linux.go
hk21702, line 1: So, this mock might actually be representative of an attack vector where a fake FUSE filesystem lies about its metadata, in cases where an attacker has FUSE mount capabilities. An unprivileged user could rename or replace `.ubuntupro`, create a replacement mount point with a FUSE filesystem, and make it so that FUSE reports every node as UID/GID 0 with modes 0700/0600.

We probably need to validate the FS associated with an already-open root descriptor to verify that its 9P, and then add `unix.RESOLVE_NO_XDEV` to every `openat2` call.

In theory, someone could create an arbitrary 9P mount then, but that should at least require privilege.

wsl-pro-service/internal/daemon/secure_reader.go
hk21702: This is probably rather artificial, given that it requires mutation rights and mostly depends on if an unprivileged user can cause this from Windows, but currently, your loop to validate the nodes is traversing all the path components from root down to the target, but, you're reading through that pinned file descriptor. You never validated what the FD actually points to.

If `targetFile` was initially illegitimate, but an attacker swapped out the pathname with a compliant inode before your validation loop, then the original illegitimate `targetFile` would be read.

We can probably fix this by just validating `targetFile` itself before reading.

wsl-pro-service/internal/daemon/secure_reader_linux.go
```diff
@@ -0,0 +1,171 @@
...
+}
+
+func (s statFileInfo) Name() string       { return s.name }
+func (s statFileInfo) Size() int64        { return s.size }
+func (s statFileInfo) Mode() fs.FileMode  { return s.mode }
+func (s statFileInfo) ModTime() time.Time { return s.modTime }
+func (s statFileInfo) IsDir() bool        { return s.isDir }
+func (s statFileInfo) Sys() any           { return &s.sys }
+
+func fileInfoFromStat(name string, stat *unix.Stat_t) fs.FileInfo {
```
hk21702, line 133: This removes everything but `rwx` permissions, and we reconstruct the file-type bits in the switch, but we never reconstructed the special bits. Besides needing to translate setuid, setgid, and sticky bits explicitly, we'd then need to reject them explicitly in `defaultValidate`.

wsl-pro-service/debian/control
hk21702, line 1: Do we know for sure that we'd be able to use Fuse in the Launchpad and autopkgtest runners?

hk21702 (changes requested):

wsl-pro-service/debian/control
CarlosNihelton: The short answer for `autopkgtest` is yes, but the price is not worth it. We'd need to request  `isolation-machine` and we'd still need a test seam to inject the allowed filesystem type given we want to reject FUSE in production. The pattern for injecting the allow list extension is not my preferred, but given we'd have to resort to it anyway, we can use the same pattern to store the expected GID and UID values and get rid of FUSE entirely. :/ 5d8e57faaa71eef44de0c475108b9a88b4bcc347
Having reached that conclusion I didn't look into the Launchpad builder machines, but I suspect they run on unprivileged containers where `/dev/fuse` won't be available, thus that would likely fail at package build time anyway.

wsl-pro-service/internal/daemon/secure_reader_linux.go
CarlosNihelton: Given we're no longer using `os.Root` we can get rid of this and the surrounding boilerplate and use the `Stat_t` more directly.

wsl-pro-service/internal/testutils/fakefs_linux.go
CarlosNihelton: FUSE is not totally unprivileged either. But given it became iffy to integrate when we had to allow-list the filesystem types and the complications of using it in autopkgtests, better is to remove it entirely :/
It looked so nice how it helped tests from outside, but...

CarlosNihelton:
I had to rebase due conflicts in go.mod. New content starts at commit 2411ad18c74ed3a87d297b0877e129d38f9b5e0e.

wsl-pro-service/debian/control
CarlosNihelton: Although I don't like this pattern of overriding values via `init()`, it did simplify things and made the PR slightly smaller than before.

wsl-pro-service/internal/daemon/secure_reader.go
```diff
@@ -0,0 +1,183 @@
...
+// with strict permissions (0700 for directories, 0600 for regular files), as mandated by the
+// Secure Projection contract. Refusals report only the actual state observed.
+var (
+	expectedUID uint32
+	expectedGID uint32
+)
+
+func defaultValidate(stat fileStat) error {
+	if stat.UID != expectedUID || stat.GID != expectedGID {
+		return fmt.Errorf("not strictly owned by root (uid %d, gid %d)", stat.UID, stat.GID)
```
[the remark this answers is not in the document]
CarlosNihelton: Fine, there is a WIP branch for that, so as long as we don't merge this before that we should be fine.

wsl-pro-service/internal/daemon/secure_reader_linux.go
```diff
@@ -0,0 +1,173 @@
...
+// reject symlinks at all levels.
+type openat2Root struct {
+	fd   int
+	path string
+}
+
+// Production allowed filesystem magic numbers for WSL Windows-mounts.
+const (
+	Magic9P       = unix.V9FS_MAGIC // 0x01021997
+	MagicVirtioFS = 0x5a657366      // virtiofs
```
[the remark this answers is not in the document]
CarlosNihelton: Very good finding! Indeed `virtiofs` relies on FUSE so it reports `FUSE_SUPER_MAGIC` to stat calls, so this approach cannot be used as a defense mechanism. We have to iterate over `/proc/self/mountinfo` instead. :/

CarlosNihelton:
> The fixes for the broken links in documentation will come in a separate PR.

https://github.com/canonical/ubuntu-pro-for-wsl/pull/1831

PR #1781, opened 2026-08-26
fix(CI): TiCS buildable scores drop

CarlosNihelton:
Since the GH runners migrated to VS 2026 and we migrated to the new packaging tool for Windows, TiCS lost complete track of the C++ source files we have in this repo.

We need to:
1. make sure our builds can still happen in VS 2022 and 
2. make an easy entrypoint for TiCS.

I'm addressing 1 by running some steps of our CI in a matrix with `windows-2022` and `windows-2025` runners, and 2 by adding a `Makefile` that the TiCS tool can invoke to watch compilation of those sources in it's own environment.

With those changes in place we see a reduction of "Unbuidable files" from 20 to 9. Further improvements may come next, but those are worth promoting now.

image


hk21702 (approved):
Unfortunate about the need to downgrade. LGTM.

CarlosNihelton:
> Unfortunate about the nerd to downgrade. LGTM.

:nerd_face:

PR #1782, opened 2026-08-26
fix(CI): TiCS requires running on windows-2022

CarlosNihelton:
Because that's the runner that comes with VS 2022.

That was an oversight from the previous PR related to TiCS.


CarlosNihelton:
GitHub Actions remained stale for a moment, I had to force push some changes to make the CI jobs show up again.

hk21702 (approved):

PR #1812, opened 2026-09-10
docs: add docs AGENTS file

edibotopic:
This adds a docs-specific AGENTS.md for consistency when using LLMs with docs-related tasks.

It follows a similar pattern used in the Server and ADSys docs.

UDENG-10951


CarlosNihelton (approved):
I think it's worth referencing this file in the repo root AGENTS.md as well.

edibotopic:
@CarlosNihelton --- good point. Oversight. Any convention for doing that? Just relative paths?

CarlosNihelton:
I'd just cite it in the last section via relative path with a prose saying something like "When working on user facing documentation, follow the rules in `docs/AGENTS.md`.

PR #1816, opened 2026-09-15
fix(wsl-pro-service): Pass the Pro token via ephemeral config file in System.ProAttach

CarlosNihelton:
`wsl-pro-service` currently passes the Pro token to the pro-client via CLI, which is leaky by default.
Since version `27.7` the pro client supports receiving the token via a config file, all versions available in Focal and later support that flag. Using it means the `wsl-pro-service` must write a file only readable by root containing the token and pass it to the client. Better yet if we make that file private and temporary. Thanks to systemd `PrivateTmp` feature, we can pretend to write to `/tmp` and get a private location where our children processes can also read from with limited exposure.

---
UDENG-11491


CarlosNihelton:
I can't make `yaml.Marshall` fail to serialize a string nor it's worth making it a test seam, thus we'll have to accept the low patch coverage.

.gitignore
hk21702, line 1: What are the changes here for?

wsl-pro-service/internal/system/pro.go
```diff
@@ -32,14 +35,37 @@ func (s System) ProStatus(ctx context.Context) (attached bool, err error) {
...
+		Token string `yaml:"token"`
+	}
+
+	cfgData, err := yaml.Marshal(attachConfig{Token: token})
+	if err != nil {
+		return fmt.Errorf("could not serialize attach config: %v", err)
+	}
+
+	tmpFile, err := os.CreateTemp("", "pro-attach-*.yaml")
+	if err != nil {
```
hk21702: I think this respects `$TMPDIR`, meaning its possible to escape systemd's `PrivateTmp`. It should probably be specific if the intent is to take advantage of `PrivateTmp`. I don't think it is a significant issue since the service is handled by systemd, and `$TMPDIR` requires privilege to change, but being explicit would eliminate the assumption.

wsl-pro-service/internal/system/pro.go
```diff
@@ -32,14 +35,37 @@ func (s System) ProStatus(ctx context.Context) (attached bool, err error) {
...
+		return fmt.Errorf("could not create temporary attach config file: %v", err)
+	}
+	defer os.Remove(tmpFile.Name())
```
hk21702, line 64: This may be considered a security issue if we're relying on the removal of the temp config file, but we're discarding errors. It's probably fine given that we're using `PrivateTmp`, but something to keep in mind, possibly adding a comment..

hk21702 (changes requested):

.gitignore
CarlosNihelton: A directory where I usually dump text files for agents: plans, tasks, handoffs etc. Are you aware of any convention that would make it explicit? I've seen all sorts of patterns, none really standout. I considered naming the directory `.run/` or `.tmp/` but didn't find those explicit either. `.agents/` is already taken for other purposes so I didn't want to use that.

wsl-pro-service/internal/system/pro.go
CarlosNihelton: Good point. It's theoretically still easier for an attacker to inject environment variables than overwrite variables in the process memory.

wsl-pro-service/internal/system/pro.go
CarlosNihelton: Even logging that error may lead to undesired consequences, so a comment is all we got. Returning it is not an option because it would be possible to have errors returned when the main post condition of that function would have been achieved: pro-attach.

.gitignore
hk21702: The one I've been using has been `.scratch/` but, you could probably just use whatever as long as the agent knows to use it.

hk21702 (approved):
LGTM, I'll let you decide how you want to handle your agent files directory.

